PR #20, opened 2020-06-09
Update docker commands in Workflow, Signal, Query

devhritwik:
The docker commands didn't work and the changes have been made to make them work with the latest configuration.


rylandg:
@feedmeapples can you take a look at this for me?

PR #21, opened 2020-06-09
Update the Query Method getCount

devhritwik:
int causes problem to be interpreted on command line, shows the error JSON cannot be parsed. 
String can be parsed and hence changed accordingly.

### DIFF
--- a/docs/java-quick-start.md
+++ b/docs/java-quick-start.md
@@ -353,7 +353,7 @@ Update the workflow code to:
     void updateGreeting(String greeting);
 
     @QueryMethod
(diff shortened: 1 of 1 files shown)


mfateev:
Thanks a lot for reporting! I believe this should be fixed in the CLI and work with int.

cullywakelin:
Thanks for point out this issue. Rather than update the docs, where are releasing this fix: https://github.com/temporalio/temporal/pull/542 which should merge by the end of today.

Let us know if you see any other issues with the docs! https://community.temporal.io

PR #24, opened 2020-06-24
Fix relative links

feedmeapples:


feedmeapples:
> Can I ask how this was tested?

tested locally

Tried few options:
1. 
a) Navigate to any page that has a relative link (ex. `localhost:3000/docs/overview`)
b) Append a forward slash to the current URL (ex. `localhost:3000/docs/overview/`)
c) Press on the relative link (ex. _fault-oblivious stateful programming model_ )

2. Same as 1, without the step #2

3. 
a) Make few navigations using the sidebar
b) press on a relative link

PR #47, opened 2020-07-15
Refactored sidebar categories

cullywakelin:
The information architecture needs some work to provide a better user experience with the current information.
The first step is refactoring the sidebar.


sidebars.js
```diff
@@ -1,63 +1,87 @@
 module.exports = {
   docs: {
```
cullywakelin: Here is what this would look like:
image

sidebars.js
```diff
@@ -1,91 +1,97 @@
...
-          'java-distributed-cron',
-        ],
-      },
-    ],
-    'About': [
-      'docs-sla',
-      'license',
-    ],
-  },
+  docs: [
```
cullywakelin, line 2: Here is what this would look like:

Screen Shot 2020-07-16 at 11 37 38 AM

PR #58, opened 2020-07-27
New "How to route Tasks" guide

cullywakelin:
New user guide that covers "Task routing" and how to do it using Task Queues.

This PR does not yet add the guide to sidebar navigation. To preview the guide, go here:
https://deploy-preview-58--mystifying-fermi-1bc096.netlify.app/docs/next/route-tasks/

### DIFF
--- a/docs/route-tasks.md
+++ b/docs/route-tasks.md
@@ -18,7 +18,7 @@ Task routing is the development technique that pairs Tasks with Workers using Ta
 2. Register Workflow and/or Activity Tasks with the Worker.
 3. Configure Workflows and Activities to send Tasks to the Task Queue.
 
-Let's look a simple "Hello World" sample to see how this is done. In the following code snippet, we [create a new Worker](https://github.com/temporalio/go-samples/blob/master/helloworld/worker/main.go#L26) that is subscribed to the "hello-world" Task Queue and is registered to handle the `helloworld.Workflow` and `helloworld.Activity`:
+Let's look a simple "Hello World" sample to see how this is done. In the following code snippet we create a new Worker, subscribe it to the "hello-world" Task Queue, and register it to handle "Hello World" Workflow and "Hello World" Activity implementations:
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
@@ -33,6 +33,7 @@ import TabItem from '@theme/TabItem';
 
 <TabItem value="java">
 
+[Sample source reference](https://github.com/temporalio/java-samples/blob/master/src/main/java/io/temporal/samples/hello/HelloActivity.java#L91-L99)
 ```java
 // Worker is created and subscribed to the "hello-world" Task Queue 
 Worker worker = factory.newWorker("hello-world");
@@ -45,8 +46,7 @@ worker.registerActivityImplementationTypes(HelloWorldActivityImpl)
 </TabItem>
 <TabItem value="go">
 
-[source](https://github.com/temporalio/temporal-go-samples/tree/master/helloworld)
-
+[Sample source reference](https://github.com/temporalio/go-samples/blob/master/helloworld/worker/main.go#L20-L23)
 ```go
 // Worker is created and subscribed to the "hello-world" Task Queue
 w := worker.New(c, "hello-world", worker.Options{})
@@ -60,7 +60,7 @@ w.RegisterActivity(helloworld.Activity)
 </TabItem>
 </Tabs>
 
-In the next code snippet, we [start the `helloworld.Workflow`](https://github.com/temporalio/go-samples/blob/master/helloworld/starter/main.go#L25):
+In the next code snippet, we start the "Hello World" Workflow:
 
 <Tabs
     defaultValue="go"
@@ -72,6 +72,7 @@ In the next code snippet, we [start the `helloworld.Workflow`](https://github.co
 
 <TabItem value="java">
 
+[Sample source reference](https://github.com/temporalio/java-samples/blob/master/src/main/java/io/temporal/samples/hello/HelloActivity.java#L91-L99)
 ```java
 HelloWorldWorkflow workflow =
     client.newWorkflowStub(
@@ -85,8 +86,9 @@ WorkflowExecution workflowExecution =
 ```
 
 </TabItem>
-<TabItem value="go"> 
+<TabItem value="go">
 
+[Sample source reference](https://github.com/temporalio/go-samples/blob/master/helloworld/starter/main.go#L20-L25)
 ```go
 // Workflow Options specify the "hello-world" Task Queue
 // Which is the Task Queue that the Workflow will send its Tasks to
@@ -114,13 +116,11 @@ If a Worker pulls a Task that it has not been registered to run, an "unknown wor
 
 ## Task routing for Activity dependencies
 
-The Temporal service will persist Activity results as long as the results are less than 2MB in size and can be serialized for storage (which happens to be true for most Activity results). However, there are some scenarios where an Activity execution will do something that results in data, such as a large file, being stored directly on the host where the Worker is running. It is also possible that the next Activity in the Workflow has a dependency on that data and needs to run on the same host.
+The Temporal service will persist Activity results as long as the results are less than 2MB in size and can be serialized for storage (which happens to be true for most Activity results). However, there are some scenarios where an Activity execution will do something that results in data or files being stored directly on the host where the Worker is running. It is also possible that the next Activity in the Workflow has a dependency on that data and needs to run on the same host.
 
 Let's look at a file processing Workflow example where files are downloaded, processed in some way, and then uploaded to some other location. If a large video file is being processed only the name of the file can persist normally. The file itself must persist locally on the Worker host, as it is too big for normal Temporal storage. This means that any future Activities that act on the video file must also take place on the host where the Worker downloaded it.
 
-So, how do we ensure that a Worker on the correct host will execute the next Activity that depends on the video file?
-
-This is accomplished in different ways depending on the SDK.
+So, how do we ensure that a Worker on the correct host will execute the next Activity that depends on the video file? This is accomplished in different ways depending on the SDK.
 
 ### Java
 
--- a/docs/route-tasks.md
+++ b/docs/route-tasks.md
@@ -13,6 +13,7 @@ The question is how do we make sure a specific Worker is executing a specific Ta
 ## Task routing with Task Queues
 
 Task Queues are what enable "Task routing". Task routing is the development technique that pairs Tasks with Workers. The basics of it can be broken down into 3 steps:
+
 1. Create a Worker that is registered to handle Workflow and/or Activity Tasks.
 2. Subscribe the Worker to listen to a Task Queue.
 3. Configure Workflows and Activities to send Tasks to the Task Queue.
@@ -67,7 +68,7 @@ This is accomplished in different ways depending on the SDK.
 
 ### Java
 
-Consider the [file processing sample in Java](https://github.com/temporalio/java-samples/tree/master/src/main/java/io/temporal/samples/fileprocessing), the paradigm is to create two Workers. Each will run on the same host and thus have access to the same file system, but one will listen to a global (shared) Task Queue, which it will use to start the Workflow, and the other will listen to its own host-specific (private) Task Queue. 
+Consider the [file processing sample in Java](https://github.com/temporalio/java-samples/tree/master/src/main/java/io/temporal/samples/fileprocessing). The paradigm is to create two Workers. Each will run on the same host and thus have access to the same file system, but one will listen to a global (shared) Task Queue, which it will use to start the Workflow, and the other will listen to its own host-specific (private) Task Queue. 
 
 The following code snippet is stripped down from the [FileProcessingWorker.java sample](https://github.com/temporalio/java-samples/blob/master/src/main/java/io/temporal/samples/fileprocessing/FileProcessingWorker.java) and highlights the creation of two workers for this purpose:
 
@@ -118,21 +119,14 @@ The host-specific Task Queue is then set on [subsequent Activities](https://gith
 
 ```java
 private void processFileImpl(URL source, URL destination) {
-    StoreActivities.TaskQueueFileNamePair downloaded = defaultTaskQueueStore.download(source);
-
     // Initialize stubs that are specific to the returned task queue.
     ActivityOptions hostActivityOptions =
         ActivityOptions.newBuilder()
-        .setTaskQueue(downloaded.getHostTaskQueue())
+        // Set the host-specific Task Queue
+        .setTaskQueue(hostSpecificTaskQueue)
         .build();
-    StoreActivities hostSpecificStore =
-        Workflow.newActivityStub(StoreActivities.class, hostActivityOptions);
 
-    // Call processFile activity to zip the file.
     // Call the Activity to process the file using worker-specific task queue.
-    String processed = hostSpecificStore.process(downloaded.getFileName());
-    // Call upload activity to upload the zipped file.
-    hostSpecificStore.upload(processed, destination);
 }
 ```
 
--- a/docs/route-tasks.md
+++ b/docs/route-tasks.md
@@ -14,13 +14,38 @@ The question is how can we ensure a specific Worker will execute a given Task? W
 
 Task routing is the development technique that pairs Tasks with Workers using Task Queues. At the basic level it consists of 3 steps:
 
-1. Create a Worker that is registered to handle Workflow and/or Activity Tasks.
-2. Subscribe the Worker to listen to a Task Queue.
+1. Subscribe a Worker to a Task Queue.
+2. Register Workflow and/or Activity Tasks with the Worker.
 3. Configure Workflows and Activities to send Tasks to the Task Queue.
 
-Let's look at the ["Hello World" sample in Go](https://github.com/temporalio/temporal-go-samples/tree/master/helloworld) to see how this is done.
+Let's look a simple "Hello World" sample to see how this is done. In the following code snippet, we [create a new Worker](https://github.com/temporalio/go-samples/blob/master/helloworld/worker/main.go#L26) that is subscribed to the "hello-world" Task Queue and is registered to handle the `helloworld.Workflow` and `helloworld.Activity`:
 
-In the following code snippet, we [create a new Worker](https://github.com/temporalio/go-samples/blob/master/helloworld/worker/main.go#L26) that is subscribed to the "hello-world" Task Queue and is registered to handle the `helloworld.Workflow` and `helloworld.Activity`:
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs
+    defaultValue="java"
+    values={[
+        { label: 'Java', value: 'java', },
+        { label: 'Go', value: 'go', },
+    ]
+}>
+
+<TabItem value="java">
+
+```java
+// Worker is created and subscribed to the "hello-world" Task Queue 
+Worker worker = factory.newWorker("hello-world");
+// Worker is registered to handle helloworld Workflow
+worker.registerWorkflowImplementationTypes(HelloWorldWorkflowImpl.class)
+// Worker is registered to handle helloworld Activity
+worker.registerActivityImplementationTypes(HelloWorldActivityImpl)
+```
+
+</TabItem>
+<TabItem value="go">
+
+[source](https://github.com/temporalio/temporal-go-samples/tree/master/helloworld)
 
 ```go
 // Worker is created and subscribed to the "hello-world" Task Queue
@@ -32,7 +57,35 @@ w.RegisterWorkflow(helloworld.Workflow)
 w.RegisterActivity(helloworld.Activity)
 ```
 
+</TabItem>
+</Tabs>
+
 In the next code snippet, we [start the `helloworld.Workflow`](https://github.com/temporalio/go-samples/blob/master/helloworld/starter/main.go#L25):
+
+<Tabs
+    defaultValue="go"
+    values={[
+        { label: 'Java', value: 'java'},
+        { label: 'Go', value: 'go' },
+    ]           
+}>
+
+<TabItem value="java">
+
+```java
+HelloWorldWorkflow workflow =
+    client.newWorkflowStub(
+        HelloWorldWorkflow.class,
+        // Workflow Options specify the "hello-world"" Task Queue
+        // Which is the Task Queue that the Workflow will send its Tasks to
+        WorkflowOptions.newBuilder().setTaskQueue("hello_world").build());
+// Workflow is executed
+WorkflowExecution workflowExecution =
+    WorkflowClient.start(workflow::helloWorld);
+```
+
+</TabItem>
+<TabItem value="go"> 
 
 ```go
 // Workflow Options specify the "hello-world" Task Queue
@@ -49,6 +102,9 @@ we, err := c.ExecuteWorkflow(
   "Temporal",
 )
 ```
+
+</TabItem>
+</Tabs>
 
 :::caution
 
--- a/docs/route-tasks.md
+++ b/docs/route-tasks.md
@@ -4,8 +4,15 @@ title: How to route Tasks
 sidebar_label: Route Tasks
 ---
 
+## Overview
+
+This guide covers the recommended techniques to ensure that a Workflow and/or Activity will be executed by an appropriate Worker. In this guide we will look at two examples:
+
+1. A basic example that illustrates the fundamentals of routing a Task to a Worker.
+2. A more complex example involving Activities with host-specific file dependencies.
+
 Consider the specifications of a Worker:
-- A Worker only handles [Activity](docs/learn-activities) and [Workflow](docs/learn-workflows) Tasks that it is registered to handle. This means that a Task must be paired with a Worker that is registered to handle that Task.
+- A Worker only handles [Activity](activities) and [Workflow](workflows) Tasks that it is registered to handle. This means that a Task must be paired with a Worker that is registered to handle that Task.
 - A Worker runs on a specific host. This means that Activity Tasks that depend on host-specific resources, such as a local file, must be executed by a Worker on that host.
 
 The question is how can we ensure a specific Worker will execute a given Task? We do this through "Task routing".
@@ -18,7 +25,7 @@ Task routing is the development technique that pairs Tasks with Workers using Ta
 2. Register Workflow and/or Activity Tasks with the Worker.
 3. Configure Workflows and Activities to send Tasks to the Task Queue.
 
-Let's look a simple "Hello World" sample to see how this is done. In the following code snippet we create a new Worker, subscribe it to the "hello-world" Task Queue, and register it to handle "Hello World" Workflow and "Hello World" Activity implementations:
+Let's look at a simple "Hello World" sample to see how this is done. In the following code snippet we create a new Worker, subscribe it to the "hello-world" Task Queue, and register it to handle "Hello World" Workflow and "Hello World" Activity implementations:
 
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
@@ -60,7 +67,7 @@ w.RegisterActivity(helloworld.Activity)
 </TabItem>
 </Tabs>
 
-In the next code snippet, we start the "Hello World" Workflow:
+In the next code snippet, we configure the "Hello World" Workflow to send its Tasks to the "hello-world" Task Queue and then start its execution:
 
 <Tabs
     defaultValue="go"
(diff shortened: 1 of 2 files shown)


docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
+---
+id: route-tasks
+title: How to route Tasks
+sidebar_label: Route Tasks
+---
+
+Consider the specifications of a Worker:
+- A Worker is meant to handle only the Tasks that it is registered to handle. This means that Workers registered to handle Tasks for Workflow A are not capable of handling Tasks for Workflow B, and vice versa.
```
rylandg, line 8:
```suggestion
- A Worker only handles the Tasks that it is registered to handle. This means that Workers registered to handle Tasks for Workflow A are not capable of handling Tasks for Workflow B, and vice versa. 
```

Other than rewording the first sentence, I think the description is a bit confusing. It sort of suggests that a Worker can't handle tasks for both Workflow A and B inherently. Obviously I can have Worker listen on a single taskqueue and support multiple types (A and B).

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
+---
+id: route-tasks
+title: How to route Tasks
+sidebar_label: Route Tasks
+---
+
+Consider the specifications of a Worker:
+- A Worker is meant to handle only the Tasks that it is registered to handle. This means that Workers registered to handle Tasks for Workflow A are not capable of handling Tasks for Workflow B, and vice versa. 
+- A Worker runs on a speicific host. This means that Activity Tasks that depend on resources specific to a particular host must be executed by a Worker on that host.
```
rylandg, line 9:
```suggestion
- A Worker runs on a speicific host. This means that Activity Tasks that depend on host-specific resources must be executed by a Worker on that host.
```

I think giving a one sentence example might be helpful here. Ie: file on the filesystem

rylandg: Also the word specific is mispelled.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+So, for any given Temporal Workflow, how do we make sure a specific Worker is executing a specific Task?
```
rylandg, line 11: The double comma makes this sentence complex and unfriendly IMO.

rylandg: Maybe something like, "This means the question is..."

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+sidebar_label: Route Tasks
+---
+
+Consider the specifications of a Worker:
+- A Worker is meant to handle only the Tasks that it is registered to handle. This means that Workers registered to handle Tasks for Workflow A are not capable of handling Tasks for Workflow B, and vice versa. 
+- A Worker runs on a speicific host. This means that Activity Tasks that depend on resources specific to a particular host must be executed by a Worker on that host.
+
+So, for any given Temporal Workflow, how do we make sure a specific Worker is executing a specific Task? 
+
+## Use Task Queues
```
rylandg, line 13: Why "Use"? I say drop it.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+"Task Routing" is the development technique that makes use of Task Queues to orchestrate the execution of Tasks by specific Workers.
```
rylandg, line 15: This sentence is very wordy and sort of cyclical.

rylandg: Specifically " is the development technique that makes use of Task Queues" but the whole thing could probably be simpler.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+Let's look at the ["Hello World" sample in Go](https://github.com/temporalio/temporal-go-samples/tree/master/helloworld) to see how this is done.
+
+Here, we are creating a new Worker that is subscribed to the "hello-world" Task Queue, and is registered to handle the `helloworld.Workflow` and `helloworld.Activity`:
```
rylandg, line 19: I'm not a fan of "here" or "this". Also I think linking to the exact line of the sample as a hyperlink would be nice.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+// The Worker is created and subscribed to the "hello-world" Task Queue
+w := worker.New(c, "hello-world", worker.Options{})
+defer w.Stop()
+// The Worker is registered to handle the helloworld Workflow
+w.RegisterWorkflow(helloworld.Workflow)
+// The Worker is registered to handle the helloworld Activity
+w.RegisterActivity(helloworld.Activity)
+```
+
+And here, we are starting the `helloworld.Workflow`:
```
rylandg, line 31: Again no here. Maybe, "In the next part of the code we start..."

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+## Use Task Queues
+
+"Task Routing" is the development technique that makes use of Task Queues to orchestrate the execution of Tasks by specific Workers.
+
+Let's look at the ["Hello World" sample in Go](https://github.com/temporalio/temporal-go-samples/tree/master/helloworld) to see how this is done.
+
+Here, we are creating a new Worker that is subscribed to the "hello-world" Task Queue, and is registered to handle the `helloworld.Workflow` and `helloworld.Activity`:
+
+```go
+// The Worker is created and subscribed to the "hello-world" Task Queue
```
rylandg, line 22: Don't start code comments with "The". You'll notice that almost all code comments could be started with The.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+w.RegisterWorkflow(helloworld.Workflow)
+// The Worker is registered to handle the helloworld Activity
+w.RegisterActivity(helloworld.Activity)
+```
+
+And here, we are starting the `helloworld.Workflow`:
+
+```go
+// The Workflow Options specify the "hello-world" Task Queue
+// This is the Task Queue that the Workflow will add its Tasks to
```
rylandg, line 35: I think "send" instead of "add" is more accurte.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+// The Workflow Options are passed to the execution of the Workflow
+we, err := c.ExecuteWorkflow(
+  context.Background(), 
+  workflowOptions, 
+  helloworld.Workflow, 
+  "Temporal",
+)
+```
+
+Task routing can be as simple as making sure a Worker is subscribed to a Task Queue and Tasks are being added to it.
```
rylandg, line 49: I don't find that this sentence adds value.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+:::caution
+
+If a Worker pulls a Task that it has not been registered to run, an "unknown workflow type" error will occur.
+
+:::
+
+## Task routing for Activity dependencies
+
+The Temporal service will persist Activity results as long as the results can be serialized and stored in a Cassandra or MySQL database (which happens to be true for most Activity results). However, there are some scenarios where an Activity execution will do something that results in data being stored on the host where the Worker is running. It is also possible that the next Activity in the Workflow has a dependency on that data.
```
rylandg, line 59: "The Temporal service will persist Activity results as long as the results can be serialized and stored in a Cassandra or MySQL database (which happens to be true for most Activity results)"

Few things:
* Dont think Cassandra or MySQL are relevant here. They are really a technical detail. The limits are 2mb and must be serializable, datastore doesn't matter
* "Persist Activity results" isn't wrong but it's sort of not the point you're trying to make. It might be worth rewording to use the term "event history".


```suggestion
The Temporal service will persist Activity results as long as the results can be serialized and stored in a Cassandra or MySQL database (which happens to be true for most Activity results). However, in some cases you may have an Activity which creates large files which are stored directly on the host file system of its worker. If the next activity in the workflow depends on that same file it will need to be run on the same host.
```

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+Let's look at a file processing Workflow, as an example,  where files are downloaded, processed in some way, and then uploaded to some other location. If a large video file is being processed only the name of the file can persist normally. The file itself must persist locally on the Worker host. This means that any future Activities that act on the video file must also take place on the host where the Worker downloaded it.
```
rylandg, line 61:
```suggestion
Let's look at a concrete file processing Workflow example where  where files are downloaded, processed in some way, and then uploaded to some other location. If a large video file is being processed only the name of the file can persist normally. The file itself must persist locally on the Worker host. This means that any future Activities that act on the video file must also take place on the host where the Worker downloaded it.
```

I think you should focus more on what limits the activities from passing the file back. You just sort of skip over that part.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+So, how do we ensure that a Worker on the correct host will execute the next set of Activities that depends on the video file?
```
rylandg, line 63: "set of Activities" makes it confusing, just talk about a single next Activity.

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+Each SDK offers a slightly different way to handle this.
```
rylandg, line 65: "Each SDK" is sort of a negative reflection on our product. I would say, "This is accomplished in different ways depending on the SDK".

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Java
+
+In Java the paradigm is to create two Workers. Each will exist on the same host and thus have access to the same file system, but one will listen to a shared Task Queue, which it will poll to start the Workflow, and the other will listen to its own private Task Queue.
```
rylandg, line 69: "paradigm" meh - best practice maybe

rylandg: "exist" isn't the right word.

Also the "poll to start the Workflow" part isn't super easy to grasp

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+```java
+// Initialize a WorkerFactory
+WorkerFactory factory = WorkerFactory.newInstance(client);
+// Create a Worker that is subscribed to a shared TaskQueue
```
rylandg, line 74: `TaskQueue` one word or two?

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+final Worker workerForCommonTaskList = factory.newWorker(globalTaskQueue);
+// Register Workflow and Activities
+// Create a Worker that is subscribed to a private TaskQueue
+final Worker workerForHostSpecificTaskList = factory.newWorker(hostSpecifiTaskQueue);
+// Register activities
+```
+
+Then, design your Workflow Activities to return an object that includes both the Activity result and the name of the private Task Queue to which the next Activity Task will be added.
```
rylandg, line 82: I wouldn't say "then design your"

I think it might be worth mentioning that the first activity will run on arbitrary host and pass back the host for the subsequent activities

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+    private final String activityResult;
+
+    // Get and set methods ...
+  }
+
+  // Interface methods ....
+}
+```
+
+The [FileProcessing sample in Java](https://github.com/temporalio/temporal-java-samples/tree/143902cba3e14aa11f3b90784baf53406c329fd2/src/main/java/io/temporal/samples/fileprocessing) shows an example of this design.
```
rylandg, line 100: This link could be dangerous. Why isn't it tied to master?

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Go
+
+In Go, you actually have two options.
```
rylandg, line 104:
```suggestion
In Go, you have two options.
```

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+The first option is to do the same thing as you would do in Java, which is to create two Workers (one listens to the shared Task Qeue and the other listens to the private Task Queue) and return the name of the private Task Queue back to the Workflow as an Activity result.
```
rylandg, line 106:
```suggestion
**Option 1**: same design as Java, which is to create two Workers (one listens to the shared Task Qeue and the other listens to the private Task Queue) and return the name of the private Task Queue back to the Workflow as an Activity result.
```

docs/route-tasks.md
```diff
@@ -0,0 +1,138 @@
...
+
+The second option is to create a Workflow Session which automatically ties any remaining Workflow Activity executions to the Worker that started the Workflow.
```
rylandg, line 108:
```suggestion
**Option 2**: create a Workflow Session which automatically ties any remaining Workflow Activity executions to the Worker that started the Workflow.
```

rylandg:
Also what do you think of starting a practice where we always provide code examples at the top of each page. A lot of people probably just want to learn directly from the code.

docs/route-tasks.md
cullywakelin: The heading answers the question above it. But I see your point in terms of heading "consistency", i.e. using statements, questions, or topics as headers. I disagree that we should be consistent on that point though as it is all about accurately labeling information sections, and I want to be able to use any of them to best inform a user. Limiting ourselves like that is ... silly IMO, but will concede on this one.

docs/route-tasks.md
cullywakelin: I preferred the Workflow context that the prepositional phrase provides... but OK sure.

docs/route-tasks.md
cullywakelin: AFAIK, Events are not the same as Activity results. So I don't think that would make sense.

docs/route-tasks.md
cullywakelin: I am hesitant to use "concrete" as an adjective anywhere in this documentation.
Concrete: adjective
existing in a material or physical form; not abstract.

While the code sample is indeed a comprehensive organization of these components at work - it is far from "concrete" in its existence.

docs/route-tasks.md
cullywakelin: > "paradigm" meh - best practice maybe

paradigm (noun): a typical example or pattern of something; a model.
best practice (noun): commercial or professional procedures that are accepted or prescribed as being correct or most effective.

It is much more of a pattern/model - there is really no other alternative to make this happen. Paradigm is by far a better word choice here IMO. Best practice suggests there a bunch of ways to do it - but through experience it is considered the better way... which just is not the case here. This is the pattern that Temporal forces a developer to adhere to -

cullywakelin:
> Also what do you think of starting a practice where we always provide code examples at the top of each page. A lot of people probably just want to learn directly from the code.

I think that is what the go example repos are for - 
I am not able to envision how that would look really...

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+Consider the specifications of a Worker:
+- A Worker only handles [Activity](docs/learn-activities) and [Workflow](docs/learn-workflows) Tasks that it is registered to handle. This means that a Task must be paired with a Worker that is registered to handle that Task.
+- A Worker runs on a specific host. This means that Activity Tasks that depend on host-specific resources, such as a local file, must be executed by a Worker on that host.
+
+The question is how do we make sure a specific Worker is executing a specific Task? We do this through "Task routing".
+
+## Task routing with Task Queues
+
+Task Queues are what enable "Task routing". Task routing is the development technique that pairs Tasks with Workers. The basics of it can be broken down into 3 steps:
```
rylandg, line 15:
```suggestion
Task routing is the development technique that pairs Tasks with Workers using Task Queues. At a basic level it consists of 3 steps:
```

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+1. Create a Worker that is registered to handle Workflow and/or Activity Tasks.
+2. Subscribe the Worker to listen to a Task Queue.
```
rylandg, line 18: "Subscribe to listen" is sort of a weird phrasing

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+- A Worker only handles [Activity](docs/learn-activities) and [Workflow](docs/learn-workflows) Tasks that it is registered to handle. This means that a Task must be paired with a Worker that is registered to handle that Task.
+- A Worker runs on a specific host. This means that Activity Tasks that depend on host-specific resources, such as a local file, must be executed by a Worker on that host.
+
+The question is how do we make sure a specific Worker is executing a specific Task? We do this through "Task routing".
+
+## Task routing with Task Queues
+
+Task Queues are what enable "Task routing". Task routing is the development technique that pairs Tasks with Workers. The basics of it can be broken down into 3 steps:
+
+1. Create a Worker that is registered to handle Workflow and/or Activity Tasks.
```
rylandg, line 17: I think this should either be reworded or split into two steps.

1. Create worker
2. Register worker for handle...

You could also just keep Register and drop Create as its pretty heavily implied

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+2. Subscribe the Worker to listen to a Task Queue.
+3. Configure Workflows and Activities to send Tasks to the Task Queue.
```
rylandg, line 19: It might be nice to either reference `ActivityOptions`/`WorkflowOptions` or say "configure activity execution"

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+Let's look at the ["Hello World" sample in Go](https://github.com/temporalio/temporal-go-samples/tree/master/helloworld) to see how this is done.
```
rylandg, line 21: I would combine this into a single paragraph with the next section

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+In the following code snippet, we [create a new Worker](https://github.com/temporalio/go-samples/blob/master/helloworld/worker/main.go#L26) that is subscribed to the "hello-world" Task Queue and is registered to handle the `helloworld.Workflow` and `helloworld.Activity`:
```
rylandg, line 23: It might be nice to say where this code resides. Ie: Starter, worker, workflow, activity etc. Also

"create a new worker and subscribe it", sayings "that is subscribed" implies this is done automatically as opposed to in the below code.

rylandg: also "handle the ...  `Activity` implementations"

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+// Worker is created and subscribed to the "hello-world" Task Queue
+w := worker.New(c, "hello-world", worker.Options{})
+defer w.Stop()
+// Worker is registered to handle the helloworld Workflow
+w.RegisterWorkflow(helloworld.Workflow)
+// Worker is registered to handle the helloworld Activity
+w.RegisterActivity(helloworld.Activity)
+```
+
+In the next code snippet, we [start the `helloworld.Workflow`](https://github.com/temporalio/go-samples/blob/master/helloworld/starter/main.go#L25):
```
rylandg, line 35: Again where would this reside

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+:::caution
+
+If a Worker pulls a Task that it has not been registered to run, an "unknown workflow type" error will occur.
+
+:::
+
+## Task routing for Activity dependencies
+
+The Temporal service will persist Activity results as long as the results are less than 2MB in size and can be serialized for storage (which happens to be true for most Activity results). However, there are some scenarios where an Activity execution will do something that results in data, such as a large file, being stored directly on the host where the Worker is running. It is also possible that the next Activity in the Workflow has a dependency on that data and needs to run on the same host.
```
rylandg, line 61: I just wanted to say that this is an exceptional first sentence.

rylandg: "However, there are some scenarios where an Activity execution will do something that results in data, such as a large file, being stored directly on the host where the Worker is running."

results in data, such as a large file, being directly stored

I don't think data is a large file which makes this weird, the comma after "large file" makes it even weirder.

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+Let's look at a file processing Workflow example where files are downloaded, processed in some way, and then uploaded to some other location. If a large video file is being processed only the name of the file can persist normally. The file itself must persist locally on the Worker host, as it is too big for normal Temporal storage. This means that any future Activities that act on the video file must also take place on the host where the Worker downloaded it.
```
rylandg, line 63: I think it might make sense to frontload this before the actual code etc. Right now this doc is very abstract until this section. I understand its a bit tricky because the problem itself isn't clear until you understand the limit on activities etc.

rylandg: Also "normal Temporal storage" is too generic

rylandg: "processed" with "transcoded"

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+So, how do we ensure that a Worker on the correct host will execute the next Activity that depends on the video file?
```
rylandg, line 65: Combine these two sentences into one paragraph

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+This is accomplished in different ways depending on the SDK.
+
+### Java
+
+Consider the [file processing sample in Java](https://github.com/temporalio/java-samples/tree/master/src/main/java/io/temporal/samples/fileprocessing). The paradigm is to create two Workers. Each will run on the same host and thus have access to the same file system, but one will listen to a global (shared) Task Queue, which it will use to start the Workflow, and the other will listen to its own host-specific (private) Task Queue.
```
rylandg, line 71: This is a very good explanation. It might be worth briefly commenting on how "host-specific (private)" is accomplished (unique host id/mac address etc)

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+WorkerFactory factory = WorkerFactory.newInstance(client);
+// Create a Worker that is subscribed to a shared Task Queue
+final Worker workerForCommonTaskList = factory.newWorker(globalTaskQueue);
+// Register Workflow and Activities
+// Create a Worker that is subscribed to a private TaskQueue
+final Worker workerForHostSpecificTaskList = factory.newWorker(hostSpecifiTaskQueue);
+// Register activities
+```
+
+In the [Workflow](https://github.com/temporalio/java-samples/blob/master/src/main/java/io/temporal/samples/fileprocessing/FileProcessingWorkflowImpl.java#L42), the first Activity Task is executed by the Worker listening to the global Task Queue:
```
rylandg, line 86: I think you could make it clear that the task is executed by that worker because the activity was executed using that tasklist.

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+    ActivityOptions ao =
+        ActivityOptions.newBuilder()
+        // Send the Activity to the shared Task Queue
+        .setTaskQueue(globalTaskQueue)
+        .build();
+    this.defaultTaskQueueStore = Workflow.newActivityStub(StoreActivities.class, ao);
+}
+```
+
+The Activities should be designed to [return an object](https://github.com/temporalio/java-samples/blob/master/src/main/java/io/temporal/samples/fileprocessing/StoreActivities.java) that includes both the Activity result and the name of the host-specific Task Queue to which the next Activity Task will be sent.
```
rylandg, line 100:
```suggestion
The Activities should be designed to [return an object](https://github.com/temporalio/java-samples/blob/master/src/main/java/io/temporal/samples/fileprocessing/StoreActivities.java) that includes both the Activity result and the name of the host-specific Task Queue to which the next Activity Task should be sent.
```

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+```java
+public interface StoreActivities {
+
+  final class ResultAndTaskQueuePair {
+    // Name of the host-specific  Task Qeueu
```
rylandg, line 106: Qeueu

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+    private final String activityResult;
+
+    // Get and set methods ...
+  }
+
+  // Interface methods ....
+}
+```
+
+The host-specific Task Queue is then set on [subsequent Activities](https://github.com/temporalio/java-samples/blob/master/src/main/java/io/temporal/samples/fileprocessing/FileProcessingWorkflowImpl.java#L63):
```
rylandg, line 118: '"set" on subsequent activites'

I think this isn't the most accurate way of phrasing it. Activities aren't really set, the are configured when executed.

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+    // Call the Activity to process the file using worker-specific task queue.
+}
+```
+
+### Go
+
+In Go, you have two options.
+
+**Option 1**: Implement the same design as you would in Java, which is to create two Workers (one listens to the shared Task Qeue and the other listens to the private Task Queue) and return the name of the private Task Queue back to the Workflow as an Activity result.
```
rylandg, line 137: Very concise and easy to read sentence.

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+**Option 2**: Create a Workflow Session which automatically ties any remaining Workflow Activity executions to the Worker that started the Workflow. The [FileProcessing sample in Go](https://github.com/temporalio/temporal-go-samples/tree/master/fileprocessing) is a good example of sessions being used for this purpose.
```
rylandg, line 139: "any remaining Workflow Activity executions "

I think "remaining" is misleading

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+
+The Worker must be [configured to support sessions](https://github.com/temporalio/go-samples/blob/master/fileprocessing/worker/main.go#L27).
```
rylandg, line 141: This should be part of the previous para

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
...
+sessionCtx, err := workflow.CreateSession(ctx, so)
+if err != nil {
+  return err
+}
+defer workflow.CompleteSession(sessionCtx)
+// Execute activities...
+```
+
+The session will ensure that all Activities are executed on the same host.
+
```
rylandg, line 171: It might be nice to write a brief conclusion

docs/route-tasks.md
```diff
@@ -0,0 +1,171 @@
+---
+id: route-tasks
+title: How to route Tasks
+sidebar_label: Route Tasks
+---
+
+Consider the specifications of a Worker:
+- A Worker only handles [Activity](docs/learn-activities) and [Workflow](docs/learn-workflows) Tasks that it is registered to handle. This means that a Task must be paired with a Worker that is registered to handle that Task.
```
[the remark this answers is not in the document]
cullywakelin: Yes, but that is done in the very next section.

docs/route-tasks.md
cullywakelin: My goal with this was to explain what is happening, and then in the following example show how it happens. Will think about this -

docs/route-tasks.md
cullywakelin: Where it resides definitely depends on the SDK language and app design - I would like to experiment with using a "Sample source reference" link that would send the user to the exact code lines in the sample that best reflects the snippet (the actual go samples need to be updated to be better samples for the docs - part of the sync project)

docs/route-tasks.md
```diff
@@ -0,0 +1,238 @@
...
+workflowOptions := client.StartWorkflowOptions{
+  ID:        "hello_world_workflowID",
+  TaskQueue: "hello-world",
+}
+// Workflow Options are passed to the execution of the Workflow
+we, err := c.ExecuteWorkflow(
+  context.Background(), 
+  workflowOptions, 
+  helloworld.Workflow, 
+  "Temporal",
```
mfateev, line 111: This "Temporal" parameter is confusing, especially having that the Java sample one doesn't have it.

docs/route-tasks.md
```diff
@@ -0,0 +1,238 @@
...
+)
+```
+
+</TabItem>
+</Tabs>
+
+:::caution
+
+If a Worker pulls a Task that it has not been registered to execute, an "unknown workflow type" error will occur.
```
mfateev, line 120: We should add that it implies that all workers that listen to a given task queue should have the same set of workflows and activities registered with the corresponding workers.

docs/route-tasks.md
```diff
@@ -0,0 +1,238 @@
...
+## Overview
+
+This guide covers the recommended techniques to ensure that a Workflow and/or Activity will be executed by an appropriate Worker. In this guide we will look at two examples:
+
+1. A basic example that illustrates the fundamentals of routing a Task to a Worker.
+2. A more complex example involving Activities with host-specific file dependencies.
+
+Consider the specifications of a Worker:
+- A Worker only handles [Activity](activities) and [Workflow](workflows) Tasks that it is registered to handle. This means that a Task must be paired with a Worker that is registered to handle that Task.
+- A Worker runs on a specific host. This means that Activity Tasks that depend on host-specific resources, such as a local file, must be executed by a Worker on that host.
```
mfateev, line 16: Actually a worker runs in a specific process. And it is possible for an activity to depend on process-specific resources like in-memory cache.

docs/route-tasks.md
```diff
@@ -0,0 +1,238 @@
...
+</Tabs>
+
+:::caution
+
+If a Worker pulls a Task that it has not been registered to execute, an "unknown workflow type" error will occur.
+
+:::
+
+The above examples illustrate the fundamental approach to pairing Tasks with Workers. In the next section we look at a more complex scenario involving Activities with host-specific file dependencies.
+
```
mfateev, line 125: Host specific task routing is a very important use case. But the more important use case is running different activities and workflows in different services (which are essentially pools of processes). I think we want to have a section on using task queues to route to different service/deployment.

docs/route-tasks.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+### Go
+
+In Go, you have two options.
+
+**Option 1**: Implement the same design as you would in Java, which is to create two Workers (one listens to the shared Task Qeue and the other listens to the private Task Queue) and return the name of the private Task Queue back to the Workflow as an Activity result.
+
+**Option 2**: Create a Workflow Session which automatically ties Activity executions to the Worker that started the Workflow. The [FileProcessing sample in Go](https://github.com/temporalio/temporal-go-samples/tree/master/fileprocessing) is a good example of sessions being used for this purpose. The Worker must be [configured to support sessions](https://github.com/temporalio/go-samples/blob/master/fileprocessing/worker/main.go#L27).
+
```
mfateev, line 205: I think it is worth adding that session allows limiting number of parallel sessions per process.

cullywakelin:
Closing this PR.

PR #59, opened 2020-07-29
New "How to retry Activities and Workflows" guide

cullywakelin:
## Whats in this PR?

- This PR is adding a new file/page.
- The content topic covers Activity and Workflow retries
- This does NOT add the page id to the sidebar. That change will happen in a later pr.

### DIFF
--- a/docs/retry-activities-and-workflows.md
+++ b/docs/retry-activities-and-workflows.md
@@ -1,6 +1,7 @@
 ---
-id: retrying-timeouts
-title: Retrying and timeouts
+id: retry-activities-and-workflows
+title: How to retry Activities and Workflows
+sidebar_label: Retry Activities and Workflows
 ---
 
 Since Activities are used to encapsulate logic that is non-deterministic, an important Workflow feature is the ability to "automatically" retry an Activity execution in the event of its failure.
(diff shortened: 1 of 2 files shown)


docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
+---
+id: retry-activities-and-workflows
+title: How to retry Activities and Workflows
+sidebar_label: Retry Activities and Workflows
+---
+
+Temporal is built to expect failures from Activities and Workflows. Retrying Activity or Workflow executions is part of what makes the Temporal system fault tolerant.
```
rylandg, line 7: I think the first sentence is definitely misleading but I also incorrect (not 100% sure). 

'Retrying Activity or Workflow executions is part of what makes the Temporal system fault tolerant.' - not really sure this is true

I would say timeouts and retries are a reliable way of handling potential failures. Workflows themselves cannot really fail.

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+
+In this guide we cover the difference between Activity and Workflow retries and how to use them.
+
+## Activity retries
+
+The nature of an [Activity](docs/activities) is to encapsulate logic that is non-deterministic. With Temporal, the paradigm is to not explicitly handle errors and retries within the Activity logic. Instead, the Activity should return the error directly to the Workflow which will trigger a retry per the Activity's retry policy. Each Activity can have its own retry policy, can use a shared one, or not specify one at all and use Temporal's default retry policy. Said another way, if `RetryOptions` are not specified within the `ActivityOptions` then the Activity will be retried per Temporal's default configuration.
```
rylandg, line 13:
```suggestion
[Activities](docs/activities) are intended to encapsulate logic that is non-deterministic. With Temporal, the paradigm is to not explicitly handle errors and retries within the Activity logic. Instead, the Activity should return the error directly to the Workflow which will trigger a retry per the Activity's retry policy. An Activity can either specify:

* its own retry policy
* a shared one
* nothing, which uses the default retry policy. 

Said another way, if `RetryOptions` are not specified within the `ActivityOptions` then the Activity will be retried per Temporal's default configuration.
```

rylandg: I think the second and third sentences could be reworded to be way easier to read.

Also maybe add the default RetryOptions

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+
+It is possible to explicitly handle erorrs and retries within the Activity code. However, one of the benefits of using Temporal's retry feature is that it removes the need to repetitively write the same boilerplate error handling and retry code around every piece of business logic.
```
rylandg, line 15: "It is possible to explicitly handle erorrs and retries within the Activity code." - would remove this as we don't want to encourage it at all.

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Retry policy options
+
+The following field values can be customized within the `RetryOptions` of any `ActivityOptions`:
```
rylandg, line 19: Maybe link to the source?

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+
+```java
+private final RemoteDataActivities activities =
+    Workflow.newActivityStub(
+        RemoteDataActivity.class,
+        ActivityOptions.newBuilder()
+        .setScheduleToCloseTimeout(Duration.ofSeconds(10))
+        .setRetryOptions(
+            RetryOptions.newBuilder()
+            // Here we specify a custom InitialInterval
```
rylandg, line 112: This code comment does not provide value.

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+            .setInitialInterval(Duration.ofSeconds(2))
+            // Here we specify a custom BackoffCoefficient
```
rylandg, line 114: This code comment does not provide value.

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+        .build());
+```
+
+</TabItem>
+<TabItem value="go">
+
+```go
+ao := workflow.ActivityOptions{
+    RetryPolicy: &temporal.RetryPolicy{
+        // Here we specify a custom InitialInternal backoff
```
rylandg, line 126: This code comment does not provide value.

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+        InitialInterval:    time.Second * 2,
+        // Here we specify a custom BackoffCoefficient
```
rylandg, line 128: This code comment does not provide value.

docs/retry-activities-and-workflows.md
```diff
@@ -0,0 +1,185 @@
...
+}
+ctx = workflow.WithActivityOptions(ctx, ao)
+```
+
+</TabItem>
+</Tabs>
+
+## Workflow retries
+
+While retrying Activities is enough for the majority of use cases, Temporal also provides a mechanism for retrying an entire Workflow. Workflow level retries can be very useful, such as when your Workflow consists of multiple Activities that rely on a host-specific state. Unlike Activities, Workflows are not retried by default, they must be explicitly configured to occur.
```
rylandg, line 140: Two things:

1. link to task routing
2. they must be explicitly configured to occur. - don't like the word "occur"

cullywakelin:
Closing this PR.

PR #60, opened 2020-07-29
New "How to timeout Activities and Workflows" guide

cullywakelin:
## What is in this PR?

- 1 new file/page
- page topic covers timeouts for Workflows and Activities 
- The page id has NOT been added to the sidebar in this pr, that will happen in a later pr

### DIFF
--- a/docs/timeout-activities-and-workflows.md
+++ b/docs/timeout-activities-and-workflows.md
@@ -1,0 +1,67 @@
+---
+id: timeout-activities-and-workflows
+title: Timeout Activities and Workflows
+---
+
+## Activity Timeouts
+
+The default behavior of Temporal is to not impose any restrictions/limits on the duration of activities. This makes sense as there are many use cases which depend on long running or indefinitely running activities. Even though the default behavior is to let activities run indefinitely, you'll often want to limit the length that a certain activity can run. Piggybacking on the example we used for the []Retrying Activities section, let's make a new assumption that the call to our unreliable HTTP server has a potential to hang for minutes or even hours. Based on this new assumption, it makes a lot of sense to set an upper bound on the duration of our `fetchRemoteData` activity:
+
+```diff
+    private final RemoteDataActivities activities =
+        Workflow.newActivityStub(
+            RemoteDataActivity.class,
+            ActivityOptions.newBuilder()
++                .setScheduleToStartTimeout(Duration.ofMinutes(1))
+							   .setScheduleToCloseTimeout(Duration.ofMinutes(20))
++                .setStartToCloseTimeout(Duration.ofMinutes(10))
+            .build());
+```
+
+Since it's not intuitive what effect each of those methods has on activity behavior, I've explained them below:
+
+- `ScheduleToStart` - duration a workflow should wait for a worker to start executing a requested activity. A common use for this timeout is checking if all workers are down or not able to keep up at the current request rate.
+- `StartToClose` - maximum time an activity can be executed after it was picked by a worker.
+- `ScheduleToClose` - maximum time from the workflow requesting an activity execution to its completion including retries.
+
+Let's apply these definitions to the code snippet above so we get a full picture of what's practically going to happen.
+
+1. We set `ScheduleToStart` to a duration of 1 minute. This means that our workflow will wait 1 minute at most for a worker to pick up this activity. If that 1 minute is exceeded the activity will either be retried or an error will be surfaced to the worker. In the majority of cases it makes sense to set this to the expected run duration of the activity.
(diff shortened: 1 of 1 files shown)


docs/timeout-activities-and-workflows.md
```diff
@@ -0,0 +1,119 @@
...
+
+Activity timeouts provide additional control over the way your Workflows execute Activities.
+
+### Timeout policy options
+
+The following fields values can be customized within the `ActivityOptions`:
+
+| Field | Java Type | Go Type | Description | Default |
+|-------|-----------|---------|-------------|---------|
+| `ScheduleToStart` | Duration | time.Time | Maximum time a Workflow should wait for a Worker to start executing the Activity. This is commonly used for checking if all workers are down or are not able to keep up with current request rate. | TODO |
```
mfateev, line 19: I would say that in some rare cases it is useful for a workflow to react when workers are down. One such use case is task routing to specific hosts. In this case, if the host is down it is not expected that it is going to come back (at least in K8s), so setting this timeout to a low value makes sense. But in the majority of situations, this timeout should not be set as it is expected that worker will come back and pick up a task.

docs/timeout-activities-and-workflows.md
```diff
@@ -0,0 +1,119 @@
...
+sidebar_label: Timeout Activities and Workflows
+---
+
+Temporal does not impose limits on the duration of Activities or Workflows by default, it is designed to enable these to run idefinitely if desired. Instead, timeout options are available to set limits on the time allowed for certain actions to occur.
+
+## Activity timeouts
+
+Activity timeouts provide additional control over the way your Workflows execute Activities.
+
+### Timeout policy options
```
mfateev, line 13: We need to explain which of these timeouts are required. As far as I remember only one timeout is required and it should be either StartToClose or ScheduleToClose. 

Defaults are:
If ScheduleToClose is specified:

- ScheduleToStart defaults to ScheduleToClose
- StartToClose defaults to ScheduleToClose

If only StartToClose is specified both ScheduleToStart and ScheduleToClose default to WorkflowRun timeout.

docs/timeout-activities-and-workflows.md
```diff
@@ -0,0 +1,119 @@
...
+</Tabs>
+
+In the above code snippets:
+
+1. We set `ScheduleToStart` to a duration of 1 minute. This means that our Workflow will wait a maximum of 1 minute for a Worker to pick up this Activity. If it is exceeded the Activity will either be retried or an error will be surfaced to the Worker. In the majority of cases it makes sense to set this to the expected run duration of the Activity.
+2. We set `ScheduleToClose` to a duration of 1 minute. This means that our Workflow will wait a maximum of 1 minute for an Activity execution to complete starting from when the Task was requested to completion, including retries. If it is exceeded an error will be surfaced to the Worker.
+3. We set `StartToCloseTimeout` to a duration of 10 minutes. This means that our Workflow will wait a maximum of 10 minutes for a Worker that has picked up the Activity to finish executing it. If it is exceeded the Activity will either be retried or an error will be surfaced to the Worker.
+
+## Workflow timeouts
+
```
mfateev, line 67: I would add:

Workflow timeouts are much less useful than activity timeouts. The reason is that Temporal doesn't preserve the state of activities (with exception of heartbeat details) and the only way to know that an activity worker failed is to rely on timeouts. For workflows Temporal preserves their state in case of different infrastructure failures. So there is no need for timing them out due to various infrastructure failures. 

Don't use the workflow timeouts for business level timeouts. Use timers instead.

docs/timeout-activities-and-workflows.md
```diff
@@ -0,0 +1,119 @@
...
+Activity timeouts are great when you need to impose time restrictions on a specific Activity, but there's often a need to impose similar restrictions on a workflow execution. The code snippet below shows all the timeout properties that are available to workflows.
+
+### Timeout policy options
+
+The following field values can be customized within the `WorkflowOptions`:
+
+| Field | Java Type | Go Type | Description | Default |
+|-------|-----------|---------|-------------|---------|
+| `WorkflowExecutionTimeout` | Duration | time.Time | Maximum time for the Workflow Execution including retries and "ContinueAsNew"". Use WorkflowRunTimeout to limit execution time of a single workflow run. | TODO |
+| `WorkflowTaskTimeout`      | Duration | time.Time | Maximum time from when a Worker pulls the task. If a Workflow Task is lost, it is retried after this timeout. | TODO |
```
mfateev, line 77: WorkflowTaskTimeout defines how long the workflow is going to get "stuck" in the case of a worker crash. So setting it to a large value is not a good idea. Setting it too low (below 10 seconds) is not usually recommended as it can lead to unnecessary timeouts which might increase the load on the service and worker processes.

cullywakelin:
Closing this PR.

PR #72, opened 2020-08-01
Fixes up first few steps of Java SDK tutorial that had incorrect access modifiers and import statements

mastermanu:


mastermanu:
abandoning as Ryland is sending a separate one

PR #76, opened 2020-08-04
Added content to get-started page

cullywakelin:
This PR adds "content/text" to get-started.md.
No other changes are made in this PR.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -15,7 +15,7 @@ Ready to build? [Install the Temporal server](install-temporal-server) and choos
 
 ## Learn about Temporal
 
-Applications built with Temporal are extremely reliable, but using it can be a paradigm shift for some folks. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](overview).
+Temporal enables developers to write highly reliable applications without having to worry about all the edge cases, but getting started can be paradigm shift for some. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](overview).
 
 Wondering if Temporal is a good fit for your use case? Check out our [use cases section](use-cases-orchestration) or visit the [community forum](https://community.temporal.io/tag/use-case-validation).
 
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -4,31 +4,29 @@ title: Temporal documentation
 sidebar_label: Get started
 ---
 
-Welcome to Temporal's user documentation!
-
 Temporal reimagines state-dependent service-orchestrated application development, and so much more.
 
 ## Get started
 
-Ready to develop? [Install the Temporal server](install-temporal-server) and choose your SDK.
+Ready to build? [Install the Temporal server](install-temporal-server) and choose your SDK.
 
-→ [Go SDK](go-quick-start)
-
-→ [Java SDK](java-quick-start)
+- [Go SDK](go-quick-start)
+- [Java SDK](java-quick-start)
 
 ## Learn about Temporal
 
-Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview) and common [use cases](use-cases-overview).
+Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview).
 
-Run the code samples to see how it works:
+Wondering if Temporal is a good fit for your use case? Check out our [documented use cases](use-cases-overview) or visit the [community forum](https://community.temporal.io/tag/use-case-validation).
 
-→ [Go SDK samples](https://github.com/temporalio/go-samples)
+Prefer to jump right in? Try some of our tested examples.
 
-→ [Java SDK samples](https://github.com/temporalio/java-samples)
+- [Go SDK samples](https://github.com/temporalio/go-samples)
+- [Java SDK samples](https://github.com/temporalio/java-samples)
 
 ## Best practices
 
-Want to master Temporal Workflows and recommended development techniques? Our guides will show you how.
+Want to master Temporal Workflows? Our guides will break down some of the recommended development techniques.
 
 - [How to install Temporal](install-temporal-server)
 - [How to configure Temporal](configure-temporal-server)
@@ -37,7 +35,7 @@ Want to master Temporal Workflows and recommended development techniques? Our gu
 
 ## Contribute
 
-Want to contribute? Follow the contribution instructions:
+Want to contribute? Follow the contribution instructions provided within the repositories.
 
 - [Temporal server](https://github.com/temporalio/temporal/blob/master/CONTRIBUTING.md)
 - [Documentation](https://github.com/temporalio/documentation-legacy/blob/master/README.md).
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -15,18 +15,18 @@ Ready to build? [Install the Temporal server](install-temporal-server) and choos
 
 ## Learn about Temporal
 
-Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview).
+Applications built with Temporal are extremely reliable, but using it can be a paradigm shift for some folks. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](overview).
 
-Wondering if Temporal is a good fit for your use case? Check out our [documented use cases](use-cases-overview) or visit the [community forum](https://community.temporal.io/tag/use-case-validation).
+Wondering if Temporal is a good fit for your use case? Check out our [use cases section](use-cases-orchestration) or visit the [community forum](https://community.temporal.io/tag/use-case-validation).
 
-Prefer to jump right in? Try some of our tested examples.
+Prefer to jump right in? Try out some of our examples.
 
 - [Go SDK samples](https://github.com/temporalio/go-samples)
 - [Java SDK samples](https://github.com/temporalio/java-samples)
 
 ## Best practices
 
-Want to master Temporal Workflows? Our guides will break down some of the recommended development techniques.
+Want to master Temporal Workflows? Our guides will bring you up to speed on Temporal development best practices.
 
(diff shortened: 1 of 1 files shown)


docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
 sidebar_label: Get started
 ---
 
+Welcome to Temporal's user documentation!
+
 Temporal reimagines state-dependent service-orchestrated application development, and so much more.
```
rylandg, line 9: I know this was already here but I want to say it better.

docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
 sidebar_label: Get started
 ---
 
+Welcome to Temporal's user documentation!
```
rylandg, line 7: I don't think we need this. I don't think people will think we're more genuine for it but I could be wrong. Also "user documentation" sounds demeaning or something.

docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
...
+
 Temporal reimagines state-dependent service-orchestrated application development, and so much more.
 
 ## Get started
 
-This is page will be dedicated to directing users on how to get started with Temporal.
+Ready to develop? [Install the Temporal server](install-temporal-server) and choose your SDK.
```
rylandg, line 13: I like the tone change A LOT. I'm not necessarily saying we should change the word "develop" but not sure its what we want to communicate.

rylandg: Very unsure here, just my initial thought.

docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
...
+
+→ [Go SDK](go-quick-start)
```
rylandg, line 15: So visually speaking the arrows are much more aesthetic. That being said, we also use bullets on the page and I would like to establish when we will use bullets vs arrows.

docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
...
 
-This is placeholder content.
+→ [Java SDK](java-quick-start)
 
 ## Learn about Temporal
 
-etc...
+Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview) and common [use cases](use-cases-overview).
```
rylandg, line 21: The first sentence:

"Building applications with Temporal is a paradigm shift. "

should definitely be removed, we want to avoid communicating things in a negative way. Obviously what you said is 100% true. If you want to communicate the same message with a positive spin consider something like:

"Temporal makes it a breeze to write reliable applications. Problems that developers are traditionally expected to solve themselves are handled implicitly by the system. While Temporal brings some huge advantages to developing applications, for some it requires a little getting used to..."

Or something

rylandg: Also maybe split out the use case part. into a sentence.

"Wondering if your use case is a good fit for Temporal?" And then link use cases and maybe a forum link if the use case isn't covered in our docs

docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
...
+
+Run the code samples to see how it works:
```
rylandg, line 23: If you prefer to jump right in, here are some samples to get you started:

docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
...
+
+→ [Go SDK samples](https://github.com/temporalio/go-samples)
+
+→ [Java SDK samples](https://github.com/temporalio/java-samples)
+
+## Best practices
+
+Want to master Temporal Workflows and recommended development techniques? Our guides will show you how.
```
rylandg, line 31: I like "master Temporal Workflows" but don't like how you've structured the sentence to have "master" apply to "recommend development techniques". While I guess it's true that you could master "recommended development techniques" I don't think this is something people aspire to.

docs/get-started.md
```diff
@@ -4,14 +4,43 @@ title: Temporal documentation
...
+
+- [How to install Temporal](install-temporal-server)
```
rylandg, line 33: Wow didn't know links work without any path "./". I'm not suggesting this is broken it's a genuine observation.

cullywakelin: Chayup. https://v2.docusaurus.io/docs/markdown-features#referencing-other-documents

docs/get-started.md
cullywakelin: What about "Ready to build?"

docs/get-started.md
cullywakelin: I too prefer the arrows - but wanted to see what you thought first.
As of right now, arrows need to be manually added via a copy and paste into the doc. They do not automatically format into lists - 
So, as a long term solution what we will want to do is update the custom css to make markdown `<li>` tags render as those arrows vs the bullets. I will introduce that change in another pr and convert all of these to regular list items for now.

docs/get-started.md
cullywakelin: I wrote rewrote this line from a number of sources. It wasn't copied or carried over from someplace.
Improving it, I think that is the hardest part of our job right now :) Not sure how to capture the essence in just 1-2 sentences yet. Unless you have an alternative suggestion, to keep things succinct I am going to push for keeping this as-is for now - because in its current form it is at least accurate and all encompassing - "so much more" :)

docs/get-started.md
cullywakelin: > The first sentence:
> 
> "Building applications with Temporal is a paradigm shift. "
> 
> should definitely be removed, we want to avoid communicating things in a negative way. Obviously what you said is 100% true. If you want to communicate the same message with a positive spin consider something like:
> 
> "Temporal makes it a breeze to write reliable applications. Problems that developers are traditionally expected to solve themselves are handled implicitly by the system. While Temporal brings some huge advantages to developing applications, for some it requires a little getting used to..."
> 
> Or something

I disagree. There is nothing negative about that statement. It is meant to be very to the point - otherwise we are creating confusing messaging within our docs - If we are talking about our homepage, then yes, I would agree with you. But in our docs, I think this type of clarity is very important. I think this strong, clear message helps to set expectations for our users when they are deciding which docs to read first.

cullywakelin: > Also maybe split out the use case part. into a sentence.
> 
> "Wondering if your use case is a good fit for Temporal?" And then link use cases and maybe a forum link if the use case isn't covered in our docs

I agree with this -

docs/get-started.md
cullywakelin: Chayup: https://v2.docusaurus.io/docs/markdown-features#referencing-other-documents

cullywakelin:
> The only bigger feedback I had was my gut telling me we should have the literal install instructions (not just a link to the page) on the getting started page. It would take up a lot of space and no other big doc sites seem to do it. Again, probably disregard but was just my nagging thought

Lets sit on this -
My plan was to overhaul the install guide - it is currently lacking a lot of context ... but if it ends up being short, then we can look at merging.

docs/get-started.md
```diff
@@ -4,31 +4,29 @@ title: Temporal documentation
...
+Ready to build? [Install the Temporal server](install-temporal-server) and choose your SDK.
 
-→ [Java SDK](java-quick-start)
+- [Go SDK](go-quick-start)
+- [Java SDK](java-quick-start)
 
 ## Learn about Temporal
 
-Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview) and common [use cases](use-cases-overview).
+Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview).
```
rylandg, line 18: Still not a fan of starting out with "paradigm shift". Sound daunting. Start with the positive about how we will make it so much easier etc.

docs/get-started.md
```diff
@@ -4,31 +4,29 @@ title: Temporal documentation
...
 
-Run the code samples to see how it works:
+Wondering if Temporal is a good fit for your use case? Check out our [documented use cases](use-cases-overview) or visit the [community forum](https://community.temporal.io/tag/use-case-validation).
 
-→ [Go SDK samples](https://github.com/temporalio/go-samples)
+Prefer to jump right in? Try some of our tested examples.
```
rylandg, line 22: The word "tested" is weird here.

docs/get-started.md
```diff
@@ -4,31 +4,29 @@ title: Temporal documentation
...
 
-→ [Java SDK samples](https://github.com/temporalio/java-samples)
+- [Go SDK samples](https://github.com/temporalio/go-samples)
+- [Java SDK samples](https://github.com/temporalio/java-samples)
 
 ## Best practices
 
-Want to master Temporal Workflows and recommended development techniques? Our guides will show you how.
+Want to master Temporal Workflows? Our guides will break down some of the recommended development techniques.
```
rylandg, line 29:
```suggestion
Want to master Temporal Workflows? Our guides will bring you up to speed on Temporal development best practices.
```

docs/get-started.md
```diff
@@ -4,31 +4,29 @@ title: Temporal documentation
...
+- [Go SDK](go-quick-start)
+- [Java SDK](java-quick-start)
 
 ## Learn about Temporal
 
-Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview) and common [use cases](use-cases-overview).
+Building applications with Temporal is a paradigm shift. If you are new to Temporal, or workflow applications in general, get caught up on the [core concepts](concepts-overview).
 
-Run the code samples to see how it works:
+Wondering if Temporal is a good fit for your use case? Check out our [documented use cases](use-cases-overview) or visit the [community forum](https://community.temporal.io/tag/use-case-validation).
```
rylandg, line 20:
```suggestion
Wondering if Temporal is a good fit for your use case? Check out our [use cases section](use-cases-overview) or visit the [community forum](https://community.temporal.io/tag/use-case-validation).
```

PR #78, opened 2020-08-08
Hotjar script

cullywakelin:
This PR adds a hotjar script to the docs site for two purposes:
1. Heatmapping
2. Feedback widget

Screen Shot 2020-08-07 at 6 01 35 PM


cullywakelin:
> Two things:
> 
> 1. Would like to understand how much this effects page load speed (if any)

Should be fairly minimal now - it does an async load.

> 2. It doesn't work (at least on the demo link). In chrome dev tools network tab it indicates that "scripts/feedback.js" 404d because it uses a relative path. For example I was on the events page of the docs and the console says the feedback script was requested from:
> 
> https://deploy-preview-78--mystifying-fermi-1bc096.netlify.app/docs/next/events/scripts/feedback.js

The build path was wrong. I fixed this - and it should now be visible in the preview.
However, there appears to be a render issue if:
1. some sort of ad blocker is in use.
2. a private or incognito tab is in use.
I have put in a ticket with Hotjar support for help with debugging this -

PR #80, opened 2020-08-08
Upgrade docusaurus v2.alpha.61

feedmeapples:
`yarn build` command now warns about relative broken links.
The behavior is possible to change, for example to throw error on broken links. Modify `docusaurus.config.js -> onBrokenLinks` to control that

### DIFF
--- a/docs/learn-glossary.md
+++ b/docs/learn-glossary.md
@@ -90,7 +90,7 @@ Temporal is backed by a multi-tenant service and the unit of isolation is called
 - By default a Temporal service is provisioned with a "default" Namespace. All APIs and tools, such as the UI and CLI, default to the "default" Namespace if it is not specified. So, if you are not planning to use multiple Namespaces, we recommend using the default one.
 - [**Task Queue**](#task-queue) names as well as [**Workflow Ids**](#workflow-id) correspond to a specific Namespace. For example, when a Workflow is started, it is started within a specific Namespace.
 - Temporal guarantees a unique [**Workflow Id**](#workflow-id) within a Namespace, and supports running [**Workflow Executions**](#workflow-execution) to use the same [**Workflow Id**](#workflow-id) if they are in different Namespaces.
-- Various configuration options like the retention period or Archival destination are configured per Namespace as well through a special CRUD API or through [`tctl`](./learn-cli).
+- Various configuration options like the retention period or Archival destination are configured per Namespace as well through a special CRUD API or through [`tctl`](/docs/learn-cli).
 - In a multi-cluster deployment, Namespace is a unit of fail-over.
 - Each Namespace can only be active on a single Temporal cluster at a time. However, different Namespaces can be active in different clusters and can fail-over independently.
 
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -5,7 +5,7 @@ module.exports = {
   tagline: 'Invincible applications, invisible infrastructure',
   url: 'https://docs.temporal.io',
   baseUrl: '/',
-  onBrokenLinks: 'throw',
+  onBrokenLinks: 'warn',
   favicon: 'img/favicon.svg',
   organizationName: 'temporalio', // Usually your GitHub org/user name.
   projectName: 'temporal-documentation-legacy-v2', // Usually your repo name.
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -5,6 +5,7 @@ module.exports = {
   tagline: 'Invincible applications, invisible infrastructure',
   url: 'https://docs.temporal.io',
   baseUrl: '/',
+  onBrokenLinks: 'throw',
   favicon: 'img/favicon.svg',
   organizationName: 'temporalio', // Usually your GitHub org/user name.
   projectName: 'temporal-documentation-legacy-v2', // Usually your repo name.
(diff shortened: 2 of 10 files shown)


docs/learn-glossary.md
```diff
@@ -90,7 +90,7 @@ Temporal is backed by a multi-tenant service and the unit of isolation is called
 - By default a Temporal service is provisioned with a "default" Namespace. All APIs and tools, such as the UI and CLI, default to the "default" Namespace if it is not specified. So, if you are not planning to use multiple Namespaces, we recommend using the default one.
 - [**Task Queue**](#task-queue) names as well as [**Workflow Ids**](#workflow-id) correspond to a specific Namespace. For example, when a Workflow is started, it is started within a specific Namespace.
 - Temporal guarantees a unique [**Workflow Id**](#workflow-id) within a Namespace, and supports running [**Workflow Executions**](#workflow-execution) to use the same [**Workflow Id**](#workflow-id) if they are in different Namespaces.
-- Various configuration options like the retention period or Archival destination are configured per Namespace as well through a special CRUD API or through [`tctl`](./learn-cli).
+- Various configuration options like the retention period or Archival destination are configured per Namespace as well through a special CRUD API or through [`tctl`](/docs/learn-cli).
```
cullywakelin, line 93: Can also be changed to just [`tctl`](learn-cli) - by linking to the page id - 
Docusaurus docs are not clear about that https://v2.docusaurus.io/docs/markdown-features#referencing-other-documents but you can do that too - just an FYI - as that is how I have been doing it.

feedmeapples: @cullywakelin initially we used _linking to page id_, however there was/possibly still is a docusaurus bug #5 so we switched to relative paths

feedmeapples:
> Netlify build breaks.

addressed

docusaurus.config.js
```diff
@@ -5,20 +5,10 @@ module.exports = {
   tagline: 'Invincible applications, invisible infrastructure',
   url: 'https://docs.temporal.io',
   baseUrl: '/',
+  onBrokenLinks: 'warn',
```
rylandg, line 8: Why is this warn?

docusaurus.config.js
```diff
@@ -5,20 +5,10 @@ module.exports = {
...
   favicon: 'img/favicon.svg',
   organizationName: 'temporalio', // Usually your GitHub org/user name.
   projectName: 'temporal-documentation-legacy-v2', // Usually your repo name.
-  plugins: [
-    '@docusaurus/plugin-google-gtag',
-    [
-      '@docusaurus/plugin-sitemap',
```
rylandg, line 14: Why is sitemap removed?

docusaurus.config.js
```diff
@@ -5,20 +5,10 @@ module.exports = {
   tagline: 'Invincible applications, invisible infrastructure',
   url: 'https://docs.temporal.io',
   baseUrl: '/',
+  onBrokenLinks: 'warn',
   favicon: 'img/favicon.svg',
   organizationName: 'temporalio', // Usually your GitHub org/user name.
   projectName: 'temporal-documentation-legacy-v2', // Usually your repo name.
```
rylandg, line 11: This is no longer current name

docusaurus.config.js
```diff
@@ -174,5 +164,4 @@ module.exports = {
       },
     ],
   ],
-  themes: ['@docusaurus/theme-live-codeblock'],
```
rylandg, line 177: Why was this removed

src/pages/index.js
```diff
@@ -58,34 +58,34 @@ function Feature({ imageUrl, title, description }) {
...
+              to={useBaseUrl('docs/')}
             >
               Get Started
             </Link>
           </div>
         </div>
       </header>
       <main>
-        {features && features.length && (
+        {features && features.length > 0 && (
```
rylandg, line 85: this could be `features?.length &&`

docusaurus.config.js
feedmeapples: it is already part of @docusaurus/preset-classic according to https://v2.docusaurus.io/docs/presets/#docusauruspreset-classic
Starting the recent versions of docusaurus, it has started warning at build time that this plugin is set up twice so i removed the second declaration. 

To address the comment, i've pushed a new commit that adds explicit settings for sitemap plugin + comments for clarity https://github.com/temporalio/documentation/pull/80/commits/852d5952b2ea1ad059577bb818b3bf22d86f5c68

docusaurus.config.js
feedmeapples: this was added before to address an issue that code blocks in light mode would have the "Copy" button almost invisible. With this change the Copy button started showing up properly. (P.s it's not what the plugin itself is meant to do, it's actually a plugin that allows to modify the code in code blocks and changes some styling)

In newer versions of docusaurus the copy button is showing up fine out of the box, so i cleaned the code up
image

src/pages/index.js
feedmeapples: that's an out of the box code and we don't show this page

docusaurus.config.js
feedmeapples: > Is darkmode not default anymore?
> 
> Went and checked llinks seemed to not break anything
> 
> This came up in lighthouse. Not sure if its only because the link is for preview.
> 
> Screen Shot 2020-08-12 at 10 58 49 AM

Dark mode was never explicitly set to be the default. I'm sending a new PR to set is as default (as per our offline discussion)

Indexing is disabled by Netlify because it's a preview

PR #84, opened 2020-08-12
Set dark mode as default

feedmeapples:


docusaurus.config.js
```diff
@@ -10,6 +10,10 @@ module.exports = {
   organizationName: 'temporalio', // Usually your GitHub org/user name.
   projectName: 'temporal-documentation-legacy-v2', // Usually your repo name.
   themeConfig: {
+    colorMode: {
+      // "light" | "dark"
+      defaultMode: 'dark',
```
cullywakelin, line 15: TBH I prefer the light mode... @rylandg you think we need to default to dark?

PR #87, opened 2020-08-13
Have sample code call set HostPort explicitly

markmark206:
Every dev playing with this code will likely wonder "how do I tell this thing where my server is running?", and this tweak provides an "at-a-glance" answer.

(I tested that this code compiles and runs, by first copy-pasting this from a compiling / working app, and then by copying this code into a sample that I then compiled and ran.)

### DIFF
--- a/docs/go-quick-start.md
+++ b/docs/go-quick-start.md
@@ -141,7 +141,8 @@ import (
 
 func main() {
 	// The client is a heavyweight object that should be created once
-	serviceClient, err := client.NewClient(client.Options{})
+	clientOptions := client.Options{HostPort: "localhost:7233"}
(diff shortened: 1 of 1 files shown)


docs/go-quick-start.md
```diff
@@ -141,7 +141,8 @@ import (
 
 func main() {
 	// The client is a heavyweight object that should be created once
-	serviceClient, err := client.NewClient(client.Options{})
+	clientOptions := client.Options{HostPort: "localhost:7233"}
```
cullywakelin, line 144: Is there a Go sample that does this?
We will want to merge the Go sample into the doc page eventually - curious if it exists in the sample yet or if this is the only place? I checked a few samples but didn't see any -

PR #88, opened 2020-08-13
Support netlify links validation plugin

feedmeapples:
## What
For links, adds slash symbol `/` to the end of internal doc urls 
Adds slash before the hash symbol `#` if there is any

Before `[here](/docs/filter-workflows#memo-vs-search-attributes)`
After `[here](/docs/filter-workflows/#memo-vs-search-attributes)`

## Why
Netlify links validation plugin expects links to have a slash `/` before the hash part of urls.

## How
Regex used to update `.md` links:
```(\[[\w\s]+\]\(.*\/docs\/[\w-]*)(#?)([\w-]*\))```
Replace with
`$1/$2$3`

### DIFF
--- a/blog/2020-07-31-temporal-v0.28.0-changelog.md
+++ b/blog/2020-07-31-temporal-v0.28.0-changelog.md
@@ -1,8 +1,8 @@
 ---
 tags:
-- changelog
-- temporal
-- v1
+  - changelog
+  - temporal
+  - v1
 posted_on_: 2020-07-31T00:00:00
 id: temporal-v0.28.0-changelog
 title: Temporal v0.28.0 changelog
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: blog/2020-07-31-temporal-v0.28.0-changelog.md [hunk 1]
```diff
@@ -22,69 +22,69 @@ This changelog includes all of the changes since Temporal forked from Cadence, a
...
 - **Activity retry refactors** 🔁 :
-    - Temporal includes default `RetryOptions` for Activities.
-        - `initialInterval` = 1 second.
-        - `maximumInterval` = 100 * initialInterval.
-        - `backoffCoefficient` = 2.
-    - To disable Activity retries, configure Activities with `RetryOptions` that set `maximumAttempts` to 1.
-    - `ScheduleToClose` is used to stop retries for Activities.
+  - Temporal includes default `RetryOptions` for Activities.
+    - `initialInterval` = 1 second.
+    - `maximumInterval` = 100 \* initialInterval.
```
cullywakelin, line 56: Revert this -

docs/get-started.md
```diff
@@ -28,9 +28,9 @@ Prefer to jump right in? Try out some of our examples.
 
 Want to master Temporal Workflows? Our guides will bring you up to speed on Temporal development best practices.
 
-- [How to install Temporal](/docs/install-temporal-server)
-- [How to configure Temporal](/docs/configure-temporal-server)
-- [How to filter Workflows](/docs/filter-workflows)
+- [How to install Temporal](/docs/install-temporal-server/)
+- [How to configure Temporal](/docs/configure-temporal-server/)
+- [How to filter Workflows](/docs/filter-workflows/)
```
cullywakelin, line 33: Missing below (line 34): /docs/tctl -> /docs/tctl/

cullywakelin: Not sure why, but I couldn't post a comment below line 34.... Github was forcing me to add the comment to the old diff when I did that. So I had to put it above...

cullywakelin:
Perhaps its not necessary - but right now, when accessing a page from the sidebar, the "/" is not added at that point either.

docs/get-started.md
feedmeapples: good catch, fixed

PR #89, opened 2020-08-17
Set prettier extension in vscode as recommended

feedmeapples:
Opening project in VSCode will also show a popup recommending to install Prettier.
image

This tool helps to format markdown documents. 
To use it just save a file and Prettier will automatically format the document

Example:

image


cullywakelin:
OK. This is dope.

PR #106, opened 2020-09-14
Go SDK "Hello World!" Tutorial

cullywakelin:
You can view the tutorial at this URL: https://deploy-preview-106--mystifying-fermi-1bc096.netlify.app/docs/go-helloworld/

### DIFF
--- a/docs/go-helloworld.md
+++ b/docs/go-helloworld.md
@@ -1,29 +1,38 @@
 ---
 id: go-helloworld
-title: Go SDK hello world tutorial
+title: Go SDK "Hello World!" tutorial
 sidebar_label: Go hello world
 ---
 
-By the end of this tutorial you will have a simple hello world Temporal Workflow running on your local machine. This tutorial is meant for developers who are new to Temporal or who wish to make sure their environment is working. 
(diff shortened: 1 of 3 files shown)


docs/go-helloworld.md
```diff
@@ -0,0 +1,170 @@
...
+mkdir starter
+touch starter/main.go
+```
+
+Copy the following code and paste it into starter/main.go, making sure to replace the import path `"github.com/temporalio/temporal-go-samples/helloworld"` with your Go module import path:
+
+<!--SNIPSTART go-helloworld-sample-workflow-starter-->
+<!--SNIPEND-->
+
+In the code above, we are creating a Temporal client and using it to host the Workflow with the Temporal server. Said another way, this starter program is sending the Workflow to the Temporal server where it will be broken into Tasks. The Tasks will be sent to the Task Queue that we supplied in the Workflow options.
```
mfateev, line 96: I think "host workflow" is not really a good term in this case.

mfateev: > Said another way, this starter program is sending the Workflow to the Temporal server where it will be broken into Tasks. The Tasks will be sent to the Task Queue that we supplied in the Workflow options.

This sentence contains too many new terms. Workflow is not broken into tasks by the server. We also didn't define what tasks are as we only had workflows and activities. We either need to define the relationship or do not mention tasks.

cullywakelin:
Closing in favor of: https://github.com/temporalio/documentation/pull/128

PR #108, opened 2020-09-17
Java SDK "Hello World!" tutorial

cullywakelin:
You can view the tutorial at this URL: https://deploy-preview-108--mystifying-fermi-1bc096.netlify.app/docs/java-helloworld/


docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
+---
+id: java-helloworld
+title: Java SDK "Hello World!" tutorial
+sidebar_label: Java "Hello World!"
+tags: helloworld, java, sdk, intellij, gradle, tutorial
+---
+
+This tutorial is aimed at developers who are new to [Temporal](/docs/overview), beginner Java programmers, or both. It will teach you how to build a simple "Hellow World!" Temporal Workflow application using the [Temporal Java SDK](https://github.com/temporalio/java-sdk) as well as a few basics, such as:
```
cullywakelin, line 8: - "Hellow World!" -> "Hello World!"

docs/java-sdk-tutorial-prerequisites.md
```diff
@@ -0,0 +1,27 @@
+---
+id: java-sdk-tutorial-prerequisites
+title: Java SDK tutorial prerequisites
+sidebar_label: Prerequisites
```
cullywakelin, line 4: - Prerequisites -> Tutorial prerequisites

docs/java-sdk-tutorial-prerequisites.md
```diff
@@ -0,0 +1,27 @@
...
+---
+
+To follow the Java SDK tutorials we recommend that you have the following environments set up.
+
```
cullywakelin, line 8: - Call out the need for a terminal

docs/java-helloworld.md
mfateev, line 8: I would say that it assumes basic knowledge of Java programming. I don't think someone who has no clue about Java is the target audience.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+- How to set up your development environment and its dependencies.
+- How to define an [Activity](/docs/activities) and a [Workflow](/docs/workflows).
+- How to create and execute a Worker program and a Workflow starter program.
+- Where to view Workflow details.
+
+## Prerequisites
+
+Make sure you have the tutorial [prerequisites](/docs/java-sdk-tutorial-prerequisites) setup.
+
+## Project setup
```
mfateev, line 19: Should we consider adding instructions for command line only experience? For example a new gradle project can be easily generated using command line commands.

Another option to consider is to use maven instead of gradle.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+<!--SNIPEND-->
+
+### Workflow
+
+A Workflow is a special function that organizes the sequence of Activity executions. Opposed to an Activity, a Workflow is deterministic in nature. It is where you can specify Activity options, such as what to do if the Activity results in an error or when to time out the Activity execution. Since our Activity is just formatting strings we will only specify the required timeout.
+
+Open HelloWorldWorkflow.java and define the Workflow interface:
+
+<!--SNIPSTART java-hello-world-sample-workflow-interface-->
+<!--SNIPEND-->
```
mfateev, line 74: We need to explain what activity stub is. It is the most tricky part of the code.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+Open HelloWorldStarter.java and paste in the following code:
+
+<!--SNIPSTART java-hello-world-sample-workflow-starter-->
+<!--SNIPEND-->
+
+In the code above, we are creating a Temporal client and using it to host the Workflow with the Temporal server. Said another way, this starter program is sending the Workflow to the Temporal server where it will be broken into Tasks. The Tasks will be sent to the Task Queue that we supplied in the Workflow options.
+
+### Worker
+
+A Worker is the process that will actually execute the Workflow and Activity functions as Tasks.
```
mfateev, line 94: I'm not sure if this is something a new user can grasp without explanation. What does it mean to execute a workflow function as tasks? I know it, but for the new user it can be very confusing IMHO.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Prerequisites
+
+Make sure you have the tutorial [prerequisites](/docs/java-sdk-tutorial-prerequisites) setup.
+
+## Project setup
+
+Open IntelliJ and create a new Gradle project by following Step 1 of the [Getting started with Gradle guide](https://www.jetbrains.com/help/idea/getting-started-with-gradle.html#create_project). Make sure your project is created inside the "temporal-java-tutorials" directory and name your project "helloworld-tutorial". It will take a few moments to complete.
+
+Once Gradle has finished scaffolding you will need to customize the project dependencies. To do this, open the build.gradle file that is in the root of your project. Add the following lines to the dependencies section while replacing {version} with the most recent version listed on the [Temporal SDK Maven repository page](https://search.maven.org/artifact/io.temporal/temporal-sdk):
```
rylandg, line 23: Maybe a dumb question but why can't we just have temporal CLI have a "generate gradle" capability which does all of this stuff for the user?

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+```
+implementation 'io.temporal:temporal-sdk:{version}'
+implementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
+```
+
+The 'logback-classic' dependency will ensure that there is a logger to bind to within the SDK and prevent a default logger warning message.
+
+A "refresh" icon will appear on the screen, click it to load the changes. Gradle will rebuild with the dependencies.
+
+Lastly, to limit the logging output from the SDK, within src/main/resources create a logback.xml file and paste in the following XML:
```
rylandg, line 34: I appreciate you are trying to explain and not just command people, but this might be a bit too verbose. These things aren't practically important outside of doing it once.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+<!--SNIPSTART java-samples-logback-dependency-configuration-->
+<!--SNIPEND-->
+
+## Build "Hello World!" app
+
+Now we are ready to create our Temporal Workflow application.
```
rylandg, line 41: But why? Why are we trying to do this

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+Within src/main/java/ create the "helloworld" directory and within src/main/java/helloworld create the following files:
```
rylandg, line 43: Ok, so having to create this many files for our hello world is scary. I'm 100% sure we can either have it less files, delay creating them all at once but maybe I'm wrong. What I will say is that you should absolutely provide a command to create all of these files if you keep it how it is.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+- HelloWorldActivity.java
+- HelloWorldActivityImpl.java
+- HelloWorldWorkflow.java
+- HelloWorldWorkflowImpl.java
+- HelloWorldStarter.java
+- HelloWorldWorker.java
+
+The following code snippets can be copied and pasted directly into your Java files. However, make sure that the first line of each file specifies the package name:
```
rylandg, line 52: But why not just include that in the snippets?

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+```java
+package helloworld;
+```
+
+### Activity
+
+First, let's define our Activity. Activities are meant to handle non-deterministic code that could result in unexpected results or errors. But for this tutorial all we are doing is taking a string, appending it to "Hello", and returning it back to the Workflow.
+
+Open HelloWorldActivity.java and define the Activity interface:
```
rylandg, line 62: Tbh not 100% sure but might be worth putting the file name/path above each file, ie:


`src/HelloWorldActivity.java`
```
import io.temporal.activity.ActivityInterface;
import io.temporal.activity.ActivityMethod;

@ActivityInterface
public interface HelloWorldActivity {

  @ActivityMethod
  String composeGreeting(String name);
}
```

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+The following code snippets can be copied and pasted directly into your Java files. However, make sure that the first line of each file specifies the package name:
+
+```java
+package helloworld;
+```
+
+### Activity
+
+First, let's define our Activity. Activities are meant to handle non-deterministic code that could result in unexpected results or errors. But for this tutorial all we are doing is taking a string, appending it to "Hello", and returning it back to the Workflow.
```
rylandg, line 60: Two things:

1. It is probably helpful to first briefly explain what we are trying to achieve and what building blocks we will use to reach it. Then the activity will make more sense.
2. I'm not sure we want to talk about the determinism and stuff although I understand it makes it hard to see what the purpose of an activity is. That being said, you would probably break out logic into a separate function whether you are using Temporal or not, so maybe just use Activity as a standard programming abstraction (ie: a function).

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+Open HelloWorldActivity.java and define the Activity interface:
+
+<!--SNIPSTART java-hello-world-sample-activity-interface-->
+<!--SNIPEND-->
+
```
rylandg, line 66: I think you're missing the implementation of the Activity.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+### Workflow
+
+A Workflow is a special function that organizes the sequence of Activity executions. Opposed to an Activity, a Workflow is deterministic in nature. It is where you can specify Activity options, such as what to do if the Activity results in an error or when to time out the Activity execution. Since our Activity is just formatting strings we will only specify the required timeout.
```
rylandg, line 69: I think an analogy might be helpful here (and possibly using one analogy as a unifying narrative). For example a workflow function in Temporal shares a similar purpose to a main method, or a "runner" function - orchestrating other code you've written.

rylandg: If you really want to touch on the determinism it needs to be more prescriptive, ie:

These are the things you can put in a workflow:
*
* 

These are the things you cannot put in a workflow:
*
*

docs/java-helloworld.md
rylandg: I completely agree. I almost think the best route is by making it seem very complex. Why do I need this verbose stub? Well it enables you to make automatically loadbalanced, stateful RPC calls with a native function invocation. A stub is a small price to pay for such functionality.

rylandg: Also I should comment on the samples PR but I'm lazy. I think the comments in the Workflow impl are sort of not helpful. "Set the ActivityOptions" tells me nothing and definitely could be autogenerated comment. I think comments are great but not if they replace reading the code.

rylandg: This isn't perfect but it's bette rthan the other comments

"    // Execute the Activity within the Workflow method
    // If there were more Activities, they would be executed from within this method"

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+<!--SNIPEND-->
+
+Open HelloWorldWorkflowImpl.java and define the Workflow:
+
+<!--SNIPSTART java-hello-world-sample-workflow-->
+<!--SNIPEND-->
+
+### Starter
+
+There are two ways to start a Workflow, either via the CLI or via a starter program. In this tutorial we will start our Workflow by invoking it with a starter program.
```
rylandg, line 83: You probably want to make it clear that starting a workflow is just sending an API request to Temporal and that we provide such capabilities programatically or built into the CLI.

rylandg: Like this was one of the most confusing things to me when I was learning. I was like, "why do I need this other application"

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+Open HelloWorldStarter.java and paste in the following code:
+
+<!--SNIPSTART java-hello-world-sample-workflow-starter-->
+<!--SNIPEND-->
+
+In the code above, we are creating a Temporal client and using it to host the Workflow with the Temporal server. Said another way, this starter program is sending the Workflow to the Temporal server where it will be broken into Tasks. The Tasks will be sent to the Task Queue that we supplied in the Workflow options.
```
rylandg, line 90: " using it to host the Workflow with the Temporal server"

That is actually incorrect. The server itself is not involved and workers simply pull from queues.

rylandg: Actually yea, so there is no "sharding" step for the logic. Each line runs natively which results in a new activity or command task being generated.

rylandg: Also i think it's clear diagrams will be needed. Let me know if you want my help on that.

docs/java-helloworld.md
rylandg: Agree. Its not a fun thing to have to explain but I think we don't need to talk about the technical reality just how to use it. Later we can explain it.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Worker
+
+A Worker is the process that will actually execute the Workflow and Activity functions as Tasks.
+
+Open HelloWorldWorker.java and paste in the following code:
+
+<!--SNIPSTART java-hello-world-sample-worker-->
```
rylandg, line 98: I wish there was an easier way to do this part:

```
    // gRPC stubs wrapper that talks to the local docker instance of temporal service.
    WorkflowServiceStubs service = WorkflowServiceStubs.newInstance();
    // client that can be used to start and signal workflows
    WorkflowClient client = WorkflowClient.newInstance(service);
```

It feels like a one-liner.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+<!--SNIPEND-->
+
+In this function we are creating a Temporal client and using it to create a new Worker. The Worker is registered to handle both HellowWorldWorkflow and HelloWorldActivity Tasks and is configured to listen to the same Task Queue that the Workflow and Activity Tasks are sent to.
```
rylandg, line 101: I think this sentence is so laden with terminology that it will be very hard for new users.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+  - HelloWorldActivityImpl
+  - HelloWorldStarter
+  - HelloWorldWorker
+  - HelloWorldWorkflow
+  - HelloWorldWorkflowImpl
+```
+
+Within IntelliJ, right click on HelloWorldStarter and select Run. This will open a terminal inside of IntelliJ which compiles and executes the starter program.
+
+To verify that the Workflow is running, visit the Web UI in your browser and click on the "Workflows" tab. You will see your Workflow listed there with a "running" status. Click on the RunID to see additional details, such as the payload that is being provided to it.
```
rylandg, line 119: Is this true? Wont this workflow complete almost instantly and not be running when the user checks Temporal web.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+src/main/java/helloworld/
+  - HelloWorldActivity
+  - HelloWorldActivityImpl
+  - HelloWorldStarter
+  - HelloWorldWorker
+  - HelloWorldWorkflow
+  - HelloWorldWorkflowImpl
+```
+
+Within IntelliJ, right click on HelloWorldStarter and select Run. This will open a terminal inside of IntelliJ which compiles and executes the starter program.
```
rylandg, line 117: These feel like they are instructions which need to be followed in order. If that's the case it should probably be numbered steps. Also no screenshots or other feedback makes this section sort of unfriendly.

docs/java-helloworld.md
rylandg: Oh I see, I think you intentionally bring up worker second so the workflow is not pulled from a queue and therefore Temporal web will have it in 'running" state.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
...
+
+To verify that the Workflow is running, visit the Web UI in your browser and click on the "Workflows" tab. You will see your Workflow listed there with a "running" status. Click on the RunID to see additional details, such as the payload that is being provided to it.
+
+To actually execute the Workflow and Activity functions, we need to run the Worker. The Worker, once running, will immediately start polling the Task Queue for Tasks. Since we only have one Workflow with one Activity, then the Worker will pick these up in their respective order and actually execute the functions, returning the output to the Workflow starter process.
+
+Right click on HelloWorldWorker and select Run. Another terminal will open inside of IntelliJ and you will see the Worker log each step it is taking. Since our Workflow is very simple it will finish executing the Workflow and Activity Tasks very quickly. Switch back to the starter terminal to see "Hello World!" printed to the console.
+
+Congratulations you have successfully run a Temporal Workflow!
+
+## Lore check
```
rylandg, line 127: This is great.

rylandg (commented):
* Things could be more fun
* There should be a zip available with all of the finished tutorial state at the start. That way if you just want to try it and not deal 
* I think there could be more purpose behind the tutorial. At this point it's 100% just about getting a running workflow which is obviously very important, but the goal is show the value, it feels more like a chore at this point.
* I think there could be more casual language, its very neutral right now.
* I love how much code there is, but also there is a lot of code for a hello world. Not sure what we can do there.

docs/java-helloworld.md
```diff
@@ -0,0 +1,145 @@
+---
+id: java-helloworld
+title: Java SDK "Hello World!" tutorial
+sidebar_label: Java "Hello World!"
+tags: helloworld, java, sdk, intellij, gradle, tutorial
+---
+
+This tutorial is aimed at developers who are new to [Temporal](/docs/overview), beginner Java programmers, or both. It will teach you how to build a simple "Hellow World!" Temporal Workflow application using the [Temporal Java SDK](https://github.com/temporalio/java-sdk) as well as a few basics, such as:
+
+- How to set up your development environment and its dependencies.
```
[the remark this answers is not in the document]
cullywakelin: good stuff.

docs/java-helloworld.md
cullywakelin: Because, this is the package declaration for the Java samples:
```
package io.temporal.samples.helloworld;
```

docs/java-helloworld.md
cullywakelin: Having this makes sense in the context of splitting up tutorials I think.

docs/java-helloworld.md
cullywakelin: created special template repo that solves this.

docs/java-helloworld.md
cullywakelin: Hmm.. good idea for a snipsync feature...

docs/java-helloworld.md
```diff
@@ -0,0 +1,214 @@
...
+```
+./gradlew test
+```
+
+**IntelliJ**
+
+From within IntelliJ, right click on HelloWorldWorkflowTest and select Run.
+
+Look for "BUILD SUCCESSFUL" in the output to confirm.
+
```
sw-yx, line 143: suggestion (take it or leave it): drop the test part to the bottom as a "bonus" section, or clearly mark this part optional so as to reduce the overall length of this tutorial

docs/java-helloworld.md
```diff
@@ -0,0 +1,214 @@
...
+- `com.google.guava:guava` offers a suit of core and expanded libraries that Gradle uses.
+- `io.temporal:temporal-sdk` enables communication with the Temporal server.
+- `ch.qos.logback:logback-classic` will ensure that there is a logger to bind to within the SDK and prevent a default logger warning message.
+
+A "refresh" icon will appear on the screen, click it to load the changes. Gradle will rebuild with the dependencies.
+
+To limit the logging output from the SDK, within src/main/resources/ create a logback.xml file and paste in the following XML:
+
+<!--SNIPSTART hello-world-project-template-java-logback-dependency-configuration-->
+<!--SNIPEND-->
```
sw-yx, line 55: (just asking, maybe the answer is no) a lot of setup. is there any way to collapse this into a single command?

sw-yx (approved):

rylandg (approved):
I support Shawns suggestions but will approve because I don't see it as blockers.

docs/java-helloworld.md
cullywakelin: Unfortunately no -
But that could be a cool tool to create - Temporal Java project scaffold tool
But at present these are the steps to create a project from scratch - which is the intent of the tutorial.

docs/java-helloworld.md
cullywakelin: Hmm.. ya good idea. But I think that because this is creating an app from scratch - testing is an essential part of that. For now I will keep it as is and see what kind of feedback we get.

PR #111, opened 2020-09-17
Upgrade locked packages to address security vulnerabilities 

cullywakelin:
Bumps versions for both alerts:
https://github.com/temporalio/documentation/network/alert/yarn.lock/node-fetch/open
https://github.com/temporalio/documentation/network/alert/yarn.lock/node-forge/open

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,109 +2,109 @@
 # yarn lockfile v1
 
 
-"@algolia/cache-browser-local-storage@4.4.0":
-  version "4.4.0"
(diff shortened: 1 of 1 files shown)


feedmeapples:
you probably want to either:

1. Roll back and upgrade specific packages `yarn upgrade node-forge node-fetch`
2. Address docusaurus change: "The docs plugin option excludeNextVersionDocs=false is deprecated. Use the includeCurrentVersion=true option instead!"

feedmeapples (approved):

PR #125, opened 2020-10-08
Add documentation for tctl's command line arguments for TLS

sergeybykov:


docs/tctl.md
```diff
@@ -315,3 +318,15 @@ Then use this command to tell Temporal to auto-reset all workflows impacted by t
...
+
+#### Secure connection to Temporal cluster
+
+`tctl` supports optional Transport Level Security (TLS) for secure communication with Temporal, authentication of the server, and authentication of the client (mutual TLS).
+
+`--tls_ca_path=<certificate file path>` command line argument allows to pass a certificate authority (CA) certificate for validating server `tctl` is connecting to.
+
+`--tls_cert_path=<certificate file path>` command line argument allows to pass a certificate for server to validate the client (`tctl`) identity. Requires `--tls_key_path` to also be provided.
+
+`--tls_key_path=<private key file path>` command line argument allows to pass a private key ofr secure communication with the server. Requires `--tls_key_path` to also be provided.
```
cullywakelin, line 330:
```suggestion
`--tls_key_path=<private key file path>` command line argument that passes a private key for secure communication with the server. Requires that `--tls_key_path` is also provided.
```

cullywakelin (approved):
Awesome, thank you for updating the docs!

I made some suggestions to the phrasing which I would like to see happen before this lands, but you should double check to make sure they retain the intended meaning. Approving so you can land after updating.

PR #126, opened 2020-10-08
Fix context propogator and tracing documentation

alexshtin:


cullywakelin (approved):
The whole page needs a bit of a polish in terms of grammatical accuracy and directness. But, I don't think this PR is the place to address that - Since it aims to simply fix technical inaccuracies. So, this looks good to me as is -

PR #136, opened 2020-10-13
Update server config

wxing1292:
* Remove `name` from `membership` configuration since `name` is unused
* Change `membership` configuration description since `membership` is not required
* Change `certFile` and `keyFile` example since there is no private cert
* Remove `disableLogging` and  `logLevel` from RPC configuration since `disableLogging` and  `logLevel` are unused
* Update statsd metrics `flushInterval` defaults


wxing1292:
BTW, there are missing documentation about the following
* metrics `prometheus` and `m3`
* publicClient (which is used by internal worker maintenance workflow)

cullywakelin (approved):
Awesome! Thank you for the updates.

PR #153, opened 2020-10-14
Java start workflow executions doc update

wxing1292:


docs/java-starting-workflow-executions.md
```diff
@@ -14,9 +14,11 @@ a client side stub to the workflow, and then calling a method annotated with @Wo
...
 ```
 
-There are two ways to start workflow execution: asynchronously and synchronously. Asynchronous start initiates a workflow execution and immediately returns to the caller. This is the most common way to start workflows in production code. Synchronous invocation starts a workflow
-and then waits for its completion. If the process that started the workflow crashes or stops waiting, the workflow continues executing.
-Because workflows are potentially long running, and crashes of clients happen, this is not very commonly found in production use.
+There are two ways to start workflow execution: asynchronously and synchronously.
+
+Asynchronous start initiates a workflow execution and immediately returns to the caller. This is the most common way to start workflows in production.
+
+Synchronous invocation starts a workflow and then waits for its completion. If the process (synchronous invocation) that started the workflow crashes or stops waiting, the workflow will continue execution. This is not very commonly found in production use.
```
feedmeapples, line 21:
```suggestion
Synchronous invocation starts a workflow and then waits for its completion. If the invocation process crashes or stops waiting, the workflow will continue execution. This is not very commonly found in production use.
```

feedmeapples (approved):

docs/java-starting-workflow-executions.md
feedmeapples, line 21:
```suggestion
Synchronous invocation starts a workflow and then waits for its completion. The workflow doesn't depend on the initial invocation process and will continue executing even if the process crashed or was stopped.
```

feedmeapples, line 21:
```suggestion
Synchronous invocation starts a workflow and then waits for its completion. The started workflow will not rely on the invocation process and will continue executing even if the process crashed or was stopped.
```

feedmeapples (approved):

docs/java-starting-workflow-executions.md
feedmeapples: just to consider. cc @cullywakelin

PR #169, opened 2020-10-21
Nightmode fix sorta

rylandg:
This isn't testable on the live deploy. I tested it via the following:

1. Temporarily re-enable the theme selection for docusaurus 

```
    colorMode: {
      defaultMode: 'dark',
      disableSwitch: false,
    },
```

2. Open up the site in incognito and enable light mode
3. Open a new incognito tab and close the docusaurus one so the incognito window stays open but docusaurus does not
4. Navigate to docusaurus again. Things will still look broken. 
5. Refresh the page. The issue is permanently fixed.

To be clear standard users will need to load the site and refresh once to fix the issue. After doing so it will be gone permanently.

### DIFF
--- a/src/theme/Layout/index.js
+++ b/src/theme/Layout/index.js
@@ -4,7 +4,7 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import React from 'react';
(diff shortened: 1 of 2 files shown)


cullywakelin (approved):
lgtm 🤷

PR #170, opened 2020-10-22
add contact us links to Other SDKs list

sw-yx:
just based on feedback that we aren't clear enough 
image


cullywakelin (approved):
Ya, good addition.

PR #174, opened 2020-10-26
tweak prerequisites wording to be a little friendlier

sw-yx:
tweak prerequisites wording to be a little friendlier to people who scan docs and who are new to docker (more thoughts on this: https://dev.to/meeshkan/how-to-remove-condescending-language-from-documentation-4a5p)

### DIFF
--- a/docs/go-tutorial-prerequisites.md
+++ b/docs/go-tutorial-prerequisites.md
@@ -12,4 +12,4 @@ Make sure you have [Go](https://golang.org/doc/install) installed. These tutoria
 
 ## Temporal server
 
-Download, install, and run the Temporal server via `docker-compose`. You can find Temporal server setup instructions here: [https://docs.temporal.io/docs/install-temporal-server](https://docs.temporal.io/docs/install-temporal-server). It is easy to do and you can keep it running in the background while you build applications.
+Download, install, and run the Temporal server via `docker-compose`. You can find Temporal server setup instructions here: [https://docs.temporal.io/docs/install-temporal-server](/docs/install-temporal-server). It is easy to do and you can keep it running in the background while you build applications.
--- a/docs/go-tutorial-prerequisites.md
+++ b/docs/go-tutorial-prerequisites.md
@@ -12,4 +12,4 @@ Make sure you have [Go](https://golang.org/doc/install) installed. These tutoria
 
 ## Temporal server
 
-Download, install, and run the [Temporal server](/docs/install-temporal-server) via docker-compose. It is easy to do and you can keep it running in the background while you build applications.
+Download, install, and run the Temporal server via `docker-compose`. You can [find Temporal server setup instructions](/docs/install-temporal-server) here. You should keep it running in the background while you build applications.
--- a/docs/go-tutorial-prerequisites.md
+++ b/docs/go-tutorial-prerequisites.md
@@ -12,4 +12,4 @@ Make sure you have [Go](https://golang.org/doc/install) installed. These tutoria
 
 ## Temporal server
 
-Download, install, and run the [Temporal server via `docker-compose`](/docs/install-temporal-server). You should keep it running in the background while you build applications.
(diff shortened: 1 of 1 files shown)


docs/go-tutorial-prerequisites.md
```diff
@@ -12,4 +12,4 @@ Make sure you have [Go](https://golang.org/doc/install) installed. These tutoria
 
 ## Temporal server
 
-Download, install, and run the [Temporal server](/docs/install-temporal-server) via docker-compose. It is easy to do and you can keep it running in the background while you build applications.
+Download, install, and run the Temporal server via `docker-compose`. You can [find Temporal server setup instructions](/docs/install-temporal-server) here. You should keep it running in the background while you build applications.
```
cullywakelin, line 15: "Friendlier" is definitely subjective here. I don't know think this changes the tone at all.
My rules for using "here" to direct to a link is that you use a colon and then provide the url. I.E. :

> You can find the info here: https://someurl.com

"here" is providing direction. Which means, you shouldn't be linking text prior to the use of "here", and either "here" needs to be linked, or use a colon, and I am against linking "here". 

I think you could update the link to include docker-compose: 

```suggestion
Download, install, and run the [Temporal server via `docker-compose`](/docs/install-temporal-server). You should keep it running in the background while you build applications.
```
Or do this:

```suggestion
Download, install, and run the Temporal server via `docker-compose`. You can find Temporal server setup instructions here: [https://docs.temporal.io/docs/install-temporal-server](https://docs.temporal.io/docs/install-temporal-server) . You should keep it running in the background while you build applications.
```

As for this line: 

> You should keep it running in the background while you build applications.

vs this:

> It is easy to do and you can keep it running in the background while you build applications

I am not sure I agree with that change. If quantifying it as "easy" is problematic then perhaps just drop it to "you can keep it running in the background". But, I think "easy to do" helps folks to not get overwhelmed with the additional set up steps and is in fact.. friendlier :)

So, TBH - I think this has the opposite effect of making this section friendlier.

cullywakelin (changes requested):

docs/go-tutorial-prerequisites.md
sw-yx: ok, i will accept your 2nd suggestion!

docs/go-tutorial-prerequisites.md
```diff
@@ -12,4 +12,4 @@ Make sure you have [Go](https://golang.org/doc/install) installed. These tutoria
 
 ## Temporal server
 
-Download, install, and run the [Temporal server](/docs/install-temporal-server) via docker-compose. It is easy to do and you can keep it running in the background while you build applications.
+Download, install, and run the [Temporal server via `docker-compose`](/docs/install-temporal-server). You should keep it running in the background while you build applications.
```
sw-yx, line 15: Shall we do this?

```suggestion
Download, install, and run the Temporal server via `docker-compose`. You can find Temporal server setup instructions here: [https://docs.temporal.io/docs/install-temporal-server](https://docs.temporal.io/docs/install-temporal-server). It is easy to do and you can keep it running in the background while you build applications.
```

cullywakelin (approved):
Cool. I dig it.

docs/go-tutorial-prerequisites.md
```diff
@@ -12,4 +12,4 @@ Make sure you have [Go](https://golang.org/doc/install) installed. These tutoria
 
 ## Temporal server
 
-Download, install, and run the [Temporal server](/docs/install-temporal-server) via docker-compose. It is easy to do and you can keep it running in the background while you build applications.
+Download, install, and run the Temporal server via `docker-compose`. You can find Temporal server setup instructions here: [https://docs.temporal.io/docs/install-temporal-server](https://docs.temporal.io/docs/install-temporal-server). It is easy to do and you can keep it running in the background while you build applications.
```
cullywakelin, line 15: nit:
```suggestion
Download, install, and run the Temporal server via `docker-compose`. You can find Temporal server setup instructions here: [https://docs.temporal.io/docs/install-temporal-server](/docs/install-temporal-server). It is easy to do and you can keep it running in the background while you build applications.
```
I guess you can do a relative link for the actual link - and show the full URL as the text.

PR #178, opened 2020-10-28
Merging SDK tutorials

flossypurse:
This PR shows what the tutorials will look like if there is a single set. All languages are available in tabs vs having separate tutorials for each language.

It introduces a new category in the menu called 'SDK tutorials' and lists them there.

Most of the content is the same, just reorganized to suit the format.


flossypurse:
Decided to keep tuts separated by language. Moved improvements to each tutorial in this pr: https://github.com/temporalio/documentation/pull/207

PR #190, opened 2020-11-09
Adding links in menu to existing API reference for Go and Java

cullywakelin:


sidebars.js
```diff
@@ -49,6 +49,11 @@ module.exports = {
       collapsed: true,
       label: 'Go SDK',
       items: [
+        {
+          type: 'link',
+          label: 'API Reference',
```
sw-yx, line 54: is there some way we can mark these as external links? would be a nice to have the svg icon in here, or failing which, an emoji.

sw-yx (approved):

sidebars.js
cullywakelin: I will take note of this and see about doing some customizations there -

sw-yx: can use 🔗 emoji if we can't use the external link svg

rapmastersethmc (commented):
Agree - we should mark external links in some way.

rapmastersethmc (approved):

sidebars.js
cullywakelin: Cool. Now looks like this:
Screen Shot 2020-11-10 at 8 49 37 AM

PR #226, opened 2021-01-08
Updating Go Worker registration of a Workflow example

flossypurse:
Per user feedback, fixing outdated example of registering a Workflow with a Worker.


rapmastersethmc (approved):
Go Go Go!

PR #238, opened 2021-01-21
Changed SDK section to not be collapsed by default

flossypurse:
Default sidebar:
Screen Shot 2021-01-21 at 12 32 31 PM


rylandg (approved):
Awesome

PR #243, opened 2021-01-28
Revising Elasticsearch setup instructions per feedback

flossypurse:


sw-yx (approved):

docs/server-elasticsearch-setup.md
```diff
@@ -51,7 +48,12 @@ persistence:
 
 ## Kafka
 
-**Note that you only need to complete this step if you are using a Server version that is < 1.6.0**
+:::note
```
sw-yx, line 51: i've never seen this :::note syntax!

docs/server-elasticsearch-setup.md
```diff
@@ -7,31 +7,28 @@ sidebar_label: Set up Elasticsearch
...
+If you are not using Kafka, all you need to do is edit the `persistence` section.
+If you are using Kafka, you will need to edit the `kafka` section as well.
 
 You can use the Temporal Server [development_es.yaml](https://github.com/temporalio/temporal/blob/master/config/development_es.yaml) file as working example.
 
 ## Persistence
 
 Elasticsearch relies on storage to operate.
 Storage related configurations are defined within the `persistence` section of the configuration file.
-The Temporal Server needs to know where to connect to the Elasticsearch storage.
```
alexshtin, line 24: The last sentence is ok. I don't like the first one. We are not describing how ES works here, right? I would replace these 3 sentences with something like: "Elasticsearch is a search engine. The Temporal Server needs to know where to connect to the Elasticsearch. Add the `advancedVisibilityStore: es-visibility` key value pair to the `persistence` section."

docs/server-elasticsearch-setup.md
```diff
@@ -7,31 +7,28 @@ sidebar_label: Set up Elasticsearch
...
 
-To do this, first add the `advancedVisibilityStore: es-visibility` key value pair to the `persistence` section.
+To configure Elasticsearch add the `advancedVisibilityStore: es-visibility` key value pair to the `persistence` section.
```
alexshtin, line 23: We are not configuring ES, we are configuring Temporal to work with ES.

docs/server-elasticsearch-setup.md
```diff
@@ -7,31 +7,28 @@ sidebar_label: Set up Elasticsearch
...
-Temporal Server versions < 1.6.0 require the use of Kafka + Zookeeper to use Elasticsearch.
-Temporal Server versions >= 1.6.0 do not require any additional dependencies to use Elasticsearch.
-The [Server versions and dependencies](/docs/server-versions-and-dependencies) page lists compatible versions of dependencies.
+The [Server versions and dependencies](/docs/server-versions-and-dependencies/#workflow-search) page describes which versions require the use of Kafka, and which ones do not.
 
 Integration with Elasticsearch is defined in the Server configuration files.
-For Server versions >= 1.6.0, all you need to do is edit the `persistence` section.
-For Server versions < 1.6.0, you will need to edit the `kafka` section as well.
+If you are not using Kafka, all you need to do is edit the `persistence` section.
+If you are using Kafka, you will need to edit the `kafka` section as well.
```
alexshtin, line 14: Unfortunately, this is not true. `kafka` section can be omitted only in `v1.7.0`. It is required to be there for all current releases even if it is not used. There are two options here:
1. Release another patch for 1.5.x and 1.6.x and leave doc as is (less confusing but requires some work).
2. Fix doc and say that `kafka` section is required for versions `< 1.7.0`.

alexshtin (changes requested):

docs/server-elasticsearch-setup.md
flossypurse: No, but we are trying to make the connection between configuring persistence (storage) and Elasticsearch.
I.e why am editing a section called "persistence" when I am trying to set up a search engine?

docs/server-elasticsearch-setup.md
flossypurse: We are not saying that it can be omitted from the config. We are saying that you don't need to edit that section.

If Kafka is disabled, per the instructions here: https://github.com/temporalio/temporal/releases/tag/v1.5.0 - then why would you need to edit this section? That doesn't make any sense to me - either the configuration is used or not?

PR #246, opened 2021-02-07
swap slack invite link

sw-yx:
we were getting things like this: 
image


sw-yx:
also fixed this broken link 
image

flossypurse (approved):

flossypurse:
How long is the invite link good for? Does it expire?

sw-yx:
confirmed that we'll keep refreshing the same link

PR #249, opened 2021-02-10
Rename clone destination of temporal server to temporal-server

daisy1754:
docker-compose is very common name in this context. I suggest to change it to be specific


flossypurse (approved):
Hey - this is a great suggestion. Thank you!

PR #250, opened 2021-02-10
Fix broken link to localhost (add http://)

daisy1754:
existing link didn't work for me with this error - https://stackoverflow.com/questions/64797802/what-is-this-scheme-dont-have-a-registered-handler-error


flossypurse (approved):
Thank you for the fix!

PR #274, opened 2021-02-24
Adding Go selectors page

flossypurse:


docs/go-selectors.md
```diff
@@ -0,0 +1,120 @@
...
+title: Go Selectors
+sidebar_label: Selectors
+---
+
+## Overview
+
+In Go, the `select` statement lets a goroutine wait on multiple communication operations.
+A `select` **blocks until one of its cases can run**, then it executes that case.
+It chooses one at random if multiple are ready.
+Temporal's Go SDK `Selector`s are similar, but can block on sending and receiving from `Channel`s as well as `Future` deferred work.
```
mfateev, line 12: I would rewrite this paragraph to emphasize that Go select cannot be used in Temporal workflows due to determinism. That's why Selector is provided as a replacement. As a bonus, it has additional features like the ability to listen on a Future.

docs/go-selectors.md
```diff
@@ -0,0 +1,120 @@
...
+	})
+```
+
+Merely matching on the channel doesn't consume the message; it has to be explicitly consumed with a `c.Receive(ctx, &signalVal)` call.
+
+## Querying Selector State
+
+TODO: document `selector.HasPending`
+Context: This was added to ensure that signals are not lost when Workflow closed.
+Explain how to ensure signals are not lost.
```
sw-yx, line 103: any thoughts on whether/how to document this?

docs/go-selectors.md
```diff
@@ -0,0 +1,120 @@
+---
+id: go-selectors
+title: Go Selectors
```
sw-yx, line 3: possible alternative: "Go SDK Selectors"? to make it clear its our thing, not a Go thing

docs/go-selectors.md
flossypurse: Maybe, but right now its making this pr stale -

flossypurse: Made a task - lets tackle this in another iteration -
What else needs to happen to get this into the docs?

sw-yx: i dont see anything else blocking IMO

sw-yx (approved):

PR #275, opened 2021-02-25
Refreshing and consolidating Java Activity information

flossypurse:
This PR renames java-implementing-activities -> java-activities
Information from the old java-implementing-activities & java-activity-interface pages is updated and consolidated into java-activities

Preview is here: https://deploy-preview-275--mystifying-fermi-1bc096.netlify.app/docs/java-activities


docs/java-activities.md
```diff
@@ -0,0 +1,207 @@
+---
+id: java-activities
+title: Activities in Java
+sidebar_label: Activities
+---
+
+## Overview
+
+In Java, a Temporal Activity is simply an object method that contains any business-level logic.
```
Sushisource, line 9: This is maybe not as informative as it could be? Workflows can and should probably have the majority of your business logic, while activities exist to do things workflows can't - like block, perform heavy computation, side effects, be non-deterministic, etc. Probably worth alluding to that distinction here as well.

docs/java-activities.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Activity Interface
+
+Activities are defined as methods of a plain Java interface annotated with `@ActivityInterface`.
+Each method defines a single Activity type.
+A single Workflow can use more than one Activity interface and call more than one Activity method from the same interface.
+The only requirement is that Activity method arguments and return values are serializable to a byte array using the provided `io.temporal.common.converter.DataConverter` implementation.
+The default implementation uses a JSON serializer, but an alternative implementation can be configured through `io.temporal.client.WorkflowClientOptions.Builder.setDataConverter()`.
+
+Example of an interface that defines four Activities:
```
Sushisource, line 19: Maybe worth saying "four activities for interacting with S3"

Sushisource (approved):
LGTM just some minor stuff

docs/java-activities.md
flossypurse: I am actually going to scrap the overview altogether.
The idea of "what" an Activity is should be in the concepts section. Otherwise, we are repeating ourselves over and over.

PR #277, opened 2021-02-26
add external resources

sw-yx:
blogposts and tutorials and tweets bc why not

ok result here: https://603953a9caae29000865475f--mystifying-fermi-1bc096.netlify.app/docs/external-resources


flossypurse (approved):
Gorgeous.

PR #294, opened 2021-03-07
Adding redirect rules 

flossypurse:


sw-yx (approved):

netlify.toml
```diff
@@ -4,3 +4,7 @@
     [plugins.inputs]
 
     skipPatterns = ["#!", "www.google-analytics.com", "www.googletagmanager.com"]
+
+[[redirects]]
+  from = "/docs/overview"
+  to = "/docs/concept-overview"
```
sw-yx, line 10: i think not necessary if you have `_redirects` in the right place

flossypurse: The _redirects file needs to be in the build folder - 
With Docusaurus the build folder is created when the site is built.
So, alternative is to use the .toml file - which we already have and doesn't need to be any place other than the root directory.

_redirects
```diff
@@ -0,0 +1,3 @@
+/docs/overview   /docs/concept-overview
+/docs/activities /docs/concept-activities
+/docs/workflows  /docs/concept-workflows
```
sw-yx, line 3: i think this file needs to be in `static`, aka `/static/_redirects`

netlify.toml
sw-yx: we can also put it in `/static` :)

flossypurse: probably, but the .toml file is already there and provides more comprehensive redirect options -
And it works -

PR #302, opened 2021-03-16
Namespaces are a feature of the Server

flossypurse:
The Namespace concept page is legacy content that was aimed at explaining xdc replication.
However, in the new era - we want to represent Namespaces as a feature of the Server.
They are not an abstract programming framework concept like Workflows, Activities, etc...


sw-yx (approved):
great move

PR #319, opened 2021-03-29
Add cadence-to-temporal migration page

flossypurse:
Preview: https://deploy-preview-319--mystifying-fermi-1bc096.netlify.app/docs/cadence-to-temporal

## What was changed:

A new page was added and the sidebar was updated.

## Why?

Information about the key differences between Cadence and Temporal needed to be highlighted and discoverable.

## Closes issue: 

None.

## How has this been tested?

Built locally.

## Any docs updates needed?

This is the docs.

### DIFF
--- a/docs/cadence-to-temporal.md
+++ b/docs/cadence-to-temporal.md
@@ -1,0 +1,66 @@
+---
+id: cadence-to-temporal
+title: Cadence to Temporal migration highlights
+sidebar_label: Migrate from Cadence
+---
+
+This page highlights the key changes between Cadence and Temporal that you will need to account for when migrating.
+
+There is no direct upgrade path from Cadence to Temporal.
+The Temporal Server requires a fresh install, and we recommend choosing one of the [latest releases](https://github.com/temporalio/temporal/releases).
+
+The [changelog blog post](/blog/temporal-v0.28.0-changelog) includes many of the core changes that occurred prior to Temporal's v1.0.0 release.
+
+## Key differences
+
+1. **Cadence uses Thrift/TChannel | Temporal uses [Protobuf](https://developers.google.com/protocol-buffers)/[gRPC](https://grpc.io/)**
+
+When using Temporal, this means the following things:
+
+  - Communication over can be encrypted via [TLS](/docs/server-configuration/#tls).
+  - Custom service errors backed by standard [gRPC error codes](https://pkg.go.dev/google.golang.org/grpc/codes).
+  - Errors are serialized using the Protobuf type `Failure` and they be chained together and passed across different SDKs in different languages.  
+  - All payloads (Workflow input, Activity input, etc...) sent to Temporal now have headers and data fields.
+  - Standard [gRPC health service](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) is used for health checks.
+  - Standard Protobuf types are used for [time](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp) and [duration](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration).
+
+2. **Entities names are different**:
+
+| Cadence | Temporal |
+|---------|----------|
+| Decision | Command |
+| DecisionTask | WorkflowTask |
+| Decider | Workflow |
+| Domain | Namespace |
+| NamespaceStatus | NamespaceState |
+| TaskList | TaskQueue |
+| ArchivalStatus | ArchivalState |
+
+3. **Workflow timeout are different in Temporal**
+
+  - Workflow timeouts are optional.
+  - Attempt counters start at 1.
+  - Workflow Executions have 3 timeouts available:
+    - `WorkflowRunTimeout`.
+    - `WorkflowExecutionTimeout`.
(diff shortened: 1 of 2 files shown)


docs/cadence-to-temporal.md
```diff
@@ -0,0 +1,66 @@
...
+4. **Workflow retries are different in Temporal**
+
+  - In Temporal, `WorkflowExecutionTimeout` is used, instead of `ExpirationInterval`, to stop retries for Workflows.
+  - Attempt counters start at 1.
+
+5. **Activity retries are different in Temporal**
+
+  - Temporal includes default `RetryOptions` for Activities.
+    - `initialInterval` = 1 second.
+    - `maximumInterval` = 100 \* initialInterval.
```
alexshtin, line 57: `\*` ?

docs/cadence-to-temporal.md
```diff
@@ -0,0 +1,66 @@
...
+  - Workflow timeouts are optional.
+  - Attempt counters start at 1.
+  - Workflow Executions have 3 timeouts available:
+    - `WorkflowRunTimeout`.
+    - `WorkflowExecutionTimeout`.
+    - `WorkflowTaskTimeout`.
+
+4. **Workflow retries are different in Temporal**
+
+  - In Temporal, `WorkflowExecutionTimeout` is used, instead of `ExpirationInterval`, to stop retries for Workflows.
```
alexshtin, line 50: I'd mention somehow that `ExpirationInterval` is completely gone/removed.

docs/cadence-to-temporal.md
```diff
@@ -0,0 +1,66 @@
...
+    - `maximumInterval` = 100 \* initialInterval.
+    - `backoffCoefficient` = 2.
+  - To disable Activity retries, configure Activities with `RetryOptions` that set `maximumAttempts` to 1.
+  - `ScheduleToClose` is used to stop retries for Activities.
+
+6. **[Temporal dependencies](/docs/server-versions-and-dependencies) are different**
+
+  - Uber YARPC library is not used by Temporal.
+  - Temporal no longer needs to use Kafka.
+  - Temporal does not use statsd.
```
alexshtin, line 66: "statsd" is not used by default but it is still supported. "Prometheus" is the new default.

alexshtin (commented):
Few more things:
1. Config changes. There were few and some of them might (I am not sure) prevent driect reuse of Cadence config.
2. Is this doc about server only? What's about SDK changes?

docs/cadence-to-temporal.md
```diff
@@ -0,0 +1,66 @@
...
+
+When using Temporal, this means the following things:
+
+  - Communication over can be encrypted via [TLS](/docs/server-configuration/#tls).
+  - Custom service errors backed by standard [gRPC error codes](https://pkg.go.dev/google.golang.org/grpc/codes).
+  - Errors are serialized using the Protobuf type `Failure` and they be chained together and passed across different SDKs in different languages.  
+  - All payloads (Workflow input, Activity input, etc...) sent to Temporal now have headers and data fields.
+  - Standard [gRPC health service](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) is used for health checks.
+  - Standard Protobuf types are used for [time](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp) and [duration](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration).
+
```
alexshtin, line 26: Another thing worth to mention here, I believe that persistence layer is also using protobufs blobs for most of the data for both MySQL and Cassandra. In Cadence Thrift was used in MySQL only.

sidebars.js
```diff
@@ -152,6 +152,7 @@ module.exports = {
       ],
     },
     'external-resources',
+    'cadence-to-temporal',
```
sw-yx, line 155: personally think we should be very sparing with top level menu items. keep in mind that most people will not be migrating so this will be irrelevant to them.  what do you think of sticking it under Server?

flossypurse: Info in here is not Server specific.
Part of the objective is to increase discoverability, and this info already exists as part of the changelog. Burying it only defeats the purpose.
Its existence also has a time-limit. So its there for now while we need it - then it goes away.

docs/cadence-to-temporal.md
flossypurse: This is so that the build system doesn't think it is some sort of operator and print it is as a simple `*`. Otherwise build system errors.

sidebars.js
flossypurse: Maybe we could have an "Additional resources" ?

sw-yx: ah if it is time limited then thats not such a big deal :)

i dislike "additional resource" because too vague. in absence of a better idea, lets keep this as-is!

sw-yx (approved):

alexshtin (approved):

flossypurse:
> Few more things:
> 
> 1. Config changes. There were few and some of them might (I am not sure) prevent direct reuse of Cadence config.

OK - lets add a quick blurb about it.

> 2. Is this doc about server only? What's about SDK changes?

It is meant to highlight all major differences (not just Server).

PR #357, opened 2021-04-19
attempt algolia search fix

sw-yx:
> This doesnt work yet! we THINK we have a cause, but we are not sure it is the ROOT cause. have filed a pair issue on the docusaurus repo https://github.com/facebook/docusaurus/issues/4644

## What was changed:

changed algolia search config

## Why?
our search page [always returns blank](https://docs.temporal.io/search?q=workflow) right now. we think its because as of this PR
https://github.com/temporalio/documentation/pull/327 (When we upgraded Docusaurus from alpha 66 to  alpha 72)

- Working: https://deploy-preview-326--mystifying-fermi-1bc096.netlify.app/search/?q=workflows
- Broken: https://deploy-preview-327--mystifying-fermi-1bc096.netlify.app/search/?q=workflows

the search has started to send in default facets `["language","docusaurus_tag"]` and facetFilters `[["language:en"],["docusaurus_tag:default","docusaurus_tag:docs-default-current"]]"}` where it didn't use to before (internal notes: https://www.notion.so/temporalio/debug-search-8213484654214fb784952d0e5d474501)

as far as I can tell this is undocumented and i'm not sure it is intentional behavior (cc @slorber in case this was supposed to be documented)

## How was this tested:

testing in deploy preview:

https://deploy-preview-357--mystifying-fermi-1bc096.netlify.app/search/?q=workflow

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -187,7 +187,10 @@ module.exports = {
       indexName: 'temporal',
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
-      // algoliaOptions: {}, // Optional, if provided by Algolia
+      searchOptions: { // used to be called algoliaOptions
+        facetFilters: null,
+        facets: null
+      }
     },
   },
   presets: [
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -188,8 +188,8 @@ module.exports = {
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
       searchParameters: { // used to be called algoliaOptions https://docusaurus.io/docs/search#using-algolia-docsearch
-        facetFilters: null,
-        facets: null
+        facetFilters: [],
+        facets: []
       }
     },
   },
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -187,7 +187,7 @@ module.exports = {
       indexName: 'temporal',
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
-      searchOptions: { // used to be called algoliaOptions
+      searchParameters: { // used to be called algoliaOptions https://docusaurus.io/docs/search#using-algolia-docsearch
         facetFilters: null,
         facets: null
       }
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -185,7 +185,7 @@ module.exports = {
     algolia: {
       apiKey: '14805ba2eb682edb2e719df4d5e03c8a',
       indexName: 'temporal',
-      // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
+      contextualSearch: true, // see if this matters
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -187,7 +187,10 @@ module.exports = {
       indexName: 'temporal',
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
-      // algoliaOptions: {}, // Optional, if provided by Algolia
+      searchOptions: { // used to be called algoliaOptions
```
sw-yx, line 190:
```suggestion
      searchParameters: { // used to be called algoliaOptions https://docusaurus.io/docs/search#using-algolia-docsearch
```

docusaurus.config.js
```diff
@@ -187,7 +187,10 @@ module.exports = {
       indexName: 'temporal',
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
-      // algoliaOptions: {}, // Optional, if provided by Algolia
+      searchParameters: { // used to be called algoliaOptions https://docusaurus.io/docs/search#using-algolia-docsearch
+        facetFilters: null,
+        facets: null
+      }
```
sw-yx, line 193:
```suggestion
      searchParameters: { // used to be called algoliaOptions https://docusaurus.io/docs/search#using-algolia-docsearch
        facetFilters: [],
        facets: []
      }
```

docusaurus.config.js
```diff
@@ -187,7 +187,10 @@ module.exports = {
       indexName: 'temporal',
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
       // appId: 'app-id', // Optional, if you run the DocSearch crawler on your own
-      // algoliaOptions: {}, // Optional, if provided by Algolia
+      searchParameters: { // used to be called algoliaOptions https://docusaurus.io/docs/search#using-algolia-docsearch
+        facetFilters: [],
+        facets: []
```
sw-yx, line 192: this does not work to clear the facetFilters field. i also tried `facetFilters: null` and that did not work.

docusaurus.config.js
```diff
@@ -187,7 +187,10 @@ module.exports = {
       indexName: 'temporal',
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
```
sw-yx, line 188:
```suggestion
      contextualSearch: true, // see if this matters
```

slorber:
hi,

It looks like your DocSearch config has diverged quite a lot from the regular Docusaurus 2 docsearch config:
- https://github.com/algolia/docsearch-configs/blob/master/configs/temporal.json
- https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json

In particular, your config does not permit Algolia to index the facetting fields such as `docusaurus_tag ` that we set in pages and use for contextual search filtering.

When we released Contextual Search, I think Algolia run a batch update of all the Docusaurus 2 configs to add those new fields, but maybe it didn't update your config for some reason?

cc @shortcuts @shipow

docusaurus.config.js
slorber: somehow the legacy search page does not respect these settings, it's like contextual search is always enabled by default in this page so you can't remove the hardcoded facet filters.

But this highlighted a more important problem: your docsearch config is wrong. Fix it, and contextual search + facet filters will both work.

shortcuts:
It indeed seems like this one hasn't been updated @slorber.

I've opened a PR here https://github.com/algolia/docsearch-configs/pull/3986, which will allow you to use the Contextual Search @sw-yx

PR #367, opened 2021-04-22
elaborate on worker service vs Worker vs Task Queue, fix typo in code sample

vkarpov15:
## What was changed:

Changed the intro to the workers section to be a bit more clear. Also fixed a small typo in the code sample.

## Why?

Worthwhile to explain that Workers are objects that poll a Task Queue, and execute Workflows in response to tasks.

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?


docs/go-workers.md
```diff
@@ -4,9 +4,14 @@ title: Workers in Go
...
 ---
 
-A worker or _worker service_ is a service that hosts the Workflow and Activity implementations. The worker polls the _Temporal service_ for tasks, performs those tasks, and communicates task execution results back to the _Temporal service_. Worker services are developed, deployed, and operated by Temporal customers.
+A _worker service_ is a service that hosts Workflow and Activity implementations.
+A worker service starts one or more Workers.
+A Worker polls a Task Queue for tasks, and executes Workflows in response to tasks.
+When you call `ExecuteWorkflow()`, Temporal adds a new task to the Task Queue, and a Worker executes the task.
 
-You can run a Temporal worker in a new or an existing service. Use the framework APIs to start the Temporal worker and link in all Activity and Workflow implementations that you require the service to execute.
+Worker services are developed, deployed, and operated by Temporal customers.
```
tsurdilo, line 12: i would use "Temporal users" rather than "customers" wdyt?

vkarpov15: Works for me, "customers" implies payment so I'm ok to make this change.

docs/go-workers.md
```diff
@@ -4,9 +4,14 @@ title: Workers in Go
...
+You can run a Worker in a new or an existing service.
+Use the framework APIs to start the Worker and link in all Activity and Workflow implementations that you require the service to execute.
```
tsurdilo, line 14: Temporal SDK rather than framework APIs right?
Also I would use "register" rather than "link".

docs/go-workers.md
tsurdilo, line 12: I would use "defined" rather than "developed". To me "developed" implies that you have to write the actual impl of the worker service rather than just use the sdk apis. wdyt?

vkarpov15: :+1: Maybe "As a Temporal user, you're responsible for developing and deploying worker services." ?

tsurdilo: I would maybe have:
"Just like workflows and activities, the workers and worker service are hosted on user applications, and not the Temporal server"
wdyt?

vkarpov15:
@tsurdilo I addressed your comments :+1:

tsurdilo (approved):

PR #368, opened 2021-04-22
add QueryWorkflow() example

vkarpov15:
## What was changed:

Added an example of how to call `QueryWorkflow()` in query docs

## Why?

The current docs show how to register a query handler, but not how to execute the query.

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?


docs/go-queries.md
```diff
@@ -55,7 +55,17 @@ You can now query `current_state` by using the CLI:
 
 `tctl --namespace samples-namespace workflow query -w my_workflow_id -r my_run_id -qt current_state`
 
-You can also issue a query from code using the `QueryWorkflow()` API on a Temporal client object.
+You can also issue a query from code using the `QueryWorkflow()` API on a Temporal client object as shown below.
+
+```golang
+response, err := temporal.QueryWorkflow(context.Background(), workflowID, runID, "current_state")
```
tsurdilo, line 61: if "current_state" refers to the workflow state, maybe we should list all the possible options that uses can enter here? wdyt?

docs/go-queries.md
```diff
@@ -55,7 +55,17 @@ You can now query `current_state` by using the CLI:
 
 `tctl --namespace samples-namespace workflow query -w my_workflow_id -r my_run_id -qt current_state`
 
-You can also issue a query from code using the `QueryWorkflow()` API on a Temporal client object.
+You can also issue a query from code using the `QueryWorkflow()` API on a Temporal client object as shown below.
+
+```golang
```
flossypurse, line 60:
```suggestion
```go
```

Change to `go` as syntax highlighting won't work for `golang`

flossypurse (approved):

flossypurse:
I am approving so that once the syntax highlighting is fixed and Tiho's comments are addressed, you can land.

docs/go-queries.md
vkarpov15: @tsurdilo @flossypurse I added some more info on calling `QueryWorkflow()` with extra positional args, and a section on `QueryWorkflowWithOptions()`.

Also, it doesn't look like the "Consistent Query" section's example is valid. There's no `QueryConsistencyLevel` option on `QueryWorkflowWithOptionsRequest`, here's the [relevant code in the Go SDK](https://github.com/temporalio/sdk-go/blob/6580cbe0aa41a8b515791f95c2c15bb37db1dab1/internal/internal_workflow_client.go#L888). Go prints a "unknown field 'QueryConsistencyLevel' in struct literal of type internal.QueryWorkflowWithOptionsRequest" compiler error when you use that option.

flossypurse (approved):

PR #371, opened 2021-04-22
Proposing blue color scheme

flossypurse:
Primary color is `#127ae5`
Per our Color Palette.
Shades and tints generated here: https://coolors.co/127ae5

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -79,7 +79,7 @@ body {
 }
 
 .footer {
-  background-color: var(--ifm-color-primary);
+  background-color: var(--ifm-color-primary-darkest);
 }
 
 .footer a:hover{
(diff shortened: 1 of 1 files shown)


sw-yx:
imo this blue is a bit too bright still

src/css/custom.css
```diff
@@ -79,7 +79,7 @@ body {
 }
 
 .footer {
-  background-color: var(--ifm-color-primary);
+  background-color: var(--ifm-color-primary-darkest);
```
sw-yx, line 82: made it darker for more contrast - wdyt?

sw-yx (approved):

rylandg (approved):
7.5/10 but strong improvement over current

src/css/custom.css
flossypurse: Nice - I dig it!

PR #372, opened 2021-04-22
Pageids

flossypurse:
## What was changed:

- Files have been moved to dedicated directories i.e `/docs/concepts/`, `/docs/server/`, `/docs/go/`, etc...
- Page ids have been updated, i.e. `concept-workflows` -> `workflows`
- Paths are now using forward slashes like so: `/docs/concepts/<concept>`, i.e. `/docs/concepts/workflows` or `/docs/go/introduction`
- Associated relative site links have been changed
- Redirects have been added to catch old links from external sources.

## Why?

1. The docs folder has a lot of files, this adds a level of organization
Screen Shot 2021-04-22 at 4 30 21 PM
2. It allows us to have better url paths, i.e. https://docs.temporal.io/docs/concepts/workflows vs https://docs.temporal.io/docs/concept-workflows
Screen Shot 2021-04-22 at 4 31 13 PM
3. It allows us to take advantage of the "activeBasePath" feature such that the main menu highlights Server when you are on a Server page.
Screen Shot 2021-04-22 at 4 32 29 PM
## Checklist

How was this tested:

Built locally using `netlify build --context deploy-preview`


feedmeapples (approved):

sw-yx:
woo this is a big change.

can we add client/server redirects for the root paths of the new folders we made?
image

flossypurse:
> woo this is a big change. imo, doing this closes the door for a ong time on doing language-agnostic docs with toggles. so we oughta be really really sure this is the URL structure we want. i dont necessarily have a strong opinion on what is best.. just want to raise it to you heh

We can definitely sit on this over the weekend.
I think we are going to have a really hard time trying to maintain and curate purely language agnostic docs - 
I think it's possible to have a sort of middle ground, but then we would always end up with language specific sections like this. Snipsync will allow us to reuse content within the docs site now, so that is an option for creating reusable content bits. I totally hear what you on this - but I would say, just look at the divergence between Node, Java, PHP, and Go right now - all of them in different states with slightly different features. Feels like this is being pushed on us vs deciding, and not choosing is just going to put us in a weirder spot as we move forward.


> sidenote: can we add client/server redirects for the root paths of the new folders we made?
> image

Ya, actually, that pretty much solves the "default page" feature I was trying to get Joseph to add to Docusaurus. Will add these shortly.

flossypurse:
> woo this is a big change. imo, doing this closes the door for a ong time on doing language-agnostic docs with toggles. so we oughta be really really sure this is the URL structure we want.

The more I think about this - the more I don't see how this actually closes any doors, nor is it really a big change - it is just one of a series that moves us in the direction we already agreed on going.
The URL structure we have now is something we inherited. We use redirects to help change to something else. We can use the same technique to move to something else if we need.

I am going to merge this.

tsurdilo (approved):

tsurdilo:
+1 for this change, the docs dir was huge and depending on file names to hold the section info was a pain

tsurdilo:
@flossypurse the only issue i have with this is for example:

docs/go/go-create-workflows.md

why still call this file "go-create-workflow.md" when its already in the "go" directory.
just  call it create-workflow.md for consistency

rylandg:
I think @tsurdilo suggestion is small but nice.

PR #381, opened 2021-04-26
add labels to section headers in each sidebar

sw-yx:
this is an attempt to solve this problem

image

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -1,9 +1,9 @@
 module.exports = {
   serverSidebar: [
     {
-      type: 'link',
+      type: 'doc',
       label: 'Temporal Server', 
-      href: '/server/introduction',
+      id: '/server/introduction',
     },
     'server/quick-install',
     'server/production-deployment',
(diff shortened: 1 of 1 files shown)


sw-yx:
the search doesnt change until this gets merged and... as far as i can tell... this change has no outward visible changes. imo we should give this a try and see how the search reranks

sidebars.js
```diff
@@ -1,6 +1,10 @@
 module.exports = {
   serverSidebar: [
-    'server/introduction',
+    {
+      type: 'doc',
+      label: 'Temporal Server', 
+      id: 'server/introduction',
+    },
```
flossypurse, line 7: I don't think this will have any effect.
These are defined in the file already.

flossypurse:
I think this will have the effect we are looking for:
https://github.com/temporalio/documentation/pull/383

sw-yx:
yeah i think so. lets try your thing first!

PR #383, opened 2021-04-26
Categorizing search results

flossypurse:


sw-yx (commented):
i was trying to avoid the visual shift but if you're fine with that i think this will work :)

sw-yx (approved):

flossypurse:
> i was trying to avoid the visual shift but if you're fine with that i think this will work :)

Cool. Ya, I think I am good with this - I like it actually.
And we have the option to completely disable the category collapsing on the site if we think that will be a problem.

PR #388, opened 2021-04-28
delete orphan PHP page

sw-yx:
## Why?

this is showing up in search https://docs.temporal.io/docs/php/activities/


flossypurse:
Instead of deleting it - let's migrate content to it.

sw-yx:
handled in #415

PR #393, opened 2021-04-28
Refreshing info for Go Workflows & Workers

flossypurse:
## What was changed:

Consolidated Workflow info. creating-workflows.go & sync-and-async-start .go-> workflows.go
Refreshing Go Worker info.
Updated links. Added redirects.

Previews:
- https://deploy-preview-393--mystifying-fermi-1bc096.netlify.app/docs/go/workflows
- https://deploy-preview-393--mystifying-fermi-1bc096.netlify.app/docs/go/workers

## Why?

Focusing on core concept documentation and creating an easy to follow story for users.

## Checklist

1. Closes issue: 

N/A

2. How was this tested:

Build locally using `netlify build --context deploy-preview`

3. Any docs updates needed?

N/A


docs/go/workflows.md
```diff
@@ -42,4 +43,114 @@ A Workflow can have one or more such parameters.
...
+we = client.GetWorkflow(workflowID)
+var result string
+we.Get(ctx, &result)
+```
+
+## How to get data in or out of a running Workflow
+
+[Signals](/docs/go/signals) are the mechanism by which you can get data into already running Workflow.
+
+[Queries](/docs/go/queries) are the mechanism by which you can get data out of currently running Workflow.
```
sw-yx, line 156: i feel like we should draw a diagram to represent this at some point

docs/go/workflows.md
```diff
@@ -42,4 +43,114 @@ A Workflow can have one or more such parameters.
...
+    if err != nil {
+        // Handle failure
+    }
+
+    // Do something with the result
+    printResults(workflowResult, we.GetID(), we.GetRunID())
+}
+```
+
+In Go, the only difference between a synchronous start and an asynchronous start is whether you use the `WorkflowExecution` that is returned from `ExecuteWorkflow` to get the result of the Workflow from the same process in which you started it.
```
sw-yx, line 141: i find this sentence hard to read. too long. also i feel like we should have a stronger recommendation for async start. care to take another try at this?

sw-yx (approved):
just one comment about the sync/async thing you may wish to address

docs/go/workflows.md
flossypurse: 👍 Agreed, I will make task to follow up.

docs/go/workflows.md
flossypurse: Good feedback, rewrote this paragraph and broke it down more with an explicit leaning towards async.

docs/go/workflows.md
```diff
@@ -42,4 +43,118 @@ A Workflow can have one or more such parameters.
...
+Workflows do not rely on the process that invoked it, and will continue executing even if the waiting process crashes or stops.
+As long as you have the WorkflowID, which is available from the `WorkflowExecution` you can retrieve the result of the Workflow from a completely different process by using the `GetWorkflow` call on the client.
+
+```go
+we = client.GetWorkflow(workflowID)
+var result string
+we.Get(ctx, &result)
+```
+
+In most uses cases it is better to be prepared to execute the Workflow asynchronously.
```
sw-yx, line 154:
```suggestion
You can start workflows either **asynchronously** (most common) or **synchronously**.
In Go, the only difference is whether the code waits for the result of the Workflow in the same process in which you started it.

When you start a Workflow, a `WorkflowExecution` is returned.
The `WorkflowExecution` can be used to get the result (synchronous) or capture the WorkflowId (asynchronous).
Workflows do not rely on the process that invoked it, and will continue executing even if the waiting process crashes or stops.

As long as you have the WorkflowID, which is available from the `WorkflowExecution`, you can retrieve the result of the Workflow from a completely different process by using the `GetWorkflow` call on the client.

```go
we = client.GetWorkflow(workflowID)
var result string
we.Get(ctx, &result)
```

```

trying to break up the paragraph a little

flossypurse: I hear ya.

Though, I don't think saying something is "most common" is actually a recommendation per se - more of an observation.

I will adopt most of your recommendations here.
However, I do have a strong preference towards laying out the state of things, and then providing a soft recommendation at the end. As the story preceding the recommendation lays the foundation for it.

wdyt about this?

Screen Shot 2021-04-29 at 12 20 52 PM

PR #395, opened 2021-04-29
Java Workflow, Worker, & continue-as-new refresh

tsurdilo:

## What was changed:

Adding new doc page for Continue as new Temporal feature (for Java)

## Why?

It was missing

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?


sw-yx:
i have an idea for you and for @flossypurse  - what if we didn't put this page on its own as the "ContinueAsNew" page. most people wouldn't know what that is.

ideas:

- rename it as the "Long Running Workflows" page or the "Infinitely Long Workflows" page
- inline this page into the SDK Workflows page, so that everyone who reads the SDK Workflows page is guaranteed to learn about ContinueAsNew - this is because it is a concept that almost everybody ends up asking about

basically we should come up with an idea of what justifies a new page, and how we name it. I'd like us to think about the user POV.

tsurdilo:
@sw-yx the go page has little and imo confusing info on it (seems just a copy of the equivalent Cadence doc page on this subject)

We can use the explanation i put in this pr for all 3 languages and just chance the code to match.

tsurdilo:
@sw-yx I would honestly not call it "Long Running Workflows" or "Infinitely Long Workflows". 
There are cases IMO where you can use it even if you have very short lived periodically invoked workflow that may never even run into history size limit if you ran it for a year. MO "Continue As New" kinda describes itself in a way :)

tsurdilo:
Oh and i forget, the fact that you can continue as new as a completely new workflow type, it also has some uses for versioning as well. I did not really want to go into that as we don't have a good example or use case for it yet but yeah there are uses outside of just long running workflows and history size limitations.

sw-yx:
ok then. happy to proceed

tsurdilo:
@sw-yx sorry i was just talking about the first point 

Totally agree with 

* inline this page into the SDK Workflows page, so that everyone who reads the SDK Workflows page is guaranteed to learn about ContinueAsNew - this is because it is a concept that almost everybody ends up asking about

I just need some help to figure out where to put this :)

sw-yx:
lets have a focused chat on this on monday with cully gets back - can you book a short meeting with the 3 of us

flossypurse:
Lots of awesome angles here - lets definitely set some time aside early in the week to discuss.

sw-yx:
as discussed, we are going to:

- move ContinueAsNew content into the new `/shared` folder, and re-import it from Go/Java etc
- for Java docs, we consolidate all the Workflows content into the new `/java/workflows` page
- for Go docs, we move `/go/continueasnew` -> `/go/workflows` and use the new shared continueasnew content

sidebars.js
```diff
@@ -179,9 +179,7 @@ module.exports = {
         },
         'java/activities',
         'java/quick-start',
-        'java/workflow-interface',
-        'java/implementing-workflows',
-        'java/starting-workflow-executions',
+        'java/workflows',
```
sw-yx, line 182: these are the java docs we collapsed into a single doc - need to doublecheck redirects before we merge this

tsurdilo: thanks!

docs/java/hello-world-tutorial.md
```diff
@@ -8,16 +8,17 @@ tags: helloworld, java, sdk, intellij, gradle, tutorial
...
 
 :::note Tutorial information
+
 - **Level:** ⭐ Temporal beginner
 - **Time:** ⏱️ ~20 minutes
 - **Goals:** 🙌
   - Learn how to set up, build, and test a Temporal application project from scratch using the [Java SDK](https://github.com/temporalio/java-sdk).
   - Become more familiar with core concepts and the application structure.
-:::
+    :::
```
flossypurse, line 17: Make sure to revert this indent, as it causes the whole page to turn into a note block.

docs/java/run-your-first-app-tutorial.md
```diff
@@ -9,14 +9,15 @@ import { ResponsivePlayer } from '../../src/components'
...
+
 - **Level**: ⭐ Temporal beginner
 - **Time**: ⏱️ ~20 minutes
 - **Goals**: 🙌
   - Complete several runs of a Temporal Workflow application using the Temporal server and the [Java SDK](https://github.com/temporalio/java-sdk).
   - Practice reviewing the state of the Workflow.
   - Understand the inherent reliability of Workflow functions.
   - Learn many of Temporal's core terminology and concepts.
-:::
+    :::
```
flossypurse, line 20: Here as well. Need to revert this -

flossypurse (approved):
Excellent!

flossypurse:
not building due to #Large-Event-Histories link -

tsurdilo:
rebased

PR #390, opened 2021-04-28
flesh out workers concept doc a little

sw-yx:
## Why?

existing doc doesnt really say much.

PR #408, opened 2021-05-06
replace Workers system diagram with something better

sw-yx:
## What was changed:

#390 introduced a potentially confusing diagram... this reverts it and/or fixes it, depending if @tsurdilo has an alternative image he'd like to use


flossypurse (approved):

flossypurse:
Approving to remove.
Possible @dtornow already has something even somewhat basic that could help here?

PR #415, opened 2021-05-07
Consolidate PHP Workflow & Activities info

flossypurse:
## What was changed:

Per our docs plan, PHP information has been consolidated into Workflows and Activities pages.

## Why?

Doing this to strengthen our core concept documentation.

## Checklist

1. Closes issue: 

N/A

2. How was this tested:

Ran `netlify build --context deploy-preview` locally to build and run link checks.

3. Any docs updates needed?

N/A


flossypurse:
I started to "refresh" the content, but then realized that was a lot more work than I have time for right now.
So this pretty much just consolidates the content that already existed into either Workflows or Activities pages.

I did try to make sure that each sentence was on a newline to make future changes + reviews easier.

sw-yx:
we can move more and more stuff into the `/shared` folder so that refreshing content is leveraged across all SDKs. just a thought

sw-yx (approved):

docs/php/activities.md
```diff
@@ -3,3 +3,309 @@ id: activities
...
+}
+```
+
+### Activity Heart Beating
+
+Some Activities are long-running.
+To react to a crash quickly, use a heartbeat mechanism.
+`Activity::heartbeat()` lets the Temporal service know that the Activity is still alive.
+You can piggyback `details` on an Activity heartbeat.
+If an Activity times out, the last value of `details` is included in the `TimeoutFailure` delivered to a Workflow.
```
mfateev, line 161: I believe it is also available inside the activity through Activity::[getHeartbeatDetails](https://github.com/temporalio/sdk-php/blob/f53f1627f63fca73bd0e533831a8104bacd44309/src/Activity.php#L90). When activity is restarted after failure getHeartbeatDetails allows to learn value from the last successful heartbeat.

Ideally, we should modify the sample to include getHeartbeatDetails.

docs/php/activities.md
```diff
@@ -3,3 +3,309 @@ id: activities
...
+        // ...
+    }
+
+    // ...
+}
+```
+
+## Calling Activities
+
+`Workflow::newActivityStub` returns a client-side stub that implements an Activity interface.
```
mfateev, line 198: client-side (one that can be used from the workflow code)

docs/php/activities.md
```diff
@@ -3,3 +3,309 @@ id: activities
...
+It takes Activity type and Activity options as arguments.
+Activity options are needed only if some of the required timeouts are not specified through the `#[ActivityMethod]` annotation.
```
mfateev, line 200: this comment is outdated as ActivityMethod doesn't support any timeouts.

docs/php/activities.md
```diff
@@ -3,3 +3,309 @@ id: activities
...
+
+Calling (via `yield`) a method on this interface invokes an Activity that implements this method.
```
mfateev, line 202: This is not really correct. Calling a method on this interface returns a Promise without blocking for activity completion. Calling yield on that future blocks waiting for the activity completion. 

I think having two sections (as we do in the Java version of the docs) for sync and async doesn't make sense for PHP when the call always returns a Promise.

docs/php/activities.md
```diff
@@ -3,3 +3,309 @@ id: activities
...
+        Activity::heartbeat(100); // download complete
+
+        // ...
+    }
+
+    // ...
+}
+```
+
+## Calling Activities
```
mfateev, line 196: I think calling activities section belongs to workflows, not to the activities part of the documentation.

docs/php/activities.md
flossypurse: I took a task to track this change.

@wolfy-j any chance we could sync on this section?

PR #431, opened 2021-05-14
Breaking up Java quick-start, surface Java Signals and Queries

flossypurse:
## What does this PR do?

This PR does the following:
- Breaks up Java quick-start content into appropriate places and deletes the page.
- Surfaces Java Signals and Queries.
- Refreshes subsections of content (Java Activities, CLI )
- Sentences on newlines where I found them.


docs/system-tools/tctl.md
```diff
@@ -4,42 +4,60 @@ title: Temporal CLI - tctl
 sidebar_label: CLI
 ---
 
-The Temporal CLI is a command-line tool you can use to perform various tasks on a Temporal server. It can perform
-namespace operations such as register, update, and describe as well as workflow operations like start
+The Temporal CLI is a command-line tool you can use to perform various tasks on a Temporal server.
+It can performm namespace operations such as register, update, and describe as well as workflow operations like start
```
sw-yx, line 8:
```suggestion
It can perform namespace operations such as register, update, and describe as well as workflow operations like start
```

docs/java/activities.md
```diff
@@ -4,24 +4,66 @@ title: Activities in Java
...
-Java SDK programming model are classes which implement an Activity Interface:
+## What is an Activity?
+
+Having fault tolerant code that maintains state is useful.
+But in most practical applications, a Workflow will be expected to act upon the external world.
+Temporal supports such externally-facing code in the form of Activities.
+
+An Activity is essentially a function that can execute any code like DB updates or service calls.
+The Workflow is not allowed to directly call any external APIs;
+It can do this only through Activities.
```
sw-yx, line 15:
```suggestion
it can do this only through Activities.
```

docs/java/activities.md
```diff
@@ -4,24 +4,66 @@ title: Activities in Java
...
+The Workflow is essentially an orchestrator of Activities.
+
```
sw-yx, line 17:
```suggestion
Activities that fail or timeout will be retried; therefore they should be written with care to be idempotent.
```

docs/java/activities.md
```diff
@@ -4,24 +4,66 @@ title: Activities in Java
...
+```java
+@ActivityInterface
+public interface GreetingActivities {
+    String composeGreeting(String greeting, String name);
+}
+```
+
+An Activity interface must be annotated with the `@ActivityInterface`.
+Each method that belongs to an Activity interface defines a separate Activity type.
```
sw-yx, line 30:
```suggestion
Each method that belongs to an `@ActivityInterface` defines a separate Activity type.
```

docs/java/activities.md
```diff
@@ -4,24 +4,66 @@ title: Activities in Java
...
 ```
 
-Activity interface must be annotated with the `@ActivityInterface`. You can annotate each method in the
-Activity interface with the `@ActivityMethod` annotation, but this is completely optional.
-Note that the `@ActivityMethod` annotation has a `name` parameter which can be used to define
-the Activity type.
-If not specified, the method name (with the first letter capitalized) is used by default, so in 
-the above example it would be "ComposeGreeting".
+Note that the `@ActivityMethod` annotation has a `name` parameter which can be used to define the Activity type.
+If not specified, the method name (with the first letter capitalized) is used by default, so in the above example it would be  ComposeHello. If the annotation did not specify the name then the Activity Type would be "ComposeGreeting".
```
sw-yx, line 42:
```suggestion
If not specified, the method name (with the first letter capitalized) is used by default, so in the above example it would be `ComposeHello`. If the annotation did not specify the name then the Activity Type would be `ComposeGreeting`.
```

docs/java/signals.md
```diff
@@ -3,3 +3,61 @@ id: signals
 title: Signals in Java
 sidebar_label: Signals
 ---
+
+Workflows can listen to external events and save the state accordingly.
```
sw-yx, line 7: should we have something in samples-java to show signals @tsurdilo? https://github.com/temporalio/samples-java

sw-yx (approved):
lgtm, left some comments and want tiho to approve too

docs/java/activities.md
flossypurse: Oh, ya my b.

docs/java/activities.md
```diff
@@ -4,24 +4,66 @@ title: Activities in Java
...
-Activities can be invoked during Workflow execution. Similar to Workflows, Activities in Temporal
-Java SDK programming model are classes which implement an Activity Interface:
+## What is an Activity?
+
+Having fault tolerant code that maintains state is useful.
+But in most practical applications, a Workflow will be expected to act upon the external world.
+Temporal supports such externally-facing code in the form of Activities.
+
+An Activity is essentially a function that can execute any code like DB updates or service calls.
+The Workflow is not allowed to directly call any external APIs;
```
flossypurse, line 14:
```suggestion
The Workflow is not allowed to directly call any external APIs; it can do this only through Activities.
```

docs/java/activities.md
```diff
@@ -4,24 +4,66 @@ title: Activities in Java
...
-Java SDK programming model are classes which implement an Activity Interface:
+## What is an Activity?
+
+Having fault tolerant code that maintains state is useful.
+But in most practical applications, a Workflow will be expected to act upon the external world.
+Temporal supports such externally-facing code in the form of Activities.
+
+An Activity is essentially a function that can execute any code like DB updates or service calls.
+The Workflow is not allowed to directly call any external APIs; it can do this only through Activities.
+It can do this only through Activities.
```
flossypurse, line 15:
```suggestion
```

docs/java/activities.md
```diff
@@ -4,24 +4,65 @@ title: Activities in Java
...
+
+Having fault tolerant code that maintains state is useful.
+But in most practical applications, a Workflow will be expected to act upon the external world.
+Temporal supports such externally-facing code in the form of Activities.
+
+An Activity is essentially a function that can execute any code like DB updates or service calls.
+The Workflow is not allowed to directly call any external APIs; it can do this only through Activities.
+The Workflow is essentially an orchestrator of Activities.
+Activities that fail or timeout will be retried; therefore they should be written with care to be idempotent.
+### Activity interface
```
tsurdilo, line 17: Add space above?

docs/java/activities.md
```diff
@@ -4,24 +4,65 @@ title: Activities in Java
 sidebar_label: Activities
 ---
 
-Activities can be invoked during Workflow execution. Similar to Workflows, Activities in Temporal
-Java SDK programming model are classes which implement an Activity Interface:
+## What is an Activity?
```
tsurdilo, line 7: Would it be ok to rewrite this paragraph some? If so let me know I can provide some suggestions

docs/java/activities.md
```diff
@@ -4,24 +4,65 @@ title: Activities in Java
...
+```
+
+An Activity interface must be annotated with the `@ActivityInterface`.
+Each method that belongs to an `@ActivityInterface` defines a separate Activity type.
+You can annotate each method in the Activity interface with the `@ActivityMethod` annotation, but this is completely optional.
 
 ```java
 @ActivityInterface
 public interface GreetingActivities {
+    @ActivityMethod(name=composeHello)
```
tsurdilo, line 35: this is not needed

docs/java/activities.md
```diff
@@ -4,24 +4,65 @@ title: Activities in Java
...
-If not specified, the method name (with the first letter capitalized) is used by default, so in 
-the above example it would be "ComposeGreeting".
+Note that the `@ActivityMethod` annotation has a `name` parameter which can be used to define the Activity type.
+If not specified, the method name (with the first letter capitalized) is used by default, so in the above example it would be `ComposeHello`. If the annotation did not specify the name then the Activity Type would be `ComposeGreeting`.
+
+### Activity implementation
 
-### Calling Activities inside Workflows
+An Activity implementation is just a normal [POJO](https://en.wikipedia.org/wiki/Plain_old_Java_object).
+The `out` stream is passed as a parameter to the constructor to demonstrate that the Activity object can have any dependency.
```
tsurdilo, line 46: This is because when registering an activity with worker you use the activity instance, and not the type like with Workflows. maybe we can add

docs/java/queries.md
```diff
@@ -3,3 +3,53 @@ id: queries
...
+  @WorkflowInterface
+  public interface HelloWorld {
+    @WorkflowMethod
+    void sayHello(String name);
+
+    @QueryMethod
+    int getCount();
+  }
+```
+
```
tsurdilo, line 21: Similar to the activities section, QueryMethod also has a "name" parameter. We should maybe describe it as well

docs/java/queries.md
```diff
@@ -3,3 +3,53 @@ id: queries
...
+    }
+  }
+```
+
+The restrictions on the implementation of the Query method are the following:
+
+- It can not modify Workflow state in any way.
+- It is not allowed to block its thread in any way.
+
+The Query method usually just returns a value derived from the fields of the Workflow object.
```
tsurdilo, line 55: We are missing here that query methods can also take in parameters which can help with things like limiting whats is returned. maybe we should add an example of that or expand existing

docs/java/signals.md
```diff
@@ -3,3 +3,61 @@ id: signals
...
+
+```java
+import io.temporal.workflow.SignalMethod;
```
tsurdilo, line 10: maybe we can use just the main class here and not include the entire class with includes etc, like we do in other exaples. wdyt?

docs/java/signals.md
```diff
@@ -3,3 +3,61 @@ id: signals
...
+```
+
+The Workflow interface now has a method annotated with @SignalMethod.
+It is a callback method that is invoked every time a new Signal of `HelloWorld.updateGreeting` is delivered to the Workflow.
+The Workflow interface can have only one @WorkflowMethod which is the main function of the Workflow, but can have as many Signal methods as needed.
+
+This Workflow implementation demonstrates a few important Temporal concepts:
+
+- The Workflow is stateful and can have fields of any complex type.
+- `Workflow.await` blocks until it receives a parameter that evaluates to true.
```
tsurdilo, line 62: I think this should be "until the expression in the "away" is evaluated to true"

tsurdilo:
@flossypurse left some comments. is it OK with you if i update the sections directly in this pr ?

docs/java/activities.md
flossypurse: Of course.

sw-yx:
maybe use the suggested changes feature in github so that you save cully some time

tsurdilo:
@flossypurse made some updates . let me know if you think is ok

docs/java/queries.md
tsurdilo: added

docs/java/queries.md
tsurdilo: added

docs/java/activities.md
tsurdilo: added

docs/java/activities.md
tsurdilo: removed

docs/java/activities.md
tsurdilo: updated. thanks!

docs/java/activities.md
tsurdilo: added

docs/java/activities.md
```diff
@@ -4,24 +4,81 @@ title: Activities in Java
 sidebar_label: Activities
 ---
 
-Activities can be invoked during Workflow execution. Similar to Workflows, Activities in Temporal
-Java SDK programming model are classes which implement an Activity Interface:
+## What is an Activity?
+
+Activities are implementations of certain tasks which need to be performed during workflow execution.
```
flossypurse, line 9:
```suggestion
Activities are implementations of certain tasks which need to be performed during a Workflow execution.
```

docs/java/activities.md
```diff
@@ -30,10 +87,10 @@ It takes Activity type and Activity options as arguments.
 Activity options allow you to specify different Activity timeout and retry options.
 
 Calling a method on this interface invokes an Activity that implements this method.
-An Activity invocation synchronously blocks until the Activity completes, fails, or times out. Even if Activity
-execution takes a few months, the Workflow code still sees it as a single synchronous invocation.
-It doesn't matter what happens to the processes that host the Workflow. The business logic code
-just sees a single method call.
+An Activity invocation synchronously blocks until the Activity completes, fails, or times out.
+Even if Activity execution takes a few months, the Workflow code still sees it as a single synchronous invocation.
```
flossypurse, line 91:
```suggestion
Even if an Activity's execution takes a few months, the Workflow code still sees it as a single synchronous invocation.
```

PR #434, opened 2021-05-17
Update Node docs for TLS and package initializer options

bergundy:
At some point before we hit beta I'll restructure the node docs, for now I'm accumulating material while it's still fresh in my mind.


sw-yx (approved):
great, keep them updated

PR #440, opened 2021-05-18
improve readability of single blog display

sw-yx:
## What does this PR do?

left sidebar was too noisy and distracting. moved it down to footer

before

image

after

image

## Notes to reviewers

- i also handled the mobile view
- if there's only one subheading, we do not display table of contents now
- shift top nav to left for more intuitive discovery


rylandg (approved):
LGTM

PR #448, opened 2021-05-24
Add node about never using dryrun option

tsurdilo:

## What does this PR do?
Using this option is dangerous. We have ran into issues with this and removed it, but this will take effect in next tools version. We need to warn users to not use this option to not lose data.

https://github.com/temporalio/temporal/pull/1300

### DIFF
--- a/docs/server/versions-dependencies.md
+++ b/docs/server/versions-dependencies.md
@@ -89,6 +89,10 @@ To use a more recent version of Temporal, first [check to see](https://github.co
 
 We ensure that any consecutive versions are compatible in terms of database schema upgrades, features, and system behavior, however there is no guarantee that there is compatibility between *any* 2 non-consecutive versions. Please reach out to us or check the forums at [community.temporal.io](http://community.temporal.io) for more information.
 
+We strongly recommend *never* using the "dryrun" (`-y`, or `--dryrun`) options in any of your schema update commands.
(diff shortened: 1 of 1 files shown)


flossypurse (approved):

docs/server/versions-dependencies.md
```diff
@@ -89,6 +89,10 @@ To use a more recent version of Temporal, first [check to see](https://github.co
 
 We ensure that any consecutive versions are compatible in terms of database schema upgrades, features, and system behavior, however there is no guarantee that there is compatibility between *any* 2 non-consecutive versions. Please reach out to us or check the forums at [community.temporal.io](http://community.temporal.io) for more information.
 
+We strongly recommend *never* using the "dryrun" (`-y`, or `--dryrun`) options in any of your schema update commands.
+Using this option might lead to potential loss of data, as when using this will create a new database and drop your
+existing one.
+
```
mfateev, line 95: I believe it was removed from some version. Should we mention that?

tsurdilo: https://github.com/temporalio/temporal/pull/1300

how can i tell which version of schema tools this was included in ?

tsurdilo: updated. it was 1.8.0. thanks

PR #449, opened 2021-05-24
add section about testing queries

vkarpov15:
## What does this PR do?

I added a section about testing queries to Go testing docs with a nice example. Also made a note about https://github.com/temporalio/sdk-go/issues/433.

### DIFF
--- a/docs/go/testing.md
+++ b/docs/go/testing.md
@@ -217,21 +217,22 @@ Note that you should always query the Workflow either after `ExecuteWorkflow()` 
 func (s *UnitTestSuite) Test_ProgressWorkflow() {
 	value := 0
 
-	// After 10 seconds plus some padding, the progress should be 10.
+	// After 10 seconds plus padding, progress should be 10.
 	// Note that `RegisterDelayedCallback()` doesn't actually make your test wait for 10 seconds!
-	// This test should take far less than 1 second.
+	// Temporal's test framework advances time internally, so this test should take < 1 second.
 	s.env.RegisterDelayedCallback(func() {
 		res, err := s.env.QueryWorkflow("getProgress")
 		s.NoError(err)
 		err = res.Get(&value)
 		s.NoError(err)
 		s.Equal(10, value)
-	}, time.Second*10+time.Millisecond*2)
+	}, time.Second*10+time.Millisecond*1)
 
 	s.env.ExecuteWorkflow(ProgressWorkflow, 0)
 
 	s.True(s.env.IsWorkflowCompleted())
 
+	// Once the workflow is completed, progress should always be 100
 	res, err := s.env.QueryWorkflow("getProgress")
 	s.NoError(err)
 	err = res.Get(&value)
--- a/docs/go/testing.md
+++ b/docs/go/testing.md
@@ -202,6 +202,8 @@ func ProgressWorkflow(ctx workflow.Context, percent int) error {
 	}
 
 	for percent = 0; percent<100; percent++ {
+                // Important! Use `workflow.Sleep()`, not `time.Sleep()`, because Temporal's
+                // test environment doesn't stub out `time.Sleep()`.
 		workflow.Sleep(ctx, time.Second*1)
 	}
 
--- a/docs/go/testing.md
+++ b/docs/go/testing.md
@@ -186,7 +186,7 @@ example, we assert that the `value` param has the same content as the value para
 
 ## Queries
 
-`TestWorkflowEnvironment` instances have a [`QueryWorkflow()` method](https://pkg.go.dev/go.temporal.io/temporal/internal#TestWorkflowEnvironment.QueryWorkflow) that lets you query the currently running Workflow.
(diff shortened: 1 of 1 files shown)


docs/go/testing.md
```diff
@@ -183,3 +183,59 @@ function matches the signature of the original Activity function.
...
+
+	err := workflow.SetQueryHandler(ctx, "getProgress", func(input []byte) (int, error) {
+		return percent, nil
+	})
+	if err != nil {
+		logger.Info("SetQueryHandler failed.", "Error", err)
+		return err
+	}
+
+	for percent = 0; percent<100; percent++ {
```
tsurdilo, line 204: Wondering why you need this loop. I think you can just change the value before returning and just query it in your test after workflow execution, or you can set it and wait for a signal in your test. 
This way your test depends on this 10 seconds + 2 milliseconds thing which can fail on some slow machines imo.

docs/go/testing.md
```diff
@@ -183,3 +183,59 @@ function matches the signature of the original Activity function.
...
+	// After 10 seconds plus some padding, the progress should be 10.
+	// Note that `RegisterDelayedCallback()` doesn't actually make your test wait for 10 seconds!
+	// This test should take far less than 1 second.
+	s.env.RegisterDelayedCallback(func() {
+		res, err := s.env.QueryWorkflow("getProgress")
+		s.NoError(err)
+		err = res.Get(&value)
+		s.NoError(err)
+		s.Equal(10, value)
+	}, time.Second*10+time.Millisecond*2)
```
tsurdilo, line 229: I don't think this is a good idea to do in a test. If people run this with a slow machine or have some network issues during test exec, the s.Equal(10, value) could potentially fail.

docs/go/testing.md
```diff
@@ -183,3 +183,59 @@ function matches the signature of the original Activity function.
...
+
+	s.env.ExecuteWorkflow(ProgressWorkflow, 0)
+
+	s.True(s.env.IsWorkflowCompleted())
+
+	res, err := s.env.QueryWorkflow("getProgress")
+	s.NoError(err)
+	err = res.Get(&value)
+	s.NoError(err)
+	s.Equal(value, 100)
```
tsurdilo, line 239: you should probably add comment here that you are querying the workflow after its in the Completed status so its clear to readers

docs/go/testing.md
```diff
@@ -183,3 +183,59 @@ function matches the signature of the original Activity function.
...
+Below is how you would write a test case that queries this Workflow.
+Note that you should always query the Workflow either after `ExecuteWorkflow()` is done or in a `RegisterDelayedCallback()` callback, otherwise you'll get a `runtime error` panic.
+
+```go
+func (s *UnitTestSuite) Test_ProgressWorkflow() {
+	value := 0
+
+	// After 10 seconds plus some padding, the progress should be 10.
+	// Note that `RegisterDelayedCallback()` doesn't actually make your test wait for 10 seconds!
+	// This test should take far less than 1 second.
```
tsurdilo, line 222: can you elaborate why it will take less than 1 second so users understand? is it that you are using a test api from sdk that advances time? from readers perspective they will still execute a workflow which has a loop with a 1 second delay in it

docs/go/testing.md
vkarpov15: I was trying to get an example of a workflow that has multiple intermediate states, not just 3 states like this example: https://github.com/temporalio/samples-go/blob/master/query/query_workflow.go. Think of it like a progress bar for an async operation, where the progress bar increments by 1% every second.

Sushisource: Agreed, you don't want to rely on time for the test to pass. However, I like the idea of the progress loop, it's a fun example for query.

The other way to fix this is to change the test not to rely on a certain amount of time passing, but to simply keep querying in a loop (usually with a small sleep) until the value reaches 100. 

It's generally a good approach for this sort of thing. Don't wait the time you expect, but wait until the condition you expect is fulfilled (and, if time matters, which it doesn't really here, assert that reaching that condition didn't take too long)

docs/go/testing.md
vkarpov15: I'm not entirely sure why the `time.Millisecond*2` is necessary. But I'm not entirely sure whether a slow machine or network issues could affect this because `RegisterDelayedCallback()` doesn't use wall clock time, the test framework advances time on its own. I asked Spencer to take a look.

docs/go/testing.md
vkarpov15: Good suggestion. TBH I don't know the internals very well, but the test framework advances time so that you don't actually have to wait 10 seconds of wall clock time for `RegisterDelayedCallback()` to run.

docs/go/testing.md
Sushisource: edit: Yeah, I forgot these were time-skipping tests. The concern isn't really important in these ones.

Sushisource (approved):

flossypurse (approved):

tsurdilo (approved):

tsurdilo:
@vkarpov15 thanks for making the updates :) +1

PR #452, opened 2021-05-26
add syntax for how to import the sleep() function

vkarpov15:
## What does this PR do?

Added a quick note about how to import the `sleep()` function from the Node SDK. Not sure where else you would figure this out other than reading the code.


flossypurse (approved):

bergundy (approved):
Thanks!

PR #459, opened 2021-06-02
add warning labels to experimental features

sw-yx:
## What does this PR do?

add experimental feature warnings

## Notes to reviewers

this is the basic idea for how labels should work - now that we have this foundation we can be more creative with how these things work, lets talk design ideas when you have them

@tsurdilo  - i have not added version tags for now, let me know what sections you want to have version tags for

### DIFF
--- a/docs/shared/CustomWarning.js
+++ b/docs/shared/CustomWarning.js
@@ -1,0 +1,54 @@
+import React from "react";
+
+export default function CustomWarning({children}) {
+  return (
+    <div
+      style={{
+        backgroundColor: "var(--ifm-color-warning)",
+        borderColor: "var(--ifm-color-warning)",
+        borderRadius: "var(--ifm-alert-border-radius)",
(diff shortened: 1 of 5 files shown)


docs/shared/CustomWarning.js
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+
+export default function CustomWarning({children}) {
```
sw-yx, line 3: the reason we need a custom component is that the default admonition is hard to read in light mode:

CleanShot 2021-06-02 at 18 00 22

here it is with the custom component

CleanShot 2021-06-02 at 18 00 47


this is a known issue for docusaurus but they are still working on it https://github.com/facebookincubator/infima/issues/55

flossypurse (approved):

rylandg (approved):
LGTM. That being said its not clear at all that I can click and expand those. I actually only figured it out because I saw your screenshot.

PR #465, opened 2021-06-08
Document Node SDK External Dependencies

bergundy:
See also:

[#118 feat: Support injecting external dependencies into the Workflow isolate](https://github.com/temporalio/sdk-node/pull/118)
[Proposal](https://github.com/temporalio/proposals/blob/master/node/logging-and-metrics-for-user-code.md)


docs/node/workflow-external-dependencies.md
```diff
@@ -0,0 +1,57 @@
...
+:::caution
+This is an advanced feature and should be used with care.
+:::
+
+Workflows in Temporal may be replayed from the beginning of their history when resumed.
+In order for Temporal to recreate the exact state Workflow code was in, the code is required to be [fully deterministic](/docs/node/determinism).
+To prevent breaking determinism, in the Node SDK, Workflow code runs in an isolated execution environment limited to functionality provided by the SDK.
+
+External Dependencies is an isolation breaking mechanism that allows injecting functions from the main NodeJS environment into a Workflow isolate.
+They are typically used in order to inject custom instrumentation (e.g. logger) functions into the isolate.
```
sw-yx, line 12: interesting - isn't logging a common workflow, rather than an advanced one? dont we expect every production user to have to use this?

docs/node/workflow-external-dependencies.md
```diff
@@ -0,0 +1,57 @@
+# External Dependencies
```
sw-yx, line 1: i like the idea of calling this "advanced" section to make clear it is optional. also we should put all these `workflow-*` docs into one page, to be inline with the other 3 SDK's

docs/node/workflow-external-dependencies.md
```diff
@@ -0,0 +1,57 @@
...
+The following configuration options are for controlling how an injected function is executed.
+
+### `callDuringReplay`
+
+A boolean controling whether or not the injected function will be called during Workflow replay - defaults to `false`.
+
+### ApplyMode
+
+The different modes for an injected function to be applied to the isolate are documented in the [API reference](https://nodejs.temporal.io/api/enums/worker.applymode).
+
```
sw-yx, line 25: i like spelling it out for people who are searching and people who are reading the next seciton

```suggestion
The different modes for an injected function to be applied to the isolate are documented in the [API reference](https://nodejs.temporal.io/api/enums/worker.applymode).

- `ASYNC`
- `ASYNC_IGNORED`
- `SYNC`
- `SYNC_IGNORED`
- `SYNC_PROMISE`

```

docs/node/workflow-external-dependencies.md
bergundy: Agreed, I want to reorganize these pages at soon.

docs/node/workflow-external-dependencies.md
bergundy: Agreed, the doc is inaccurate.

Using `ASYNC_IGNORED` is completely safe and should commonly be used to propagate logs and metrics out of a Workflow.

I'll clarify that only external dependencies that return a value are dangerous to use.

bergundy (commented):
Thanks, you've brought up a good point, I'll redo this page.

docs/node/workflow-external-dependencies.md
bergundy: For now I put all of the workflow related pages in a `Workflows` category.

sw-yx (approved):

PR #466, opened 2021-06-08
Project Jargon Mesh

flossypurse:
This is a DRAFT WIP!


docs/jargon-mesh.md
```diff
@@ -0,0 +1,658 @@
...
+id: terminology
+title: Temporal's core terminology
+sidebar_label: Terminology
+---
+
+import Mermaid from '@theme/Mermaid';
+
+## Workflow
+
+A fault-oblivious stateful function that orchestrates Activities.
```
tsurdilo, line 11: This up-front is to me confusing as:
1) Workflow execution **can** fail
2) User can write Workflow that invokes 0 Activities

docs/jargon-mesh.md
```diff
@@ -0,0 +1,658 @@
...
+
+- A Workflow has full control over which [**Activities**](#activity) are executed, and in which order.
+- A Workflow must not affect the external world directly, only through [**Activities**](#activity).
```
tsurdilo, line 14: IMO only, this is a pretty hard requirement as
we do have sections of SideEffects
I would lower this to a "recommendation" honestly

docs/jargon-mesh.md
flossypurse: Not remotely ready for feedback.

docs/jargon-mesh.md
```diff
@@ -0,0 +1,658 @@
...
+- What makes Workflow code a Workflow is that its state is preserved by Temporal.
```
tsurdilo, line 15: Are we with this saying that it i impossible to write "stateless" workflows with Temporal?

docs/jargon-mesh.md
```diff
@@ -0,0 +1,658 @@
...
+  Therefore any failure of a [**Worker**](#worker) process that hosts the Workflow code does not affect the [**Workflow Execution**](#workflow-execution).
+  The Workflow continues as if these failures did not happen.
+  At the same time, [**Activities**](#activity) can fail any moment for any reason.
+- Because Workflow code is fully fault-oblivious, it is guaranteed to get notifications about [**Activity**](#activity) failures or timeouts and act accordingly.
+- There is no limit to the duration of a Workflow Execution.
```
tsurdilo, line 20: I would maybe say something like "By default, workflow execution is umlimited, however you can set the workflow execution timeout to limit it" .. so user understands its not always "no limit"

tsurdilo: Also, even if it set to "unlimited" it can fail depending on some issues like event log size (nice way to link from here to the continueasnew section ;))

PR #470, opened 2021-06-09
add signup form on docs item

sw-yx:
## What does this PR do?

add signup form on docs item #emailmarketing

CleanShot 2021-06-10 at 04 11 13


flossypurse:
Small styling nit - "submit" button hover colors aren't very visible.
On light mode it turns white.
On dark mode it turn black.

Also, there is no border on the button in the preview:

Screen Shot 2021-06-10 at 10 00 31 AM

Otherwise 👍

sw-yx:
latest commit should fix!

flossypurse:
This happened to me when I was working with the buttons before as well - 
nit - need some spacing.
But, its Sand!

Screen Shot 2021-06-10 at 11 32 09 AM

flossypurse (approved):

PR #471, opened 2021-06-10
Remove duplicate video

stevekinney:
## What does this PR do?

We've got the same video from Ryland listed twice. This PR gets rid of the second instance.


flossypurse (approved):
👍

PR #473, opened 2021-06-11
add docs about activities in Node

vkarpov15:
## What does this PR do?

Added some learnings about using the Node SDK from working on [samples-node](https://github.com/vkarpov15/samples-node). There wasn't a doc on activities yet, so I added it


docs/node/activities.md
```diff
@@ -0,0 +1,47 @@
+# Activities
+
+In Temporal, Activities are typically used to represent side effects, like making an HTTP request.
+Unlike [Workflows](/docs/node/logging#logging-from-workflows-and-activities), activities run in the standard Node.js environment, not an [isolate](https://www.npmjs.com/package/isolated-vm).
+So any code that needs to talk to the outside world, including importing modules, needs to be in an Activity.
```
bergundy, line 5: Not true, you can import any module in WFs as long as it doesn't use the node or dom libs.

docs/node/activities.md
```diff
@@ -0,0 +1,47 @@
+# Activities
+
+In Temporal, Activities are typically used to represent side effects, like making an HTTP request.
+Unlike [Workflows](/docs/node/logging#logging-from-workflows-and-activities), activities run in the standard Node.js environment, not an [isolate](https://www.npmjs.com/package/isolated-vm).
```
bergundy, line 4: Is there a better page we can link for workflows? Maybe the one about determinism?

docs/node/activities.md
```diff
@@ -0,0 +1,47 @@
...
+So any code that needs to talk to the outside world, including importing modules, needs to be in an Activity.
+
+### Overview
+
+Below is a simple Activity that accepts a string parameter, appends a word to it, and returns the result.
+The Temporal Node SDK looks for any `.js` files in the `src/activities` directory, and automatically registers any exported functions as Activities.
```
bergundy, line 10: Not accurate, in the package initializer, it's `.ts` in `src/activites` and the compiled code in `lib/activities` (configurable).

bergundy (commented):
I'd add a section that explain that when you import an activity function from a WF it gets replaced with a stub that looks like this (psuedocode):

```
export function greet(...args) {
  return scheduleActivity('greet', args);
}
```

docs/node/activities.md
vkarpov15: Worth it to always use `lib/activities` rather than `src/activities`? If you're not using package initializer, either seems to work because of what [`activitiesPath` resolves to](https://github.com/temporalio/sdk-node/blob/9176b27270f7da77972ceb18c9c624f12e11ee93/packages/worker/src/worker.ts#L329)

sw-yx (approved):
excellent, thank you

bergundy:
> I'd add a section that explain that when you import an activity function from a WF it gets replaced with a stub that looks like this (psuedocode):
> 
> ```
> export function greet(...args) {
>   return scheduleActivity('greet', args);
> }
> ```

I'm still missing this bit in the doc, but otherwise good to merge.
In the future we should save the user's TS/JS preference and use that for selecting the default shown tab.

docs/node/activities.md
```diff
@@ -0,0 +1,68 @@
+# Activities
+
+In Temporal, Activities are typically used to represent side effects, like making an HTTP request.
```
flossypurse, line 3: "side effect" is actually the name of an API used to do some "non-deterministic" logic inside the Workflow code (at least in some of the other languages). Using it here as a way to describe an Activity is not ideal I think.

```suggestion
In Temporal, Activities are typically used to interact with external resources, like making an HTTP request.
```

docs/node/activities.md
```diff
@@ -0,0 +1,68 @@
...
+Unlike [Workflows](/docs/node/determinism), activities run in the standard Node.js environment, not an [isolate](https://www.npmjs.com/package/isolated-vm).
```
flossypurse, line 4:
```suggestion
Unlike [Workflows](/docs/node/determinism), Activities execute in the standard Node.js environment, not an [isolate](https://www.npmjs.com/package/isolated-vm).
```

docs/node/activities.md
```diff
@@ -0,0 +1,68 @@
...
+You can also use `require()` as shown below.
+
+```javascript
+const {greet} = require("@activities/greeter");
+```
+
+:::info
+
+Currently, you need to use the `@activities` prefix.
+Do **not** import your Activity using the `'../activities/greeter'` path, otherwise your Activity will run in the same V8 isolate as your Workflow and your Activity will be subject to the same restrictions as your Workflow.
```
flossypurse, line 66:
```suggestion
Do **not** import your Activity using the `'../activities/greeter'` path, otherwise your Activity will execute in the same V8 isolate as your Workflow and your Activity will be subject to the same restrictions as your Workflow.
```

djmagee, line 66:
```suggestion
Do **not** import your Activity by using the `'../activities/greeter'` path; otherwise your Activity runs in the same V8 isolate as your Workflow, and your Activity is subject to the same restrictions as your Workflow.
```

vkarpov15: Thanks :+1:

vkarpov15:
@bergundy I addressed your comments in https://github.com/temporalio/documentation/pull/473/commits/cf897bb84519c48f68f60a553077b6dacae4a425, let me know what you think :+1:

docs/node/activities.md
```diff
@@ -0,0 +1,72 @@
...
+async function greet(name) {
+  return `Hello, ${name}!`;
+}
+exports.greet = greet;
+```
+
+</TabItem>
+</Tabs>
+
+Under the hood, the Temporal Node SDK uses [Webpack](https://webpack.js.org/) to bundle your Activities before running them.
```
bergundy, line 46: Activities don't get bundled, only Workflows.

bergundy (commented):
LGTM apart from the part about bundling activities.
Feel free to merge after that's fixed.

PR #486, opened 2021-06-21
Update to 'External Workflows' section in Go workflows.md

tsurdilo:

## What does this PR do?
* Fixes the code sample on the "External Workflows" section of the Go workflows.md page
* Adds a little bit of context to invoking external workflows section
*

### DIFF
--- a/docs/go/testing.md
+++ b/docs/go/testing.md
@@ -257,7 +257,7 @@ Because of this you can often encounter a `PanicError: Potential deadlock detect
 
 To alleviate this issue, you can set the TEMPORAL_DEBUG environment variable to `true` before debugging your Workflow Definition.
 
-::: note
+:::note
 
 Make sure to set TEMPORAL_DEBUG to true only during debugging.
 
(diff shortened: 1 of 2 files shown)


flossypurse (commented):
I may be wrong, but it seems like the only thing that makes a Workflow "External" (External Workflow) is that it is executed by its Workflow Type string name - 
So, this section is a little bit confusing and a bit overkill.

This should probably just be grouped under the first section https://docs.temporal.io/docs/go/workflows#how-to-start-a-workflow

Not after async vs sync & child -

tsurdilo:
@flossypurse ok. so you want me to move this whole section into "how to start a workflow" as a sub-section there?

flossypurse (approved):
Per our chat - you are clear to merge this to fix the technical inaccuracy -

Moving forward we will aim to cover the Workflow Type in the "how to execute a Workflow" section.
And we will aim to cover "external" APIs in the "How to interact with another Workflow Execution" section

PR #488, opened 2021-06-22
terminology -> jargon/mesh

flossypurse:
## What does this PR do?

- Converts "terminology" page into "jargon/mesh" page
- Updates glossary links


docs/jargon/mesh.md
```diff
@@ -1,12 +1,14 @@
 ---
-id: terminology
-title: Temporal terminology
-sidebar_label: Terminology
+id: mesh
+title: Temporal's Jargon Mesh
```
djmagee, line 3: I liked the idea of a subtitle that Dominik (I think) raised—something like "Jargon Mesh: An introduction to Temporal concepts". (It also gets around using the possessive form of the company name, which can become problematic from a trademark point of view.)

flossypurse: There is no "sub title" per se - I did add that as the first heading in the page below.

But I think something l agree that this is a perfectly good title:
"Jargon Mesh - An introduction to Temporal's core concepts" (has to be a dash, can't be a colon due to build system"

djmagee (approved):

sw-yx:
needs redirects

flossypurse:
Page was only a few days old - but sure.

PR #495, opened 2021-06-25
add Node task queues and workers docs

vkarpov15:
## What does this PR do?

Add Task Queues and Workers sections to Node docs to line up with other languages.


docs/node/task-queues.md
```diff
@@ -0,0 +1,58 @@
...
+id: task-queues
+title: Task Queues in Node
+sidebar_label: Task Queues
+---
+
+## What is a Task Queue?
+
+import SharedTaskQueuesBasic from '../shared/task-queues-basic.md'
+
+<SharedTaskQueuesBasic
```
bergundy, line 11: Not strictly related to this PR but the first paragraph is inaccurate:
> From a high level, we can say that a Task Queue is exactly what the name suggests. It is a "first-in-first-out" queue for Tasks, where a Task is the context needed to execute a chunk of code that alters the "state" of a Workflow.

What about activity task queues? Those should be mentioned here as well IMHO.

docs/node/task-queues.md
```diff
@@ -0,0 +1,58 @@
...
+workflowLink="/docs/java/workflows"
+workerLink="/docs/java/workers"
+/>
+
+## How to use Task Queues
+
+In Node, a Task Queue is represented in code by name, as a `string`.
+There are 3 places where the name of the Task Queue is supplied by the developer.
+
+1. When starting a Workflow, you must pass the `taskQueue` option to the `Connection#workflow()` function.
```
bergundy, line 21: Please link to the API reference for `Connection.workflow`.
Why did you use the `#` sign? is it to mark that this is an instance method?

docs/node/task-queues.md
```diff
@@ -0,0 +1,58 @@
...
+
+```typescript
+const workflow = connection.workflow("my-workflow", {
+  taskQueue: "my-task-queue",
+});
+
+const result = await workflow.start();
+```
+
+2. When creating a Worker, you must pass the `taskQueue` option to the `Worker.create()` function.
```
bergundy, line 31: Same here, link to the API reference please.

docs/node/task-queues.md
bergundy: This is the first use of `#` in the node docs, I don't mind using it as long as we're consistent.

docs/node/task-queues.md
```diff
@@ -0,0 +1,58 @@
...
+});
+
+async function main(): Promise<void> {
+  await greetWithCustomTaskQueue("World");
+}
+
+export const workflow = {main};
+```
+
+If you do not specify a `taskQueue`, then the Node SDK places Activity Tasks in the same Task Queue as the Workflow Task Queue.
```
bergundy, line 58: Maybe move this up above the sample so it is not overlooked?

docs/node/workers.md
```diff
@@ -0,0 +1,51 @@
+---
+id: workers
+title: Workers in Node
+sidebar_label: Workers
+---
+
+## What is a Worker?
+
+A Worker is a service that executes [Workflows](/docs/go/workflows) and [Activities](/docs/go/activities).
```
bergundy, line 9: Link to the relevant `node` pages instead of `go`.
Also is the term service accurate here? Maybe call it a process?

docs/node/workers.md
```diff
@@ -0,0 +1,51 @@
...
+Workers are run on user controlled hosts.
+You can use the `worker` package's `Worker` classs to create and run as many Workers as your use case demands, across any number of hosts.
+
+Workers poll Task Queues for Tasks, execute chunks of code in response to those Tasks, and then communicate the results back to the Temporal Server.
```
bergundy, line 13: Link to the task queues page please.

docs/node/workers.md
```diff
@@ -0,0 +1,51 @@
...
+id: workers
+title: Workers in Node
+sidebar_label: Workers
+---
+
+## What is a Worker?
+
+A Worker is a service that executes [Workflows](/docs/go/workflows) and [Activities](/docs/go/activities).
+Workers are run on user controlled hosts.
+You can use the `worker` package's `Worker` classs to create and run as many Workers as your use case demands, across any number of hosts.
```
bergundy, line 11:
```suggestion
You can use the `worker` package's `Worker` class to create and run as many Workers as your use case demands, across any number of hosts.
```

docs/node/workers.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. The `workDir`. The Node SDK will automatically register Activities from any `.js` files in `workDir + '/../activities'` and Workflows from any `.js` files in `workDir + '/../workflows'`.
+2. The `taskQueue` the Worker should poll.
+
+Below is an example of starting a Worker that polls the Task Queue named 'tutorial'.
+
+```typescript
+import {Worker} from "@temporalio/worker";
+
+main().catch((err) => console.log(err));
```
bergundy, line 29: Missing `process.exit(1)`

docs/node/workers.md
```diff
@@ -0,0 +1,51 @@
...
+
+```typescript
+async function main(): Promise<string> {
+  return "Hello, World!";
+}
+
+export const workflow = {main};
+```
+
+The `Worker.create()` call will automatically register a Workflow named 'example' that returns the string 'Hello, World'.
```
bergundy, line 51: Here would be a good place to mention how to create multiple workers in the same process and show how to customize the server connection (`Core.install()`)

bergundy (commented):
Thanks,
Those are great additions.

docs/node/task-queues.md
sw-yx: i personally think this is not understood by all js developers and may be confused with private class fields? but i'm not going to make a huge deal out of this really. linking the reference would be nice to have

docs/node/task-queues.md
vkarpov15: This one is exactly as written in the Go and Java docs, so I think we should discuss this issue separately. I agree that it's worth mentioning that Task Queues store both Tasks and Activities.

docs/node/task-queues.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. When starting a Workflow, you must pass the `taskQueue` option to the [Connection's `workflow()` method](https://nodejs.temporal.io/api/classes/client.connection#workflow).
+
+```typescript
+const workflow = connection.workflow("my-workflow", {
+  taskQueue: "my-task-queue",
+});
+
+const result = await workflow.start();
+```
```
bergundy, line 29: This is now outdated with the recent changes to the client package.
Let's remember to update this doc before we release the new package.

bergundy (approved):

PR #496, opened 2021-06-25
add proof of concept for switching between js and ts examples on node SDK docs

vkarpov15:
## What does this PR do?

MDX-based tabs for switching between TS and JS examples on the hello world part of the docs.

image

image

Using pattern from: https://docusaurus.io/docs/markdown-features/code-blocks#multi-language-support-code-blocks.


docs/node/hello-world.md
```diff
@@ -21,11 +21,35 @@ Activities are called from Workflows in order to run non-deterministic code.
...
 `src/activities/greeter.ts`
 
 <!--SNIPSTART nodejs-hello-activity {"enable_source_link": false}-->
 <!--SNIPEND-->
 
+</TabItem>
+
+<TabItem value="js">
+
+<!--SNIPSTART nodejs-js-hello-activity {"enable_source_link": false}-->
```
bergundy, line 46: We should add the name of the file here to match the typescript version.
(This applies to all JS snippets).

bergundy (commented):
Yes please, let's do this all over the place.

To do this properly let's remember the user's choice and change all of the tabs on the site.
I'd also port the comments from TS to the JS snippets.
We should also add a note that JS doesn't need to define interfaces.

snipsync.config.yaml
```diff
@@ -25,6 +25,8 @@ origins:
     repo: samples-java
   - owner: temporalio
     repo: temporal-ecommerce
+  - owner: vkarpov15
+    repo: samples-node
```
sw-yx, line 29: for temporal long term cleanliness we should only rely on samples repos we own here. TODO - move this to the official node repo

flossypurse:
> Yes please, let's do this all over the place.
> 
> To do this properly let's remember the user's choice and change all of the tabs on the site.

You can do this using the `groupId` field: 
https://docusaurus.io/docs/markdown-features/tabs#syncing-tab-choices

slorber:
FYI @phryneas build a remark plugin to do this automagically for you: https://github.com/phryneas/remark-typescript-tools

vkarpov15:
Thanks @slorber , we'll try this out :+1:

vkarpov15:
@slorber thanks for the suggestion, this plugin seems to work well for us with a couple of tweaks. I had to put in a small PR here: https://github.com/phryneas/remark-typescript-tools/pull/7 .

docusaurus.config.js
```diff
@@ -222,6 +224,51 @@ module.exports = {
...
+                    }
+                    node.value = "// @ts-nocheck\n" + node.value;
+                  }
+
+                  visit(tree, "code", visitor);
+                },
+              {},
+            ],
+            [
+              require("remark-typescript-tools").transpileCodeblocks,
```
bergundy, line 247: This is so nodejs :)

docusaurus.config.js
```diff
@@ -222,6 +224,51 @@ module.exports = {
...
+                function addTSNoCheck(tree) {
+                  // Disable TS type checking for any TypeScript code blocks.
+                  // This is because imports are messy with snipsync: we don't
+                  // have a way to pull in a separate config for every example
+                  // snipsync pulls from.
+                  function visitor(node) {
+                    if (!/^ts$/.test(node.lang)) {
+                      return;
+                    }
+                    node.value = "// @ts-nocheck\n" + node.value;
```
bergundy, line 239: What a hack, I wish there was a better way around it, can't think of one though.

bergundy (approved):
I looked at the preview but couldn't tell what this change is supposed to do.
Is this still WIP?
Also, I don't see `yarn.lock` updated in this PR even though new packages were installed.

bergundy:
Accidentally approved

docusaurus.config.js
phryneas: So you want to transpile, but not get any errors? 
I would also accept a PR for something like that over in the remark plugin.

vkarpov15: @phryneas there's a couple issues why I'm adding `// @ts-nocheck`: 1) setting up all the dependencies has proven to be tricky, especially since `@activities` can refer to different files depending on where it is being imported from, 2) some samples would be too verbose if we include all the boilerplate, so we'd like the option to include smaller snippets that may omit some code that the compiler would think is necessary.

phryneas: Oh, okay :)

Generally, you can also add some boilerplate-like code as files and include that from all your examples, but if it's tested already it'll probably be fine.

sw-yx:
lets try to get this tied up this week, setup a call if you need higher bandwidth to resolve the remaining issues

docusaurus.config.js
bergundy: I'm fine with this because we make sure to test the codebase.

snipsync.config.yaml
bergundy: Did we end up adding more samples? I couldn't find them here.
If we did and want to keep this origin I'd at least open an issue so it's tracked.

bergundy (changes requested):

bergundy:
The generated JS code isn't usable at the moment, first sample in hello world looks like this:
Let's play with tsconfig and see if it generates something that we can use.
image

vkarpov15:
@sw-yx I got the deploy preview working - it looks like the build issue was due to (1) missing tsconfig, and (2) netlify build directory so docusaurus couldn't find `./tsconfig.json`. Here's the working preview:

https://deploy-preview-496--mystifying-fermi-1bc096.netlify.app/docs/node/hello-world

snipsync.config.yaml
bergundy: @vkarpov15 please comment here

bergundy (approved):
Awesome! Please see my comments:
1 .The generated JS code uses es module imports instead of commonjs which needs to be enabled (https://nodejs.org/api/esm.html#esm_enabling) or transpiled, I'd change this.
2. The Activities page doesn't use this cool mechanism: https://deploy-preview-496--mystifying-fermi-1bc096.netlify.app/docs/node/activities/
3. We use a different prettier config here than the one the codebase uses, I'd rather unify.
   (see [linked source](https://github.com/temporalio/sdk-node/blob/main/packages/test/src/workflows/cancel-timer-immediately.ts) vs. [sample in docs](https://deploy-preview-496--mystifying-fermi-1bc096.netlify.app/docs/node/workflow-scopes-and-cancellation))

vkarpov15:
@bergundy I addressed comments 2-4:

2) Activities page now uses the built-in switch, but I had to add an example of using `greet()` because [TypeScript unfortunately always removes unused imports](https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-imports-being-elided-in-my-emit).
3) Looks like both the prettier GitHub action and [remark-typescript-tools' default post-processor](https://github.com/phryneas/remark-typescript-tools/blob/eb979b3b3e66df956d6c0210f27233809f7ebdaf/src/transpileCodeblocks/postProcessing.ts) ended up running on the TS code samples. I fixed that.
4) I removed the vkarpov15/samples-node repo from snipsync sources.

Re (1), the ESM vs CommonJS issue, it doesn't look like TypeScript will be able to handle this for us because if you set TypeScript's target to "CommonJS" it outputs ugly JS with `defineProperty(exports, "__esModule", { value: true })` and other junk we don't want in our JS code samples. I'm tinkering with Babel to see if I can generate CommonJS without mangling existing code formatting.

bergundy:
Okay, great progress!

vkarpov15:
@bergundy try out the latest preview, I put together some babel logic that converts some common import and export patterns into CommonJS. It works _almost_ perfectly.

image

The only quirk I can see is that sometimes babel ends up removing the newline between import and exports as shown below:

image

What do you think, worth taking the time to fix that newline issue?

.prettierignore
```diff
@@ -3,3 +3,4 @@
 blog
 .docusaurus
 .github
+docs/node/*.md
```
bergundy, line 6: I wouldn't do that, this applies to all of the markdown not just code samples

bergundy (approved):
Great progress @vkarpov15.

One thing that bugs me is that we use different styles in ts and js, for example double quotes for ts and single quotes for js and padding in curly braces in js with no padding in ts.
Let's just use single quotes everywhere and padding in curlies, that's the style we have in the `sdk-node` repo.

Might be worth trying to add back the spaces but AFAIC, it can happen in a later PR, just make sure to open an issue and track it.

`.prettierrc.json`
```
{
  "printWidth": 120,
  "singleQuote": true
}
```

.prettierignore
vkarpov15: Does our prettier config modify markdown, not just code samples? I might have missed something here but it seems like prettier just formats code samples in this repo.

bergundy: I’m pretty sure it does, that should be easy to confirm

vkarpov15: I confirmed that prettier does in fact modify markdown in addition to code samples: https://prettier.io/blog/2017/11/07/1.8.0.html .

To avoid prettier mangling our TS samples, we need to add `<!-- prettier-ignore -->` before each example.

vkarpov15:
@bergundy I switched to using regexps to make the output formatting more consistent. It won't be quite as robust as using Babel, but I think the formatting benefits and code simplicity outweigh these concerns.

vkarpov15:
@bergundy I switched to using regexps to make the output formatting more consistent. It won't be quite as robust as using Babel, but I think the formatting benefits and code simplicity outweigh these concerns.

.prettierignore
lorensr: This PR gets prettier doing single quotes and bracket spacing so you don't need to ignore: https://github.com/temporalio/documentation/pull/573

bergundy:
@vkarpov15 as discussed yesterday, let's use ES module syntax for the examples, you can drop the CommonJS conversion code, sorry you spent so much time on it.

vkarpov15:
@bergundy I removed the ESM -> CommonJS conversion logic, let me know if you have any further comments :+1:

package.json
```diff
@@ -26,7 +26,9 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-player": "^2.6.0",
-    "snipsync": "1.5.0"
+    "remark-typescript-tools": "1.0.9",
```
bergundy, line 29: Should yarn.lock be updated in this commit?

bergundy (approved):
Looks good, I've already approved this, approving again.
We should probably remove `prettier-ignore` in the blocks but it can wait for @lorensr's prettierrc change.

sw-yx:
@vkarpov15 for when you come back - this functionality broke at some point in the last 2 months. we had to comment out for now until someone has time to figure it out.

image

phryneas:
Did you change the target in the tsconfig? 
-- 
Sent from my Android phone with K-9 Mail. Please excuse my brevity.

PR #497, opened 2021-06-30
Added reference to code of conduct

flossypurse:


sw-yx (approved):
we should probably have our own CoC at some point but nice initiative

PR #501, opened 2021-07-02
Update signals.md

eastfisher:
fix function param

## What does this PR do?

fix AddReceive() func param


flossypurse (approved):
Thanks for the fix!

PR #515, opened 2021-07-14
Upgrade Docusaurus for new mobile UX

slorber:
## What does this PR do?

Handle feedbacks from @sw-yx.

Also updates the swizzled DocItem so that it renders the new mobile collapsible TOC

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -44,7 +44,7 @@ module.exports = {
             },
             {
               to: "/application-development",
-              activeBasePath: "/application-development",
+              activeBaseRegex: "(/application-development)|(/docs/(go|java|php|node))",
               label: "SDKs",
             },
             {
--- a/src/theme/DocItem/index.js
+++ b/src/theme/DocItem/index.js
@@ -10,11 +10,13 @@ import DocVersionBanner from "@theme/DocVersionBanner";
 import Seo from "@theme/Seo";
 import LastUpdated from "@theme/LastUpdated";
 import TOC from "@theme/TOC";
+import TOCCollapsible from '@theme/TOCCollapsible';
 import EditThisPage from "@theme/EditThisPage";
 import {MainHeading} from "@theme/Heading";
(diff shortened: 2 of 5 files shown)


src/theme/DocItem/index.js
```diff
@@ -73,6 +88,14 @@ function DocItem(props) {
                   </span>
                 </div>
               )}
+
+              {renderTocMobile && (
+                <TOCCollapsible
```
slorber, line 93: changes needed to render this

image

docusaurus.config.js
```diff
@@ -44,7 +44,7 @@ module.exports = {
             },
             {
               to: "/application-development",
-              activeBasePath: "/application-development",
+              activeBaseRegex: "(/application-development)|(/docs/(go|java|php|node))",
```
slorber, line 47: @sw-yx this permits to have the "SDKs" dropdown item to be highlighted when browsing any of the 4 SDKs. Docusaurus can't guess about those SDKs by itself, it just sees a regular page.

image


New mobile collapsible auto-expansion implemented in https://github.com/facebook/docusaurus/pull/5168:

image

(seen after navigating back to main sidebar menu)

slorber (commented):
@sw-yx I think both items reported in your video feedback (https://github.com/facebook/docusaurus/pull/4273#issuecomment-877360318) have been fixed.

Let me know if you see anything weird left. 

If you see other issues regarding "active sidebar items", it's probably worth opening a new issue

docusaurus.config.js
sw-yx: thats so smart! I never thought to use it like that.

sw-yx (approved):
this looks great!! thank you so much for the fix!

slorber:
thanks for deploying, should help capture last edge cases 🤪 

Actually just found one, and you'll want to backport these changes in your swizzled DocItem: https://github.com/facebook/docusaurus/pull/5172

PR #469, opened 2021-06-08
add example of passing a struct as an arg to a signal

vkarpov15:
Fix #283

## What does this PR do?

Add an example of passing a struct to a signal and decoding it using [mapstructure](https://github.com/mitchellh/mapstructure)

PR #523, opened 2021-07-20
update Signal docs for signalling with structs

si-stripe:
## What does this PR do?
small update to the Signal docs. From my testing, an external library like `mapstructure` was not required to decode into a struct, so I updated the docs to reflect this (unless this changed after 1.9.2). 

The recommendation to use an external library was made in this [PR](https://github.com/temporalio/documentation/pull/469) but it seems to be unnecessary (per the linked [Temporal forum question](https://github.com/temporalio/documentation/commit/a8923441195860e8eece94154104d4c5dc812071)). This was confusing when I wanted to pass a `struct` in a Signal and thought I needed to add another external dependency.

* `Receive` decodes structs without the need for an external library
* The code example has a return inside the anonymous function, but the anonymous function can't return

### DIFF
--- a/docs/go/signals.md
+++ b/docs/go/signals.md
@@ -75,7 +75,7 @@ You can find more example Signals usage in our [Go Samples](https://github.com/t
 ## Signal with structs
 
 You can also send structs in Signals, as long as the struct is [serializable](https://pkg.go.dev/go.temporal.io/sdk/converter#CompositeDataConverter.ToPayload).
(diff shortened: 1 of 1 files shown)


vkarpov15 (approved):
Works for me :+1:

flossypurse (approved):
Thank you for contributing!
Yes, mapstructure is not needed here - this makes the info much cleaner.

PR #540, opened 2021-07-25
Add max-width to email form input

lorensr:
As is currently, my eyes skip over the input because it feels like a section divider. After the change, it looks like:

image


flossypurse:
Perhaps locally the css is applied? The deploy preview does not show the same result on any of the pages I have checked -

sw-yx:
think he forgot to import the css module..

sw-yx (approved):

PR #543, opened 2021-07-27
Add page that compares traditional service function execution to Temporal Workflow Execution

djmagee:
## What does this PR do?

**Preview:** https://deploy-preview-543--mystifying-fermi-1bc096.netlify.app/docs/content/what-are-the-differences-between-temporal-and-a-traditional-system/

Adds topic and one image.

## Notes to reviewers

@dtornow, given our recent discussions, please check my word choice throughout.


docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,65 @@
...
+
+export const headingText = 'What are the differences between Temporal and a traditional system?'
+
+<DetermineHeader
+hLevel={props.heading}
+hText={headingText}
+/>
+
+The Temporal platform exists to facilitate Workflow Executions.
+Compare a function execution in a traditional system to a Workflow Execution in Temporal.
```
flossypurse, line 19:
```suggestion
This page compares a function execution in a traditional system to a Workflow Execution in Temporal.
```

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,65 @@
...
+
+![Diagram that shows traditional function execution and Temporal Workflow Execution](/img/docs/function-execution-vs-reentrant-process.png)
```
flossypurse, line 21: Let's get @dtornow to get us an svg here -
And then we can update this to use the new CenteredImage component when it lands.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,65 @@
...
+## Failure
+
+In a traditional system, a service function execution (λ) is both volatile and short-lived.
+
+- If a function execution fails, it's not resumable because all execution state is lost. The longer a function execution awaits, the higher the change of failure.
+- A traditional function execution typically has a limited lifespan, often measured in minutes.
+
+In Temporal, a Workflow Execution (π) is resumable.
+
+- A Workflow Execution is fully resumable after a failure of any kind.
```
flossypurse, line 42: @dtornow what are the "exact" types of failures that are resumable from?
I feel like I would be hesitant to say "any kind" here.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,65 @@
...
+- Temporal imposes no deadline on a Workflow Execution.
```
flossypurse, line 43:
```suggestion
- Temporal imposes no deadlines on Workflow Executions.
A single Temporal Workflow Execution may live for years.
```

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,65 @@
...
+Your application (or a supporting component) must monitor the service's response to initiate a retry of the service execution.
+A retry starts from its _initial_ state.
+
+In Temporal, computation resumes from its _latest_ state. All progress is retained.
+
+## Communication
+
+In a traditional system, you can't "communicate" with a function execution.
+
+In Temporal, you can use Signals to "communicate" with a Workflow Execution.
```
flossypurse, line 57:
```suggestion
In Temporal, it is possible to send data into and get data out of a Workflow Execution.
```

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,65 @@
...
+You can also query the state of a Workflow Execution.
```
flossypurse, line 58: Maybe drop this in favor of the line above.

flossypurse (commented):
Not too shabby.
This is a great start. I think the sections make sense.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,65 @@
...
+The Temporal platform exists to facilitate Workflow Executions.
+Compare a function execution in a traditional system to a Workflow Execution in Temporal.
+
+![Diagram that shows traditional function execution and Temporal Workflow Execution](/img/docs/function-execution-vs-reentrant-process.png)
+
+:::note Legend
+
+λ = traditional function execution
+π = Temporal Workflow Execution
+𝝈 = state
```
flossypurse, line 27: In the deploy preview, these are not on newlines. They are all on the same line -
Either separate by one more newline or use bullets I think.

djmagee: Oops. Added two spaces at end of first and second lines, which in Markdown translates to a newline without starting a new para.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,64 @@
...
+---
+
+import DetermineHeader from '../components/DetermineHeader.js'
+
+export const headingText = 'What are the differences between Temporal and a traditional system?'
+
+<DetermineHeader
+hLevel={props.heading}
+hText={headingText}
+/>
```
flossypurse, line 16: OK, so it turns out we don't need this at all :)

djmagee: I'll yank it in my next commit.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,64 @@
...
+
+λ = traditional function execution  
+π = Temporal Workflow Execution  
+𝝈 = state
+
+:::
+
+Although the two systems seem similar at first glance, they differ in several significant ways.
+
+## Failure
```
flossypurse, line 33: One thing I realized is that we will be saving ourselves some trouble if we try to avoid any headings in the content pages.
Maybe make these sections just **bold text** ?
The page is short enough, I don't see a need for someone to need to link to one of these directly.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
```diff
@@ -0,0 +1,64 @@
...
+Your application (or a supporting component) must monitor the service's response to initiate a retry of the service execution.
+A retry starts from its _initial_ state.
+
+In Temporal, computation resumes from its _latest_ state. All progress is retained.
+
+## Communication
+
+In a traditional system, you can't "communicate" with a function execution.
+
+In Temporal, you can use Signals to send data into and get data out of a Workflow Execution.
```
flossypurse, line 57:
```suggestion
In a traditional system, passing data to and retrieving the state of a function execution is not really possible.

In Temporal, Signals and Queries enable data to be sent into or extracted from an open Workflow Execution.
```

Hmm... maybe something like this? "Communicate" feels off.

flossypurse:
Also, what do you think about switching "in" to "with?
> In a traditional system...
vs
> With a traditional system...

> In Temporal...
vs
> With Temporal....

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
djmagee: For now, let's go with bold. However, we need to circle back on this issue because headings are important for both accessibility and SEO.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
djmagee: Good call. I'd still like to avoid passive voice, but it's sometimes hard to avoid in technical docs.

djmagee: Re: "in" vs. "with". I could make a case for "in" because a service function execution occurs _in_ a traditional system (and a Workflow Function Execution _in_ Temporal). If, though, we're effectively comparing the high-level concept of a function execution in both systems, using "with" makes sense.

flossypurse: "in" vs "with" - At that moment when I posted "with" sounded more appropriate...
But I actually can go either way. So, don't feel obligated to keep that change.

docs/content/what-are-the-differences-between-temporal-and-a-traditional-system.md
flossypurse: svg added

sw-yx (approved):

PR #552, opened 2021-07-29
go getting started dev guide using the reusable content

flossypurse:
This PR showcases an example of what a "Getting Started Developers Guide" in Go might look like using the reusable content.

Preview: https://deploy-preview-552--mystifying-fermi-1bc096.netlify.app/docs/go/getting-started/

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -95,6 +95,7 @@ module.exports = {
       collapsed: false,
       items: [
         "go/introduction",
+        "go/getting-started",
         {
           type: "link",
           label: "Reference",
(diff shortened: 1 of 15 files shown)


sw-yx:
looking good 🔥

sw-yx (approved):

sidebars.js
```diff
@@ -95,6 +95,7 @@ module.exports = {
       collapsed: false,
       items: [
         "go/introduction",
+        "go/getting-started",
```
sw-yx, line 98: should this live under Tutorials instead?

sw-yx (approved):

sidebars.js
flossypurse: I suppose for now it could live under either - I am envisioning this whole category will shift in its presentation to something like this:
Go -> Go developer guide
- Getting Started
- Scheduling
- Cancelling
- Signals & Queries
- etc...

And then maybe have a whole different category for Go developer tutorials. It will become clearer I think, as we move things to be reusable pages like this.

PR #553, opened 2021-07-29
experimental index sidebar

flossypurse:


djmagee:
I'm not sure how valuable a straight alphabetic index would be, especially as the sidebar, but I could definitely see value in a sidebar that grouped topics in a high-value way, such as feature (Worker, Workflow, Task). Ideally, the sidebar should provide some kind of context so that the reader can say "I am _here_"—kinda like a fingerboard sign that points toward various nearby destinations.

sw-yx:
+1 on grouping things

PR #558, opened 2021-08-01
Copyedits

lorensr:
In queries.md, there was a mix of "query callback" and "query handler". I standardized on latter because it seemed more functionally descriptive.


flossypurse (approved):

flossypurse:
Thank you for the contribution!

PR #560, opened 2021-08-01
Organize Node.js getting started page

bergundy:
Changed this because we now support the brew node installation.
Reordered things a bit, I think it's better this way.

### DIFF
--- a/docs/node/getting-started.md
+++ b/docs/node/getting-started.md
@@ -4,20 +4,25 @@ You can run "Hello Temporal" locally in under 5 minutes, assuming you have all p
 
 ## Step 0: Prerequisites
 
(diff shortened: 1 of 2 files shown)


flossypurse (approved):

flossypurse:
> Changed this because we now support the brew node installation.
> Reordered things a bit, I think it's better this way.

In the very near future I want to push for node docs to be converted into our re-usable content structure -

bergundy:
> > Changed this because we now support the brew node installation.
> > Reordered things a bit, I think it's better this way.
> 
> In the very near future I want to push for node docs to be converted into our re-usable content structure -

Me too

PR #576, opened 2021-08-15
Document Node.js interceptors

bergundy:


docs/node/interceptors.md
```diff
@@ -0,0 +1,113 @@
...
+  WorkflowInboundCallsInterceptor,
+  WorkflowInput,
+} from "@temporalio/workflows";
+
+/**
+ * WARNING: This demo is meant as a simple auth example.
+ * Do not use this for actual authorization logic.
+ * Auth headers should be encrypted and credentials
+ * stored outside of the codebase
+ */
```
sw-yx, line 68: this is great

docs/node/interceptors.md
```diff
@@ -0,0 +1,113 @@
...
+
+```ts
+export const interceptors = {
+  outbound: [new LogActivitiesInterceptor()],
+  inbound: [],
+};
+```
+
+At the time of construction, the Workflow Context is already initialized for the current Workflow.
+Use [`Context.info`](https://nodejs.temporal.io/api/interfaces/workflow.workflowinfo) to add Workflow specific information in the interceptor.
```
sw-yx, line 113: still not very clear how to use, probably want to have a complete example in a single file

sw-yx (approved):

PR #582, opened 2021-08-19
Update Signals doc to clarify callback func

si-stripe:
## What does this PR do?
Made a small update to the Signals doc to clarify what the callback function is for, and more specifically what the `more` bool means. Found it difficult to figure out, but I'm basing it off the comment here https://github.com/temporalio/sdk-go/blob/7abca668facadd88ea0392aabbd7314a30c8db52/internal/internal_workflow.go#L676

## Notes to reviewers
Please correct me if I'm wrong, I spent some time on the community forum and StackOverflow and this is what I got


flossypurse (approved):

flossypurse:
Thank you for the update!

PR #606, opened 2021-08-30
Add an example of a valid JWT token payload and some clarifications into structure of JWT Permissions Claim

Spikhalskiy:
## What does this PR do?
Currently, the description of JWT Permissions Token claim is not clear enough to understand how a valid claim should be structured.

- Added an example of a valid Temporal JWT token
- Added clarifications to the text description of the claim Structure
- Added `NewDefaultJWTClaimMapper` method signature to the doc, because names of the parameters are not enough (for example, `provider` in the previous version of the code block can be anything, it's unclear that it's `TokenKeyProvider`)
- Separated default JWT ClaimMapper config section
- Changed "bearer" to "Bearer" in a token structure as being more idiomatic

### DIFF
--- a/docs/server/security.md
+++ b/docs/server/security.md
@@ -131,13 +131,14 @@ role := authorization.RoleReader | authorization.RoleWriter
 
 Temporal offers a default JSON Web Token `ClaimMapper` that extracts claims from JWT access tokens and translates them into Temporal `Claims`.
 The default JWT `ClaimMapper` needs a public key to perform validation of tokens' digital signatures and expects [JWT tokens](https://tools.ietf.org/html/rfc7519) to be in the certain format [described below](#format-of-json-web-tokens).
-You can use the default as an example to build your own for translating a caller's authorization information from other formats and conventions into Temporal `Claims`.
+
+You can use the default JWT `ClaimMapper` as an example to build your own `ClaimMapper` for translating a caller's authorization information from other formats and conventions into Temporal `Claims`.
+
+To get an instance of the default JWT `ClaimMapper`, call `NewDefaultJWTClaimMapper` and provide it with an instance of a [`TokenKeyProvider`](#tokenkeyprovider), a pointer to a [`config.Authorization`](#configauthorization) config, and a logger.
 
 ```go
-claimMapper := authorization.NewDefaultJWTClaimMapper(provider, cfg, logger)
+claimMapper := authorization.NewDefaultJWTClaimMapper(tokenKeyProvider, authCfg, logger)
 ```
-
-Note that the default JWT `ClaimMapper` sets "permissions" as the `permissionsClaimName` if the `config.Config.Global.Authorization.PermissionsClaimName` property is not set as an override for the JWT tokens.
 
 #### `TokenKeyProvider`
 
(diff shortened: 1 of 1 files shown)


Spikhalskiy:
I also think that it's a really good idea to make a separate sub-page "JWT" under "Security".
- It will make it more clear and loud that Temporal supports JWT out-of-the-box
- It will bring more structure, because "default JWT ClaimMapper" is a significant portion of the content and it's a specific implementation of whatever is described above.

flossypurse (approved):

docs/server/security.md
```diff
@@ -134,11 +134,10 @@ The default JWT `ClaimMapper` needs a public key to perform validation of tokens
 You can use the default as an example to build your own for translating a caller's authorization information from other formats and conventions into Temporal `Claims`.
 
 ```go
+// authorization.NewDefaultJWTClaimMapper(provider TokenKeyProvider, cfg *config.Authorization, logger log.Logger)
```
sergeybykov, line 137: Why this commented out line?

sergeybykov (approved):
Just one commented out line of code that I think shouldn't be there.

docs/server/security.md
Spikhalskiy: Actually, to make this block a valid Go code mostly.
I actually don't know why do we have snippets like this in our docs:
```
claimMapper := authorization.NewDefaultJWTClaimMapper(provider, cfg, logger)
```
What is `provider` here? What is the type of this specific `cfg`?
It would be much more useful to have a method signature instead.

sergeybykov: Never mind - I missed that this is just signature of the method for reference.

flossypurse: I would add something like this to the line before the code snippet that shows the API call:

> To get an instance of the default ClaimMapper, call [`NewDefaultJWTClaimMapper`](https://github.com/temporalio/temporal/blob/master/common/authorization/default_jwt_claim_mapper.go) and provide it with an instance of a [`TokenKeyProvider`](https://github.com/temporalio/temporal/blob/master/common/authorization/token_key_provider.go#L34) (described below), a pointer to an [`Authorization`](https://github.com/temporalio/temporal/blob/master/common/config/config.go#L479) config, and a [logger](https://github.com/temporalio/temporal/blob/master/common/log/interface.go#L42).

flossypurse:
> I also think that it's a really good idea to make a separate sub-page "JWT" under "Security".
> 
> * It will make it more clear and loud that Temporal supports JWT out-of-the-box
> * It will bring more structure, because "default JWT ClaimMapper" is a significant portion of the content and it's a specific implementation of whatever is described above.

First step in my mind is to identify all the questions that we are trying to answer - that will scope out everything naturally.
Once the content is organized in that fashion, then building the narratives becomes trivial in comparison I should think.

Spikhalskiy:
> > I also think that it's a really good idea to make a separate sub-page "JWT" under "Security".
> > 
> > * It will make it more clear and loud that Temporal supports JWT out-of-the-box
> > * It will bring more structure, because "default JWT ClaimMapper" is a significant portion of the content and it's a specific implementation of whatever is described above.
> 
> First step in my mind is to identify all the questions that we are trying to answer - that will scope out everything naturally.
> Once the content is organized in that fashion, then building the narratives becomes trivial in comparison I should think.

A lot of people will be coming specifically looking for JWT, that's why I think we should group JWT specific questions and manual into a separate page.

PR #610, opened 2021-08-30
document changes to TypeScript SDK Activities

sw-yx:
document changes to TypeScript SDK Activities


sw-yx:
i tackled this in #648

PR #611, opened 2021-08-31
Updates per review & feedback

flossypurse:
## What does this PR do?

Updates a few sections per @mfateev feedback.

### DIFF
--- a/docs/content/how-to-develop-a-workflow-definition-in-go.md
+++ b/docs/content/how-to-develop-a-workflow-definition-in-go.md
@@ -94,7 +94,7 @@ In Go specifically, Workflow Definition code can not directly do the following:
 
 - Iterate over maps using `range`, because with `range` the order of the map's iteration is randomized. (Use a Side Effect or an Activity instead.)
 - Use the native `go` statement, `select` statement, or `chan` type. (Use the [SDK Go API](#), [SDK Select API](#), or [SDK Channel API](#) instead.)
-- Call any external APIs, conduct any file I/O operations, talk to other services, etc (Use an Activity for these).
+- Call an external API, conduct a file I/O operation, talk to another service, etc. (Use an Activity for these.)
 
 <!--
 <RelatedReadList
--- a/docs/content/how-to-develop-a-workflow-definition-in-go.md
+++ b/docs/content/how-to-develop-a-workflow-definition-in-go.md
@@ -93,7 +93,7 @@ readlist={[
 In Go specifically, Workflow Definition code can not directly do the following:
 
 - Iterate over maps using `range`, because with `range` the order of the map's iteration is randomized. (Use a Side Effect or an Activity instead.)
-- Use the native `go` statement, `select` statement, or `chan` type (Use the [SDK Go API](#), [SDK Select API](#), and [SDK Channel API](#))
+- Use the native `go` statement, `select` statement, or `chan` type. (Use the [SDK Go API](#), [SDK Select API](#), or [SDK Channel API](#) instead.)
 - Call any external APIs, conduct any file I/O operations, talk to other services, etc (Use an Activity for these).
 
 <!--
--- a/docs/content/how-to-develop-a-workflow-definition-in-go.md
+++ b/docs/content/how-to-develop-a-workflow-definition-in-go.md
@@ -92,7 +92,7 @@ readlist={[
 
 In Go specifically, Workflow Definition code can not directly do the following:
 
-- Iterate over maps using `range`, because with `range` the order of the map's iteration is randomized (Use a Side Effect or an Activity.
+- Iterate over maps using `range`, because with `range` the order of the map's iteration is randomized. (Use a Side Effect or an Activity instead.)
 - Use the native `go` statement, `select` statement, or `chan` type (Use the [SDK Go API](#), [SDK Select API](#), and [SDK Channel API](#))
 - Call any external APIs, conduct any file I/O operations, talk to other services, etc (Use an Activity for these).
 
--- a/docs/content/how-to-develop-a-workflow-definition-in-go.md
+++ b/docs/content/how-to-develop-a-workflow-definition-in-go.md
@@ -81,7 +81,6 @@ However, it is not possible to receive both a custom value and an error in the c
 The caller will receive either one or the other.
 Returning a non-nil `error` from a Workflow indicates that an error was encountered during its execution and the Workflow Execution should be [Terminated](#) and any custom return values will be ignored by the system.
 
-
 <!--
 <RelatedReadList
 readlist={[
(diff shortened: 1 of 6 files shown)


docs/content/how-to-develop-a-workflow-definition-in-go.md
```diff
@@ -77,6 +77,8 @@ func YourWorkflowDefinition(ctx workflow.Context, param YourWorkflowParam) (Your
 ```
 
 Returning a non-nil `error` from a Workflow indicates that an error was encountered during its execution and the Workflow Execution should be [Terminated](#).
+This will also cause any custom return value to be ignored.
+It is not possible for both a non nil error and a custom return value from a Workflow Execution to be received as a return values, from the Workflow Execution, as is possible in pure Go.
```
djmagee, line 81:
```suggestion
It is not possible for both a non-nil error and a custom return value from a Workflow Execution to be received as return values from the Workflow Execution, as is possible in pure Go.
```
Or rewrite to avoid "It is":
A Workflow Execution cannot return both a non-nil error and a custom return value, something that is possible in pure Go.

docs/content/how-to-develop-a-workflow-definition-in-go.md
```diff
@@ -92,6 +94,7 @@ In Go specifically, Workflow Definition code can not directly do the following:
 
 - Iterate over maps using `range`, because with `range` the order of the map's iteration is randomized (Use a Side Effect or an Activity.
```
djmagee, line 95:
```suggestion
- Iterate over maps using `range`, because with `range` the order of the map's iteration is randomized. (Use a Side Effect or an Activity instead.)
```

docs/content/how-to-develop-a-workflow-definition-in-go.md
```diff
@@ -92,6 +94,7 @@ In Go specifically, Workflow Definition code can not directly do the following:
...
 - Use the native `go` statement, `select` statement, or `chan` type (Use the [SDK Go API](#), [SDK Select API](#), and [SDK Channel API](#))
```
djmagee, line 96:
```suggestion
- Use the native `go` statement, `select` statement, or `chan` type. (Use the [SDK Go API](#), [SDK Select API](#), or [SDK Channel API](#) instead.)
```

docs/content/how-to-develop-a-workflow-definition-in-go.md
```diff
@@ -92,6 +94,7 @@ In Go specifically, Workflow Definition code can not directly do the following:
...
+- Call any external APIs, conduct any file I/O operations, talk to other services, etc (Use an Activity for these).
```
djmagee, line 97:
```suggestion
- Call an external API, conduct a file I/O operation, talk to another service, etc. (Use an Activity for these.)
```

docs/content/how-to-develop-an-activity-definition-in-go.md
```diff
@@ -28,10 +28,15 @@ type YourActivityStruct struct {
...
+  // ...
+}
 
+func(a *YourActivityStruct) YourActivityDefinitionTwo(ctx workflow.Context) error {
+  // ...
 }
 ```
 
+An "Activity struct" can have more than one method, with each method acting as a separate Activity Type.
 Activities written as struct methods can make use of shared struct variables.
```
djmagee, line 40:
```suggestion
Activities written as struct methods can use shared struct variables.
```

docs/content/how-to-develop-an-activity-definition-in-go.md
```diff
@@ -28,10 +28,15 @@ type YourActivityStruct struct {
...
 The rest of this guide will show Activities written as struct methods.
```
djmagee, line 41:
```suggestion
The rest of this guide shows Activities written as struct methods.
```

docs/content/how-to-get-the-result-of-a-workflow-execution-in-go.md
```diff
@@ -26,40 +26,36 @@ The instance of `WorkflowRun` has the following three methods:
...
   if err != nil {
       // ...
   }
   // ...
 }
 ```
 
-However, the result of a Workflow Execution can be obtained from a completely different process, all that is needed is the [Workflow Id](#) and [Run Id](#).
-The result of the Workflow Execution is available for as long as it's Execution History remains in the system (See [How long do Workflow Execution Histories persist](#)).
+However, the result of a Workflow Execution can be obtained from a completely different process, all that is needed is the [Workflow Id](#) (a [Run Id](#) is optional if there is more than one closed Workflow Execution with the same Workflow Id).
```
djmagee, line 42:
```suggestion
However, the result of a Workflow Execution can be obtained from a completely different process.
All that is needed is the [Workflow Id](#).
(A [Run Id](#) is optional if more than one closed Workflow Execution has the same Workflow Id.)
```

docs/content/how-to-spawn-a-workflow-execution-in-go.md
```diff
@@ -10,12 +10,8 @@ import RelatedReadList from '../components/RelatedReadList.js'
...
 To spawn a Workflow Execution, use the `ExecuteWorkflow()` method on the Go SDK [`Client`](https://pkg.go.dev/go.temporal.io/sdk@v1.8.0/client#Client), which is available via [`NewClient()`](https://pkg.go.dev/go.temporal.io/sdk@v1.8.0/client#NewClient) in the [`go.temporal.io/sdk/client`](https://pkg.go.dev/go.temporal.io/sdk@v1.8.0/client) package.
 
-<RelatedReadList
-readlist={[
-["What is a Workflow Execution?","/docs/content/what-is-a-workflow-execution","explanation"],
-["How to spawn a Child Workflow Execution","/docs/content/how-to-spawn-a-child-workflow-execution-in-go","developer guide"],
-]}
-/>
+The Go SDK Client should never be used inside a Workflow Definition.
+To spawn a Workflow Execution from within another Workflow, use the [`ExecuteChildWorkflow` API](/docs/content/how-to-spawn-a-child-workflow-execution-in-go).
```
djmagee, line 14: To make link text match targets:
```suggestion
To spawn a Workflow Execution from within another Workflow, use the [`ExecuteChildWorkflow`](https://pkg.go.dev/go.temporal.io/sdk/workflow#ExecuteChildWorkflow) API.
For more information, see [How to spawn a Child Workflow Execution in Go](/docs/content/how-to-spawn-a-child-workflow-execution-in-go).
```

docs/content/how-to-spawn-a-workflow-execution-in-go.md
```diff
@@ -65,6 +62,9 @@ The only field, of the `StartWorkflowOptions` instance, that requires a value is
 A Task Queue name is also provided to the Worker that is registered to execute that particular Workflow Type.
 The Task Queue name must be the same for both.
 
+We recommend supplying your own custom Workflow Id that can be used to get the result of the Workflow Execution asynchronously at another point in time.
+The Workflow Id is meant to be supplied by the user and correspond to some business-level identifier.
```
djmagee, line 66: Or perhaps "A custom Workflow Id is intended to correspond to a business-level identifier."

djmagee (approved):

docs/content/how-to-develop-a-workflow-definition-in-go.md
flossypurse: Ya, the tricky thing here is that it CAN return both from the Definition... its just that only one or the other will be available to the caller who is looking for the result.

PR #612, opened 2021-08-31
Wrong struct used: MySignal in place of MyStruct

Parth-Vader:
Came across this while going through the documentation, MySignal in place of MyStruct should be used in the example.

## What does this PR do?
Make the change


flossypurse (approved):
Thanks! Good catch.

PR #623, opened 2021-09-06
Fix function signature

agbaraka:
A fix on a function signature in Large Event History section


flossypurse (approved):

flossypurse:
Thank you @agbaraka !

PR #627, opened 2021-09-09
Reflect changes from new Node.js Workflow API

bergundy:
Updates the docs with recent Node.js API changes: https://github.com/temporalio/sdk-node/pull/225

### DIFF
--- a/docs/node/failures.md
+++ b/docs/node/failures.md
@@ -18,7 +18,7 @@ If, for example, a Node.js Workflow starts a Java Child Workflow which calls an 
 <!--TODO: use snipsync-->
 
 ```ts
-import { childWorkflow } from '@temporalio/workflow';
+import { createChildWorkflowHandle } from '@temporalio/workflow';
 import {
   ActivityFailure,
   ApplicationFailure,
@@ -26,13 +26,15 @@ import {
 } from '@temporalio/common';
 
 // Define the TypeScript version of the Java Workflow interface
-// to get a type safe child WorkflowStub
+// to get a type safe child WorkflowHandle
 export type JavaWorkflow = () => {
   execute(): Promise<void>;
 };
 
 async function execute() {
-  const child = childWorkflow<JavaWorkflow>('RunAnActivityWorkflow');
+  const child = createChildWorkflowHandle<JavaWorkflow>(
+    'RunAnActivityWorkflow'
+  );
   try {
     await child.execute();
   } catch (err) {
@@ -65,14 +67,13 @@ As explained above, cancellation might not be the immediate cause of failure—i
 ```ts
 import {
   CancellationScope,
-  configureActivities,
+  createActivityHandle,
   isCancellation,
 } from '@temporalio/workflow';
 import * as activities from '../activities';
 
 export function myWorkflow(urls: string[], timeoutMs: number) {
-  const { httpGetJSON } = configureActivities<typeof activities>({
-    type: 'remote',
+  const { httpGetJSON } = createActivityHandle<typeof activities>({
     scheduleToCloseTimeout: timeoutMs,
   });
 
@@ -97,7 +98,7 @@ export function myWorkflow(urls: string[], timeoutMs: number) {
 
 </details>
 
-Outside of Workflow code, failure classes are attached to the `cause` of [`WorkflowExecutionFailedError`](https://nodejs.temporal.io/api/classes/client.workflowexecutionfailederror), which is thrown when executing a Workflow with a [`WorkflowClient`](https://nodejs.temporal.io/api/classes/client.workflowclient/) or [`WorkflowStub`](https://nodejs.temporal.io/api/interfaces/client.workflowstub/).
+Outside of Workflow code, failure classes are attached to the `cause` of [`WorkflowExecutionFailedError`](https://nodejs.temporal.io/api/classes/client.workflowexecutionfailederror), which is thrown when executing a Workflow with a [`WorkflowClient`](https://nodejs.temporal.io/api/classes/client.workflowclient/) or [`WorkflowHandle`](https://nodejs.temporal.io/api/interfaces/client.workflowhandle/).
 
 ## Failures and retries
 
--- a/docs/node/failures.md
+++ b/docs/node/failures.md
@@ -18,7 +18,7 @@ If, for example, a Node.js Workflow starts a Java Child Workflow which calls an 
 <!--TODO: use snipsync-->
 
 ```ts
-import { Context, Workflow } from '@temporalio/workflow';
+import { childWorkflow } from '@temporalio/workflow';
 import {
   ActivityFailure,
   ApplicationFailure,
@@ -27,12 +27,12 @@ import {
 
 // Define the TypeScript version of the Java Workflow interface
 // to get a type safe child WorkflowStub
-interface JavaWorkflow extends Workflow {
-  main(): void;
-}
+export type JavaWorkflow = () => {
+  execute(): Promise<void>;
+};
 
-async function main() {
-  const child = Context.child<JavaWorkflow>('RunAnActivityWorkflow');
+async function execute() {
+  const child = childWorkflow<JavaWorkflow>('RunAnActivityWorkflow');
   try {
     await child.execute();
   } catch (err) {
@@ -50,7 +50,7 @@ async function main() {
   }
 }
 
-export const workflow = { main };
+export const myWorkflow = () => ({ execute });
 ```
 
 </details>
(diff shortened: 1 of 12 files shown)


docs/node/failures.md
```diff
@@ -64,28 +64,34 @@ As explained above, cancellation might not be the immediate cause of failure—i
...
-    );
-  } catch (err) {
-    if (isCancellation(err)) {
-      console.log('Deadline exceeded while waiting for activities to complete');
-    }
-    throw err;
-  }
+export function myWorkflow(urls: string[], timeoutMs: number) {
+  const { httpGetJSON } = configureActivities<typeof activities>({
+    type: 'remote',
```
sw-yx, line 75: what is `type`?

docs/node/signals.md
```diff
@@ -29,9 +29,9 @@ You invoke a Signal with `workflow.signal.signalName(...args)`. In the above cas
 
 ```ts
 const client = new WorkflowClient();
-const workflow = client.stub<Blocked>('blocked', { taskQueue: 'test' });
+const workflow = client.stub(unblockWithSignal, { taskQueue: 'test' });
```
sw-yx, line 32: if task queue is basically always required in practice, should we do this instead? more ergonomic?

```suggestion
const workflow = client.stub({ 
	type: unblockWithSignal, // required
	taskQueue: 'test'        // practically required
});
```

also wondering if it read easier to have our users say `new Stub(...)` instead of `client.stub(...)`? pros and cons?

docs/node/task-queues.md
```diff
@@ -37,20 +37,11 @@ const worker = await Worker.create({
...
 
 ```ts
-import { Context } from '@temporalio/workflow';
-import * as activities from '../activities';
-
-const { greet } = Context.configureActivities<typeof activities>({
+const { greet } = configureActivities<typeof activities>({
   taskQueue: 'my-other-task-queue',
   startToCloseTimeout: '1s',
 });
```
sw-yx, line 46: this api bothers me a bit. `greet` appears out of nowhere, guaranteed only by `typeof activities`.

what do we think about something that more intentionally mirrors the workflow stubbing:

```tsx
const greet = new Activity({
  type: activities.greet,
  taskQueue: 'my-other-task-queue',
  startToCloseTimeout: '1s'
});
```

docs/node/versioning.md
```diff
@@ -20,10 +20,12 @@ import { ResponsivePlayer } from '../../src/components'
...
-async function main() {
-  await activityA();
-  await activityThatMustRunAfterA();
-}
+export const myWorkflow = () => ({
+  async execute() {
+    await activityA();
+    await activityThatMustRunAfterA();
+  },
+});
```
sw-yx, line 28: execute() or main()?

docs/node/signals.md
```diff
@@ -41,28 +41,35 @@ Signal handlers can be either synchronous or asynchronous, in this example, our
...
+          unblock();
+        }
+      },
+    },
+    queries: {
+      isBlocked(): boolean {
+        return blocked;
+      },
+    },
+    async main(): Promise<void> {
```
sw-yx, line 64: shoudl this be execute()?

docs/node/workers.md
```diff
@@ -6,53 +6,26 @@ sidebar_label: Workers
...
-  console.log(err);
-  process.exit(1);
-});
-
-async function main() {
-  const worker = await Worker.create({
-    workDir: __dirname,
-    taskQueue: 'test',
-  });
-  await worker.run();
```
sw-yx, line 43: maybe we should harmonize the creation of Worker, Activity, Workflows so that users dont have to memorize different verbs.

sw-yx (approved):
left a couple comments on api tweaks and one probable mistake

docs/node/failures.md
lorensr: https://nodejs.temporal.io/api/namespaces/worker#activityoptions

I imagine remote is the default?

docs/node/signals.md
```diff
@@ -41,28 +41,35 @@ Signal handlers can be either synchronous or asynchronous, in this example, our
...
-      unblock();
-    }
-  },
+export const unblockWithSignal: Blocked = () => {
+  let blocked = true;
+  let unblock: () => void;
+
+  return {
+    signals: {
+      // Unblock main by resolving the awaited Promise
```
lorensr, line 52: Unblock `execute()` by..

docs/node/signals.md
lorensr: I think I like the single-arg object.

> also wondering if it read easier to have our users say new Stub(...) instead of client.stub(...)? pros and cons?

You need the client info, right? So `new client.Stub({ })` or `new Stub({ client, ... })`?

bergundy: We need the client, so the constructor would have to take a `client` argument as @lorensr pointed out.
I don't love the name `stub` either but I'm not sure what's the best alternative.
We already have a `WorkflowClient`, `WorkflowHandle` maybe? Just `Workflow` (we already have a `Workflow` interface so maybe not the best choice)?

I also prefer the method over a constructor, it's a bit more convenient IMO, also 1 less import.
I don't have a strong preference for single vs 2 args.

Suggested alternatives:

```ts
client.workflow({ type: unblockWithSignal, taskQueue: 'test' })
client.stub({ type: unblockWithSignal, taskQueue: 'test' })
client.handle({ type: unblockWithSignal, taskQueue: 'test' })
client.workflow(unblockWithSignal, { taskQueue: 'test' })

```

docs/node/signals.md
bergundy: I should have grepped for main, missed it. Thanks!

docs/node/signals.md
bergundy: picard + 🤦

docs/node/task-queues.md
bergundy: The won't work because `activities.greet` is just the type and cannot be used as an argument.
This is the best way I find to get a typed and named reference to an activity without actually importing the activities into the Workflow isolate (which is not supported).

docs/node/workers.md
bergundy: Activities are a special case as mentioned above.
I'm open to discussing the creation of workflows / child workflows / external workflows.

I would keep `Worker.create` though, we need an async factory for instantiation.

bergundy (commented):
Thanks for the suggestions.

Take a look at my replies.
I'd really love to close all of this ASAP.

docs/node/failures.md
lorensr: I like an optional boolean w/ default false

docs/node/signals.md
lorensr: On the React theme, we could call them Refs 😄 

Similar to my Activities feedback, I like methods with verbs, so I can better understand what's happening by reading. These seem similarly convenient and understandable to me:

```js
client.createWorkflow({ type: unblockWithSignal, taskQueue: 'test' })
client.getWorkflow({ type: unblockWithSignal, taskQueue: 'test' })
client.getWorkflowRef({ type: unblockWithSignal, taskQueue: 'test' })

new client.Workflow({ type: unblockWithSignal, taskQueue: 'test' })
new client.WorkflowRef({ type: unblockWithSignal, taskQueue: 'test' })
```

bergundy: That name has actually cross my mind, definitely an option.

I changed the node PR to use the same methods as Java and PHP for lack of better ideas.

`client.newWorkflowStub` `newChildWorkflowStub` `newExternalWorkflowStub`

docs/node/failures.md
sw-yx: `local: false` is fine until we need to add a third type of activity, i'm not sure if we ever will so i guess its ok

i dislike `type: 'remote'` because "remote activity" is not defined anywhere else in other SDKs.

docs/node/task-queues.md
sw-yx: this is a good answer - we should actually explain how activities are injected into the WF isolate, as I'm sure this will come up again. giving users a good mental model will save us 100 questions down the line

docs/node/signals.md
sw-yx:
```
client.handle
client.createWorkflowHandle
client.createHandle
client.newWorkflowHandle
client.newHandle
```

we like explicit because

```
client.createActivity
client.createWorkflow
client.createChildWorkflow
client.createExternalWorkflow
```

bergundy:
I've made all the changes discussed yesterday apart from adding defaults to `WorkflowClient` (https://github.com/temporalio/sdk-node/issues/240).
Only thing left is the `ActivityOptions.type` attribute which accepts `remote` and in the future `local`.

@sw-yx had a point about not using booleans.
Let's remove it for now and make it the default.
We'll revisit this when we add local activities.

bergundy:
> Let's remove it for now and make it the default.

Done

PR #634, opened 2021-09-13
fix What is a Cluster doc inaccuracy

sw-yx:
## What does this PR do?

fixes a longstanding inaccuracy in our cluster docs

## Notes to reviewers

todo: update the chart, as it is misleading


sw-yx:
closed in favor of #636

PR #641, opened 2021-09-17
fix namespaces docs

sw-yx:
## What does this PR do?

[our current namespaces doc](https://docs.temporal.io/docs/server/namespaces) is actually about "global namespaces", a feature most users will never use. this PR fixes that by adding the prerequisite knowledge our customers (who only ever interact with us via non default namespaces) actually need.

## Notes to reviewers

reuses and refactors some of the content work we already did as part of jargon mesh drive


flossypurse (approved):
Awesome!

PR #648, opened 2021-09-21
better document Node Workers and Workflows

sw-yx:
## What does this PR do?

better document Node Workers

## Notes to reviewers

https://deploy-preview-648--mystifying-fermi-1bc096.netlify.app/docs/node/workers/
https://deploy-preview-648--mystifying-fermi-1bc096.netlify.app/docs/node/workflows/


docs/node/workers.md
```diff
@@ -7,25 +7,80 @@ sidebar_label: Workers
...
+Below is an example of starting a Worker that polls the Task Queue named `tutorial`.
+
+<!--SNIPSTART nodejs-hello-worker {"enable_source_link": false}-->
+<!--SNIPEND-->
+
+### Required Worker Options
+
+There are two required options to the `Worker.create()` function:
 
 1. The `workDir`. The Node SDK will automatically register:
```
bergundy, line 28: This is incorrect, `workDir` is not required, maybe move it below or above the table where we document `activities`, `workflowsPath` and `nodeModulesPath`?

bergundy (approved):
Very nice!

docs/node/activities.md
```diff
@@ -23,32 +24,81 @@ Note that we only import the type of our activities, the TypeScript compiler wil
...
+ms('1y')       // 31557600000
+ms('100')      // 100
+```
 
-1. Await on [`Context.current().cancelled`](https://nodejs.temporal.io/api/classes/activity.context#cancelled)
-1. Pass the context's abort signal at [`Context.current().cancellationSignal`](https://nodejs.temporal.io/api/classes/activity.context#cancelled) to a library that supports it like `fetch`
+To better understand Activity timeouts, refer to our blogpost on the [4 Types of Activity timeouts](https://docs.temporal.io/blog/activity-timeouts).
 
-[`heartbeat()`](https://nodejs.temporal.io/api/classes/activity.context/#heartbeat) in the Node.js SDK is a background operation and does not propagate errors to the caller, such as when the scheduling Workflow has already completed or the Activity has been closed by the server (due to timeout for instance). These errors are translated into cancellation and can be handled using the methods above.
+## How to register an Activity
```
flossypurse, line 58: Up to you at this stage, but this seems like a clearly scoped piece of "how to" content
Also, this does exist, perhaps you can make use of it:
https://docs.temporal.io/docs/content/how-to-develop-a-worker-program-in-node/

sw-yx: ah excellent! will merge, ty for the catch

docs/node/workers.md
sw-yx: ok got you. thanks

docs/node/activities.md
bergundy: @sw-yx @flossypurse note that the how to page is out-of-date now.

We've been rapidly changing the SDK's API in the last few weeks in prep for beta and I wasn't even aware this content was there, otherwise I'd have updated it.
It'd be nice if all of the ~~Node~~ TS SDK content was gathered in a single directory so it'd be easier to locate.

sw-yx: ya no worries about this page, i will update it

sw-yx:
i think main potentially controversial thing done here is that i combined signals and queries - please try to look at the docs with a fresh mind and a view to the page length from page to page and see if it makes sense. if it doesnt i can revert.

PR #654, opened 2021-09-28
Improve the advice for handling maps

robholland:
It's preferable to sort the keys for a map to maintain determinism rather than add the overhead of an activity or side-effect.


flossypurse (approved):
👍

PR #658, opened 2021-09-29
[Go docs] add Activity dependency sharing note

sw-yx:
## What does this PR do?

add a note on dependency sharing
pls approve/merge https://github.com/temporalio/samples-go/pull/139 alongside

## Notes to reviewers

requested by Maxim https://temporalio.slack.com/archives/CTDTU3J4T/p1632924716139800?thread_ts=1632923907.136400&cid=CTDTU3J4T

and something we had to deal with in Node https://docs.temporal.io/docs/node/activities/#sharing-dependencies-in-activity-functions probably should also do the same in Java/PHP

### DIFF
--- a/docs/content/how-to-develop-an-activity-definition-in-go.md
+++ b/docs/content/how-to-develop-an-activity-definition-in-go.md
@@ -126,8 +126,8 @@ readlist={[
 
 All native features of the Go programming language can be used within an Activity and there are no other limitations to Activity Definition logic:
 
-- **Performance**: Keep in mind that all parameters and return values are recorded in the execution history.
-  A large execution history can adversely impact the performance of your Workflows as the entire history is transferred to your workers with every event processed.
+- **Performance**: Keep in mind that all parameters and return values are recorded in the [Workflow Execution Event History](/docs/content/what-is-an-event-history).
+A large Workflow Execution Event History can adversely impact the performance of your Workflow Executions, as the entire Event History is transferred to Worker Processes with every [Workflow Task](/docs/content/what-is-a-workflow-task).
 - **Idiomatic usage**: You are free to use:
   - your own loggers and metrics controllers
   - the standard Go concurrency constructs
--- a/docs/content/how-to-develop-an-activity-definition-in-go.md
+++ b/docs/content/how-to-develop-an-activity-definition-in-go.md
@@ -9,15 +9,19 @@ tags:
 
 import RelatedReadList from '../components/RelatedReadList.js'
 
-In the Temporal Go SDK programming model, an Activity Definition is an exportable function or `stuct` method.
+In the Temporal Go SDK programming model, an Activity Definition is an exportable function or `struct` method.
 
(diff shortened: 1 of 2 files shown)


docs/go/activities.md
```diff
@@ -257,6 +257,17 @@ Activities to complete by calling the `Get()` methods of the future at a later t
 
 To implement more complex wait conditions on returned future objects, use `workflow.Selector`. Learn more on the [Go SDK Selectors](https://docs.temporal.io/docs/go/selectors) page.
 
+### Sharing Dependencies between Activities
```
flossypurse, line 260: This is already documented here: https://docs.temporal.io/docs/content/how-to-develop-an-activity-definition-in-go/

I would prefer to deprecate this page, rather than add to it.
Or, reuse the content from above.

asonawalla: 👋 I was [asking](https://temporalio.slack.com/archives/CTDTU3J4T/p1632923907136400) about this in slack this morning - the confusion (at least from my perspective) is that it's not clear from the wording linked above in what scenarios using a struct is advantageous.  I think something like:

> Activities written as struct methods can use shared struct variables _such as connection pools to external services initialized when worker processes are started_.

would do the trick.

sw-yx:
@flossypurse i've updated now with a bit of tweaking of the reused language

docs/content/how-to-develop-an-activity-definition-in-go.md
```diff
@@ -113,10 +122,16 @@ readlist={[
...
 -->
 
-**Activity logic requirements in Go**
+**Other notes for developing Activities**
+
+All native features of the Go programming language can be used within an Activity and there are no other limitations to Activity Definition logic:
 
-There are no other limitations to Activity Definition logic.
-All native features of the Go programming language can be used within an Activity, and it is idiomatic to use an Activity to make calls to other services across a network.
+- **Performance**: Keep in mind that all parameters and return values are recorded in the execution history.
```
flossypurse, line 129:
```suggestion
- **Performance**: Keep in mind that all parameters and return values are recorded in the [Workflow Execution Event History](/docs/content/what-is-an-event-history).
```

docs/content/how-to-develop-an-activity-definition-in-go.md
```diff
@@ -113,10 +122,16 @@ readlist={[
...
+  A large execution history can adversely impact the performance of your Workflows as the entire history is transferred to your workers with every event processed.
```
flossypurse, line 130:
```suggestion
A large Workflow Execution Event History can adversely impact the performance of your Workflow Executions, as the entire Event History is transferred to Worker Processes with every [Workflow Task](/docs/content/what-is-a-workflow-task).
```

flossypurse (approved):

docs/go/activities.md
flossypurse: Thanks for the feedback @asonawalla

PR #659, opened 2021-09-30
Update node/package-initializer.md

lorensr:
## What does this PR do?

Updates in line with:

https://github.com/temporalio/sdk-node/pull/273

and

https://github.com/temporalio/samples-node/pull/12

Leaving out [lesser-used options](https://github.com/temporalio/sdk-node/blob/main/packages/create-project/src/index.ts)


sw-yx:
its failing prettier but I ran prettier locally and it seemed fine.

PR #660, opened 2021-09-30
Fix GH action and update lock files

lorensr:
- Hopefully fixed CI, which previously only worked if the PR branch was in the `temporalio/documentation` repo. See for example: https://github.com/temporalio/documentation/pull/659/checks?check_run_id=3751594991
- Updated `yarn.lock`

> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

- Removed `package-lock.json`, which looked like an accidental addition, as `yarn.lock` has a longer history, and `yarn` is used in CONTRIBUTING. Added to `.gitignore` to avoid it happening in future.


lorensr:
image

Yay! 💃

When merged, we can see if `Checkout on push` to master step also works.

flossypurse (approved):
awesome

PR #674, opened 2021-10-08
document Netflix usecase of task queues for traffic routing

sw-yx:
## What does this PR do?

document Netflix usecase of task queues for traffic routing


sw-yx:
dont know how to put this into the new docs structure so giving up

PR #727, opened 2021-11-04
Add a section about sticky queues

Sushisource:
## What does this PR do?
Add a brief explanation on sticky queues

### DIFF
--- a/docs/concepts/task-queues.md
+++ b/docs/concepts/task-queues.md
@@ -80,5 +80,5 @@ The [Go file processing example](https://github.com/temporalio/samples-go/tree/m
 Workflow workers by default cache workflows they have recently executed.
 To efficiently make progress on cached workflows, Workers with the Workflow cache enabled (on by default) will automatically create a Task Queue specific to that Worker called a "Sticky Queue", and the server will send workflow tasks with incremental history updates (rather than entire histories) to that worker as long as it is still polling from this Sticky Queue.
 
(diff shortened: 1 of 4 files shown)


docs/concepts/task-queues.md
```diff
@@ -74,3 +74,15 @@ You can find implementation examples that illustrate this technique for the foll
...
+
+Workflow workers by default cache workflows they have recently executed. To efficiently make progress on cached workflows,
+Workers with the Workflow cache enabled (on by default) will automatically create a Task Queue specific to that Worker called a "Sticky Queue", and the server will send
+workflow tasks with incremental history updates (rather than entire histories) to that worker as long as it is still polling
+from this Sticky Queue.
+
+Sticky Queues are typically transparent to you as a user, but SDKs do expose a `stickyQueueScheduleToStartTimeout` option which
+defines how long a task will remain in a sticky queue before being shifted back to a normal task queue (with full history)
+to be picked up by any worker. Practically speaking, it defines how long it takes before the server considers a worker as no
+longer present.
```
sw-yx, line 88:
```suggestion
## Sticky Queues

Workflow workers by default cache workflows they have recently executed.
To efficiently make progress on cached workflows, Workers with the Workflow cache enabled (on by default) will automatically create a Task Queue specific to that Worker called a "Sticky Queue", and the server will send workflow tasks with incremental history updates (rather than entire histories) to that worker as long as it is still polling from this Sticky Queue.

Users do not typically have to know about Sticky Queues, but SDKs do expose a `stickyQueueScheduleToStartTimeout` option on Workers, which defines how long a task will remain in a Sticky Queue before being shifted back to a normal task queue (with full history) to be picked up by any worker (default is 10 seconds). 
In practice, this timeout controls how long it takes before the server considers a worker as no longer present.
```

sw-yx (approved):

PR #733, opened 2021-11-06
add signals and queries design patterns

sw-yx:
## What does this PR do?

add signals and queries design patterns to spell things out a bit

### DIFF
--- a/docs/typescript/workflows.md
+++ b/docs/typescript/workflows.md
@@ -204,29 +204,105 @@ By design of these Workflow handles, two different Workflows can use the same Si
 
 Because Signals and Queries are intentionally flexible, you can wrap them up into reusable functions:
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs
+defaultValue="export"
+values={[
+{label: 'Exportable', value: 'export'},
+{label: 'Concise', value: 'short'},
+]
+}>
+
+<TabItem value="export">
+
 ```ts
-// implementation
-function QueryableState<T = any>(name: string, initialValue: T) {
+// implementation of queryable + signallable State in Workflow file
+import {
+  defineSignal,
+  defineQuery,
+  setHandler,
+  sleep,
+} from '@temporalio/workflows';
+
+function useState<T = any>(name: string, initialValue: T) {
   const signal = defineSignal<[T]>(name);
   const query = defineQuery<T>(name);
   let state: T = initialValue;
-  setHandler(signal, (newValue: T) => void (state = newValue));
-  setHandler(query, () => state);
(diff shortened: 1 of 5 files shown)


docs/typescript/workflows.md
```diff
@@ -126,14 +126,50 @@ Listeners for both Signals and Queries can take arguments, which can be used ins
...
+    type: 'signal',
+    name,
+  };
+}
+
+/**
+ * Define a query method for a Workflow.
+ *
+ * Definitions are used to register handler in the Workflow via {@link setHandler} and to query Workflows using a {@link WorkflowHandle}.
+ * Definitions can be reused in multiple Workflows.
```
sw-yx, line 149:
```suggestion
 */
export function defineSignal<Args extends any[] = []>(name: string): SignalDefinition<Args> {
  return {
    type: 'signal',
    name,
  };
}

/**
 * Define a query method for a Workflow.
 *
```

docs/typescript/workflows.md
```diff
@@ -126,14 +126,54 @@ Listeners for both Signals and Queries can take arguments, which can be used ins
...
+ */
+export function defineQuery<Ret, Args extends any[] = []>(
+  name: string
+): QueryDefinition<Ret, Args> {
+  return {
+    type: 'query',
+    name,
+  };
+}
```
sw-yx, line 160:
```suggestion
/**
 * Define a signal method for a Workflow.
 */
export function defineSignal<Args extends any[] = []>(name: string): SignalDefinition<Args> {
  return {
    type: 'signal',
    name,
  };
}

/**
 * Define a query method for a Workflow.
 */
export function defineQuery<Ret, Args extends any[] = []>( name: string): QueryDefinition<Ret, Args> {
  return {
    type: 'query',
    name,
  };
}
```

docs/typescript/workflows.md
```diff
@@ -126,14 +126,48 @@ Listeners for both Signals and Queries can take arguments, which can be used ins
...
+```
+
+Signals/Queries are only instantiated in `setHandler` and are specific to a particular Workflow Execution.
 
 These distinctions may seem minor, but they model how Temporal works under the hood, because Signals and Queries are messages identified by "just strings" and don't have meaning independent of the Workflow having a listener to handle them.
+This will be clearer if you refer to to the Client-side APIs below.
 
 #### Why `setHandler` and not OTHER_API?
 
 We named it `setHandler` instead of `subscribe` because Signals/Queries can only have one listener at a time, whereas `subscribe` could imply an Observable with multiple consumers.
```
sw-yx, line 164:
```suggestion
#### Why `setHandler` and not OTHER_API?

We named it `setHandler` instead of `subscribe` because Signals/Queries can only have one "handler" at a time, whereas `subscribe` could imply an Observable with multiple consumers, and is a higher level construct.
```

PR #744, opened 2021-11-12
add signals ordering notes based on user question

sw-yx:
## What does this PR do?

add signals ordering notes based on user question

### DIFF
--- a/docs/typescript/workflows.md
+++ b/docs/typescript/workflows.md
@@ -363,7 +363,7 @@ let state = await handle.query<number, [string]>('print', 'Count: ');
 #### Notes on Signals
 
 - Signal handlers are only guaranteed to be called in order **per Signal Type**, not across all of them.
-   If you need strict ordering across multiple Signals, combine them into one Signal Type and use a `switch` statement.
+  If you need strict ordering across multiple Signals, combine them into one Signal Type and use a `switch` statement.
 - `WorkflowHandle.signal` resolves as soon as Temporal Server has persisted the Signal, before the Workflow's Signal handler is called.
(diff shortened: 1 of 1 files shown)


docs/typescript/workflows.md
```diff
@@ -362,22 +362,18 @@ let state = await handle.query<number, [string]>('print', 'Count: ');
...
-This Promise resolves with no value; **Signal handlers cannot return data to the caller.**
-
-:::info No Synchronous Updates
-
-A common request is for a Signal to be invoked with a bad argument, causing a validation error.
-However Temporal has no way to surface the error to the external invocation.
-Signals and Queries are always asynchronous, in other words, **a Signal always succeeds**.
-
-The solution to this is "Synchronous Update" and we plan to add it in future.
+- Signals are only guaranteed to arrive in order **per Signal**, not across Signals. If you need strict ordering across multiple Signals, combine them into one Signal and use a `switch` statement.
```
Sushisource, line 365: This should be more like "signal *handlers* are only guaranteed to be called in order..."

Otherwise what does arrival mean? At the server? At the client? At the workflow? We're actually talking about the last one here.

The advice to combine them into one signal is a bit confusing, because you can read that as "one call to the signal api" which might be impossible if the information comes from disparate processes. Simple fix is probably just "combine them into one signal type".

sw-yx:
```suggestion
- Signal handlers are only guaranteed to be called in order **per Signal Type**, not across all of them.
   If you need strict ordering across multiple Signals, combine them into one Signal Type and use a `switch` statement.
```

Sushisource: :+1:

PR #753, opened 2021-11-18
fix broken JS/TS switch in TypeScript examples

vkarpov15:
## What does this PR do?

Turns out there were a couple of issues with the JS/TS compiler:

1) Needed to change "node" to "typescript" in the tsconfig path:

```javascript
              {
                compilerSettings: {
                  tsconfig: path.join(
                    __dirname,
                    "docs",
                    "typescript", // <-- this was still "node"
                    "tsconfig.json"
                  ),
                  externalResolutions: {},
                }
```

2) Several syntax errors in the tutorials were causing the build to fail

Here's what it looks like on my machine. Will confirm that it works on Netlify deploy preview as well.

image

### DIFF
--- a/docs/typescript/activities.md
+++ b/docs/typescript/activities.md
@@ -148,11 +148,8 @@ Activities are Promises and you may retrieve multiple Activities from the same h
 ```ts
 export async function Workflow(name: string): Promise<string> {
   // destructuring multiple activities with the same options
-  const {
-    act1,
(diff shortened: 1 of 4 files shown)


bergundy (approved):

bergundy:
Thanks @vkarpov15!

sw-yx:
ah nice find

PR #768, opened 2021-12-02
fix code sample syntax for build

flossypurse:
I was getting this error when trying to build master:

```
TypeScript error in code block in line 16 of /Users/cullywakelin/Temporal/documentation/docs/typescript/workflows.md/codeBlock_1/index.ts
Expression expected.
````

This PR has a small fix for that.


sw-yx:
did this fix just as cully did it

PR #801, opened 2021-12-26
Update MySignal struct in signals documentation

jeromefroe:
## What does this PR do?

This commit updates the `MySignal` struct that is used in the Signals documentation to add a `message` field since the comment on the struct references such a field:

> Temporal serializes `Message`, not `message` because fields that start with lowercase letters are private in Go.

Before this change:

Screen Shot 2021-12-26 at 5 10 45 PM

After this change:

Screen Shot 2021-12-26 at 5 10 55 PM


flossypurse (approved):
Cool, good stuff - thank you!

PR #802, opened 2021-12-26
Update SimpleActivity function in activites documentation

jeromefroe:
## What does this PR do?

This commit updates the `SimpleActivity` function in the activities documentation to append a string to the value that it returns. This will ensure the activity matches the documentation for it:

> The following example demonstrates a simple Activity Definition that accepts a string parameter, appends a word to it, and then returns a result.

Before this change:

Screen Shot 2021-12-26 at 5 29 57 PM

After this change:

Screen Shot 2021-12-26 at 5 29 57 PM

### DIFF
--- a/docs/go/activities.md
+++ b/docs/go/activities.md
@@ -21,12 +21,13 @@ import (
 	"go.temporal.io/sdk/activity"
 )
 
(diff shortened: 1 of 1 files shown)


flossypurse (approved):

flossypurse:
Awesome thank you !

PR #805, opened 2021-12-29
feat: font preloading

johnnyreilly:
## What does this PR do?

This PR adds font preloading to the Docusaurus build. Effectively it's just applying the changes [documented here](https://blog.johnnyreilly.com/2021/12/29/preload-fonts-with-docusaurus) (by me).

A little context: I borrowed the implementation of custom font loading for my own Docusaurus blog from here.  I really liked having a different font. I didn't liked the [flash of unstyled text](https://css-tricks.com/fout-foit-foft/). So, font preloading FTW!

This PR does the following:

- adds the [`webpack-font-preload-plugin`](https://github.com/sn-satyendra/webpack-font-preload-plugin) as a dependency. This is a webpack plugin for font preloading.
- uses the webpack plugin to create a Docusaurus plugin in the `docusaurus.config.js` file.
- fixes the usage of the `@type {import('@docusaurus/types').DocusaurusConfig}` JSDoc tag; it wasn't in front of the `module.exports = ` and so I don't think it will have been providing the benefit it could. 

The effect of this PR is to make the fonts load much earlier.  Consider this screenshot of the network tab in Chrome devtools loading the Temporal docs:

screenshot of the network tab in Chrome devtools loading the Temporal docs with the font requests fairly low down

As compared to this, after the change is applied:

screenshot of the network tab in Chrome devtools loading the Temporal docs with the font requests fairly high up

In the first screenshot the fonts are one of the later requests, in the second they are one of the first requests; due to the preload elements introduced:

screenshot of preload elements in Chrome devtools 

This reduces the FOUT and should provide a better Lighthouse score to boot! 

## Notes to reviewers

Discussed with @sw-yx prior to PR.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -1,8 +1,9 @@
-/** @type {import('@docusaurus/types').DocusaurusConfig} */
-
+//@ts-check
 const path = require("path");
 const visit = require("unist-util-visit");
+const FontPreloadPlugin = require("webpack-font-preload-plugin");
 
+/** @type {import('@docusaurus/types').DocusaurusConfig} */
 module.exports = {
   title: "Temporal Documentation",
   tagline: "Build invincible applications",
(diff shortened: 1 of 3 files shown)


sw-yx (approved):
thanks so much John!

PR #812, opened 2022-01-06
Add variable explanations

laniehei:
## What does this PR do?
This adds information about dependencies to the example workflows.

## Notes to reviewers
It was unclear to me what "client" and "c" meant while I was reading the docs.

### DIFF
--- a/docs/go/workflows.md
+++ b/docs/go/workflows.md
@@ -10,6 +10,7 @@ description: In the Temporal Go SDK programming model, a Workflow is an exportab
 You can execute Workflows (including those from other language SDKs) by their type name:
 
 ```go
+
 workflowID := "myworkflow_" + uuid.New()
 workflowOptions := client.StartWorkflowOptions{
   ID:        workflowID,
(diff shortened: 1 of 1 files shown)


docs/go/workflows.md
```diff
@@ -23,7 +24,7 @@ if err != nil {
 log.Println("Started workflow", "WorkflowID", we.GetID(), "RunID", we.GetRunID())
 ```
 
-Here we execute a workflow by its type name, namely `MySimpleWorkflow`. By default, the
+In the above example, client is imported from "go.temporal.io/sdk/client". 'c' is the temporal client instantiated outside of the snippet. Here we execute a workflow by its type name, namely `MySimpleWorkflow`. By default, the
```
flossypurse, line 27: Thanks for making this clear - though I think we should just rename the variable in the sample to be `temporalClient`:

```
workflowOptions := temporalClient.StartWorkflowOptions{
// ...
}
we, err := temporalClient.ExecuteWorkflow(context.Background(), workflowOptions, FunctionName)
// ...
```

laniehei: I don't think the above example is valid, because StartWorkflowOptions and Execute workflow are different things. StartWorkflowOptions is on the `go.temporal.io/sdk/client` imported package, and ExecuteWorklow is on the returned struct from `client.NewClient()`.

flossypurse: Ya, you are right, my bad, the options come from the client package - 
So you can ignore that one as an example -
My point is that, any variable is a result of `client.NewClient()` should be named `temporalClient` in our examples.

laniehei: done!

flossypurse:
```suggestion
In the above example, `client` is imported from the "go.temporal.io/sdk/client" package. And `temporalClient` is an instance of the Temporal Client instantiated outside of the snippet.
Here we spawn a Workflow Execution by its Workflow Type name `MySimpleWorkflow`. By default, the
```

flossypurse (approved):

PR #828, opened 2022-01-18
Add metrics overview page

Sushisource:

### DIFF
--- a/docs/concepts/metrics.md
+++ b/docs/concepts/metrics.md
@@ -66,6 +66,14 @@ Note that `Core` includes the Typescript SDK.
 | workflow_task_replay_latency              | Time it takes to catch up on replaying a WFT in ms                                                                                                              | Histogram   | Core, Go, Java                              | namespace, task_queue, workflow_type                |
 | workflow_task_schedule_to_start_latency   | Workflow task schedule-to-start time in milliseconds                                                                                                            | Histogram   | Core, Go, Java                              | namespace, task_queue, workflow_type                |
 
+#### Code definitions
+
+If you're curious about how the metrics are defined in source, see:
+
+- Core - [SDK metrics](https://github.com/temporalio/sdk-core/blob/master/core/src/telemetry/metrics.rs) and [Client metrics](https://github.com/temporalio/sdk-core/blob/master/client/src/metrics.rs)
+- Java - [SDK metrics](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/internal/metrics/MetricsType.java) and [Client metrics](https://github.com/temporalio/sdk-java/blob/master/temporal-serviceclient/src/main/java/io/temporal/serviceclient/MetricsType.java)
+- Go - [SDK & Client metrics](https://github.com/temporalio/sdk-go/blob/c32b04729cc7691f80c16f80eed7f323ee5ce24f/internal/common/metrics/constants.go)
+
 ### Notes on differences between SDKs
 
 - Histograms may have different buckets in different languages.
--- a/docs/concepts/metrics.md
+++ b/docs/concepts/metrics.md
@@ -1,0 +1,76 @@
+---
+id: Metrics
+title: Temporal Metrics
+description: Temporal and Workers both emit metrics| this page describes them
+sidebar_label: Metrics
+---
+
+## Worker/SDK Metrics
+
+The Temporal SDKs emit a largely common set of metrics. Note that all the below metrics are prefixed
+with `temporal_` before being exported to their configured destination(s). `Core` 
+
+### Keys
+The metrics always have some keys/labels attached to them. Their definitions:
+
+- `task-queue` - Task queue the worker polls on
+- `namespace` - Namespace the worker is bound to
+- `poller_type` - One of `workflow_task`, `activity_task`, or `sticky_workflow_task`
+- `worker_type` - One of `ActivityWorker`, `WorkflowWorker`, or `LocalActivityWorker` (Go only)
+- `activity_type` - The function name / class name / etc of the activity the metric is associated with
+- `workflow_type` - The function name / class name / etc of the workflow the metric is associated with
+- `operation` - Only for client gRPC requests, is the name of the RPC method
+
+### Metrics definitions
+
+Note that `Core` includes the Typescript SDK.
+
+| Name                                      | Desc                                                                                                                                                            | Metric Type | Exists In SDK                               | Keys                                                |
+|-------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|---------------------------------------------|-----------------------------------------------------|
+| activity_execution_cancelled              | An activity execution was cancelled                                                                                                                             | Counter     | Java                                        | activity_type, namespace, task_queue                |
+| activity_execution_failed                 | An activity execution failed                                                                                                                                    | Counter     | Core, Go, Java                              | activity_type, namespace, task_queue                |
+| activity_execution_latency                | Time it took to complete activity execution, from the time core generated the activity task, to the time lang responded with a completion (failure or success). | Histogram   | Core, Go, Java                              | activity_type, namespace, task_queue                |
+| activity_poll_no_task                     | An activity long poll timed out                                                                                                                                 | Counter     | Core, Go, Java                              | activity_type, namespace, task_queue                |
+| activity_schedule_to_start_latency        | Activity task schedule to start time in millis                                                                                                                  | Histogram   | Core, Go, Java                              | activity_type, namespace, task_queue                |
+| activity_task_error                       | An internal error / panic occurred during activity task handling/execution                                                                                      | Counter     | Go                                          | activity_type, namespace, task_queue, workflow_type |
+| corrupted_signals                         | Number of signals whose payload could not be deserialized                                                                                                       | Counter     | Go, Java                                    | namespace, task_queue, workflow_type                |
+| local_activity_execution_cancelled        | A local activity execution was cancelled                                                                                                                        | Counter     | Go, Java                                    | activity_type, namespace, task_queue                |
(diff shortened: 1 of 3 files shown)


docs/concepts/metrics.md
```diff
@@ -0,0 +1,85 @@
...
+- Go - [SDK & Client metrics](https://github.com/temporalio/sdk-go/blob/c32b04729cc7691f80c16f80eed7f323ee5ce24f/internal/common/metrics/constants.go)
+
+### Notes on differences between SDKs
+
+- Histograms may have different buckets in different languages.
+- Some labels may not apply in every SDK
+
+## Server Metrics
+
+Documentation coming soon
```
bergundy, line 85: I think we have some docs about server side metrics.
@flossypurse?

bergundy (approved):

Sushisource:
Also dark mode table headers look unreadable. I'll fix in this PR if someone points me where

docs/concepts/metrics.md
```diff
@@ -0,0 +1,85 @@
+---
+id: metrics
```
flossypurse, line 2:
```suggestion
id: worker-metrics
```
I think this is more accurate?

docs/concepts/metrics.md
```diff
@@ -0,0 +1,85 @@
...
+title: Temporal Metrics
```
flossypurse, line 3:
```suggestion
title: Worker Entity metrics reference
```
Is this more accurate?

docs/concepts/metrics.md
flossypurse: Can we aim to put those into a separate file?
i.e. Cluster metrics reference ?

docs/concepts/metrics.md
```diff
@@ -0,0 +1,85 @@
...
+- `worker_type` - One of `ActivityWorker`, `WorkflowWorker`, or `LocalActivityWorker` (Go only)
+- `activity_type` - The function name / class name / etc of the activity the metric is associated with
+- `workflow_type` - The function name / class name / etc of the workflow the metric is associated with
+- `operation` - Only for client gRPC requests, is the name of the RPC method
+
+### Metrics definitions
+
+Note that `Core` includes the Typescript SDK.
+
+| Name                                      | Desc                                                                                                                                                            | Metric Type | Exists In SDK                               | Keys                                                |
```
flossypurse, line 29: This table is a bit of a pain to maintain.
Also, I can't guarantee the info in this table is indexed and searchable.

My strong recommendation would be to convert each metric to its own h3 like we have done with these:
https://docs.temporal.io/docs/reference/events/
https://docs.temporal.io/docs/reference/commands/

However, if we want to keep the table, then we could do a hybrid approach like this:
https://docs.temporal.io/docs/go/how-to-set-startworkflowoptions-in-go

And limit the table columns and info -

docs/concepts/metrics.md
```diff
@@ -0,0 +1,85 @@
+---
+id: metrics
+title: Temporal Metrics
+description: Temporal and Workers both emit metrics| this page describes them
```
flossypurse, line 4: I am thinking that this file should be scoped specifically to the metrics that are emitted from the "SDK" side of things -
On that note, where are these metrics emitted from specifically?
We are saying "SDK metrics", but it seems like these are metrics emitted from Worker Entities.
I.e. if there are multiple Worker Entities in a Worker Process, then each one would be emitting these metrics ?
If we end up adding Cluster metrics, then we can give that its own file as well.

docs/concepts/metrics.md
```diff
@@ -0,0 +1,85 @@
...
+sidebar_label: Metrics
```
flossypurse, line 5:
```suggestion
sidebar_label: Worker Entity metrics
tags:
    - metrics
    - reference
```

docs/concepts/metrics.md
```diff
@@ -0,0 +1,85 @@
...
+---
+
+## Worker/SDK Metrics
+
+The Temporal SDKs emit a largely common set of metrics. Note that all the below metrics are prefixed
+with `temporal_` before being exported to their configured destination(s). `Core`
```
flossypurse, line 11:
```suggestion
Each [Worker Entity](/docs/content/what-is-a-worker-entity) emits metrics.
Each SDK varies in which metrics are emitted.

All the metrics listed below are prefixed
with `temporal_` before being exported to their configured destination(s).
```
Not sure what `Core` is for there?

sidebars.js
```diff
@@ -382,6 +382,7 @@ module.exports = {
         "concepts/workflows",
         "concepts/activities",
         "temporal-explained/task-queues-and-workers",
+        "concepts/metrics",
```
flossypurse, line 385:
```suggestion
      "reference/worker-metrics",
```

Mind if we move this file to the `/docs/reference` directory?
And to be honest, this is a bit out of place for the concepts section - 
However there are other places we can link to it such as the "What is a Worker Entity" page -
We are aiming to do a revamp of the landing page soon that should highlight references like this.

docs/concepts/metrics.md
Sushisource: Like @sw-yx mentioned can we figure out if there's a way _to_ make it index-able? The table is *dramatically* more readable IMO. The hybrid is an OK hack, but it's still a hack at the end of the day. IMO a table is pretty clearly the right format for this information.

docs/concepts/metrics.md
Sushisource: Kinda. It's more like Worker & SDK metrics. I'll go for that

docs/concepts/metrics.md
Sushisource: Forgot to finish that sentence. Fixed. Not a huge fan of "worker entity" here as the SDK metrics include things that aren't necessarily tied to a specific worker, though most are.

sidebars.js
Sushisource: Done

docs/reference/metrics.md
```diff
@@ -0,0 +1,82 @@
...
+- `operation` - Only for client gRPC requests, is the name of the RPC method
+
+### Metrics definitions
+
+Note that `Core` includes the Typescript SDK.
+
+| Name                                      | Desc                                                                                                                                                            | Metric Type | Exists In SDK                               | Keys                                                |
+| ----------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------- | ------------------------------------------- | --------------------------------------------------- |
+| activity_execution_cancelled              | An activity execution was cancelled                                                                                                                             | Counter     | Java                                        | activity_type, namespace, task_queue                |
```
flossypurse, line 32:
```suggestion
| activity_execution_cancelled              | An Activity Execution was canceled                                                                                                                             | Counter     | Java                                        | activity_type, namespace, task_queue                |
```
Mind capitalizing all core concept terms?
https://docs.temporal.io/docs/reference/glossary/

docs/reference/metrics.md
```diff
@@ -0,0 +1,82 @@
+---
+id: worker-metrics
```
flossypurse, line 2:
```suggestion
id: sdk-metrics
```

Sounds like this is the appropriate slug?

docs/concepts/metrics.md
flossypurse: Yes, there is a way to index tables. We will definitely look into it. But it hasn't been needed yet.
Thinking about this more - the real reason h3s are the best here is because 1 they provide us with the ability to link directly to the metric name itself. 2 we have a place to add more info per metric - tables do not support multi line content, and we can only have so many columns before it becomes just too much. I don't think of it as a hack. I think of it as a very strategic formatting decision. I wonder if the "tables" were turned (pun intended) and I came into your world of expertise and started calling all of your decisions "hacks" how you might respond to that?

docs/concepts/metrics.md
flossypurse: We do have this: https://docs.temporal.io/docs/server/production-deployment#scaling-and-metrics 
But it is not in a reference format or complete.

docs/concepts/metrics.md
flossypurse: How would I know the difference?
Seems like we could identify, for each metric, where it is emitted from - i.e Worker Entity or Temporal Client?

docs/concepts/metrics.md
Sushisource: Sure, sorry, just my point is doesn't sound like you're saying the table is actually the wrong way to show this other than limitations imposed by the platform we're using. That's what I mean by a hack. If the limitations can be overcome, we would prefer to use a table.

If we can overcome the direct linking limitation, then the table is more compact and readable. As for adding more columns / content, sure, I can see breaking it out then, but the information needed here is unlikely to grow. If it does, we could break it out then.

flossypurse: I mean, I think it is more about what is practical in the context of the given situation.
And right now, the context of the situation is this:

1. I don't plan on spending any time in the next few weeks messing with the Algolia config to support tables - perhaps someone else might have time to do it.
2. I don't know of any way atm to overcome the direct linking of table items - which is very handy when dropping links to users in support.
3. We do try to plan for information growth - there may be any number of things that we may want to add here.
4. The table I am looking at is IMO definitely NOT compact - it is huge, and width wise, doesn't fit on the page - It definitely doesn't fit mobile or tablet screens, which a decent percentage of traffic still happens to be.
5. A stand alone table doesn't match the format of the other references we have made.

So, I get that you don't want to do more work here.
But I think you can appreciate my desire to ensure this meets some level of "quality" up front and we aren't just adding more debt for us to go "clean up" later on.

sidebars.js
```diff
@@ -382,6 +382,7 @@ module.exports = {
         "concepts/workflows",
         "concepts/activities",
         "temporal-explained/task-queues-and-workers",
+        "reference/worker-metrics",
```
flossypurse, line 385:
```suggestion
        "reference/sdk-metrics",
```

flossypurse (approved):

Sushisource:
@flossypurse Thank you! I was gonna come back here and do some of the formatting stuff but was focused on other work. Really appreciate getting this in so we've got something to start with at least :smile:

PR #840, opened 2022-01-22
docs: briefly mention activities calling activities

JoshuaKGoldberg:
## What does this PR do?

Mentions that calling an activity from another activity is generally not recommended.

I'd chatted about this with @sw-yx as a small docs mention / FAQ a bit back. No worries if you don't think this warrants a full docs mention!

### DIFF
--- a/docs/typescript/activities.md
+++ b/docs/typescript/activities.md
@@ -341,7 +341,7 @@ Async activity completion is done through a two step process:
 
 You can [read the tests](https://github.com/temporalio/sdk-typescript/blob/7d47f501cb56cced27118b5f0abb320cc0ba03ef/packages/test/src/test-async-completion.ts#L40-L98) for more information.
 
-## Activities calling Activities
+### Activities calling Activities
 
 It may sometimes seem convenient in code to have an activity call another activity.
 We generally recommend not to: activities are regular functions, so calling to one directly will not be seen -and therefore logged- by the Temporal service.
--- a/docs/typescript/activities.md
+++ b/docs/typescript/activities.md
@@ -341,14 +341,6 @@ Async activity completion is done through a two step process:
 
 You can [read the tests](https://github.com/temporalio/sdk-typescript/blob/7d47f501cb56cced27118b5f0abb320cc0ba03ef/packages/test/src/test-async-completion.ts#L40-L98) for more information.
 
-### Activities calling Activities
-
-For some use cases, having an Activity call another Activity might seem convenient.
-We generally recommend not doing so. Activities are regular functions, so calling one directly is not seen—and therefore not logged—by the Temporal Server.
-
-Instead, move logic out of the Activities and have the parent Workflow use the result of one Activity to call the other Activity.
-
-
 ### Local Activities
 
 Temporal has an optimization feature called Local Activities.
--- a/docs/typescript/activities.md
+++ b/docs/typescript/activities.md
@@ -341,6 +341,14 @@ Async activity completion is done through a two step process:
 
 You can [read the tests](https://github.com/temporalio/sdk-typescript/blob/7d47f501cb56cced27118b5f0abb320cc0ba03ef/packages/test/src/test-async-completion.ts#L40-L98) for more information.
 
+## Activities calling Activities
+
+It may sometimes seem convenient in code to have an activity call another activity.
+We generally recommend not to: activities are regular functions, so calling to one directly will not be seen -and therefore logged- by the Temporal service.
+
+Instead, prefer moving logic out of the activities so that the parent workflow uses the result of one activity to call the other activity.
+
+
 ### Local Activities
 
 Temporal has an optimization feature called Local Activities.
--- a/docs/typescript/activities.md
+++ b/docs/typescript/activities.md
@@ -343,10 +343,10 @@ You can [read the tests](https://github.com/temporalio/sdk-typescript/blob/7d47f
 
 ### Activities calling Activities
 
-It may sometimes seem convenient in code to have an activity call another activity.
-We generally recommend not to: activities are regular functions, so calling to one directly will not be seen -and therefore logged- by the Temporal service.
+For some use cases, having an Activity call another Activity might seem convenient.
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
Ah, PR checkout woes:

```
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/activities-calling-activities*:refs/remotes/origin/activities-calling-activities* +refs/tags/activities-calling-activities*:refs/tags/activities-calling-activities*
  The process '/usr/bin/git' failed with exit code 1
  Waiting 19 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/activities-calling-activities*:refs/remotes/origin/activities-calling-activities* +refs/tags/activities-calling-activities*:refs/tags/activities-calling-activities*
  The process '/usr/bin/git' failed with exit code 1
  Waiting 20 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/activities-calling-activities*:refs/remotes/origin/activities-calling-activities* +refs/tags/activities-calling-activities*:refs/tags/activities-calling-activities*
  Error: The process '/usr/bin/git' failed with exit code 1
```

https://github.com/temporalio/documentation/runs/4904810202?check_suite_focus=true

flossypurse:
Thanks @JoshuaKGoldberg !
I think this an interesting point that should be made in the high level "What is an Activity?" or description: https://docs.temporal.io/docs/content/what-is-an-activity/ (its a WIP page that will eventually be the truth of source for Activity concepts)

In terms of implementation patterns, it seems idiomatic across languages to break out Activity logic into sub-functions as needed.
But it is an anti-pattern to call an Activity from another Activity. And to support that notion, there is no API for it.
i.e - there is no "ExecuteActivity()" API call that can be used within an Activity.
So, you can string together sub-functions within an Activity - just know that if any part of it fails, then the whole set of logic would be executed again.

flossypurse:
Also @JoshuaKGoldberg don't worry about the prettier integration failure.

docs/typescript/activities.md
```diff
@@ -341,6 +341,14 @@ Async activity completion is done through a two step process:
 
 You can [read the tests](https://github.com/temporalio/sdk-typescript/blob/7d47f501cb56cced27118b5f0abb320cc0ba03ef/packages/test/src/test-async-completion.ts#L40-L98) for more information.
 
+## Activities calling Activities
```
sw-yx, line 344:
```suggestion
### Activities calling Activities
```

sw-yx (approved):

docs/typescript/activities.md
```diff
@@ -341,6 +341,14 @@ Async activity completion is done through a two step process:
 
 You can [read the tests](https://github.com/temporalio/sdk-typescript/blob/7d47f501cb56cced27118b5f0abb320cc0ba03ef/packages/test/src/test-async-completion.ts#L40-L98) for more information.
 
+### Activities calling Activities
+
+It may sometimes seem convenient in code to have an activity call another activity.
+We generally recommend not to: activities are regular functions, so calling to one directly will not be seen -and therefore logged- by the Temporal service.
+
+Instead, prefer moving logic out of the activities so that the parent workflow uses the result of one activity to call the other activity.
```
aarohib, line 349: Rephrase suggestion: "Activities are regular functions. Calling another Activity from within an Activity will not be logged by the Temporal service, and is therefore not a recommended practice. Instead, move logic out of the Activities so that the parent Workflow uses the result of one Activity to call the other Activity."

JoshuaKGoldberg:
@flossypurse sure, makes sense to me! Do you want me to directly cut-and-paste it over?

Since aarohib pinged you for rephrasing help _(+1, I'm not sold on how it is now)_, I'll hold off making changes unless explicitly asked.

docs/typescript/activities.md
djmagee, line 349:
```suggestion
Instead, move logic out of the Activities and have the parent Workflow use the result of one Activity to call the other Activity.
```

tsurdilo:
The way the docs describe it is one way that can be done, but 
note that activity impl is just code (programming language).

"Calling an activity from another" really just means call the code of activity B from code in activityA directly (as code). 
In many cases breaking these up into two activity calls makes sense but in many it does not imo and direct calls via programming language are preferred.

tsurdilo:
Info in this forum post:https://community.temporal.io/t/how-to-call-an-activity-from-other-activity/1294/2
would also be nice to be added to docs maybe in some shape / form to enhance this new section. wdyt?

flossypurse:
@JoshuaKGoldberg yes - please copy your suggestion as is to /docs/content/what-is-an-activity

And then we can work out phrasing there -

@tsurdilo yes - please also extract info from that community thread and add it to /docs/content/what-is-an-activity

@djmagee, @aarohib  - once that happens, we can then reorganize and copyedit as needed.

What we will end up doing is making sure that an Activity is defined as a function that is called by a Workflow - then someone can reason that they won't ever call an Activity from another Activity - because that is simply calling another function from within a function - and by definition is NOT calling another Activity.
The point we need to make that adds to someone's ability to reason here - is what @JoshuaKGoldberg touches on with this sentence:

> activities are regular functions, so calling to one directly will not be seen -and therefore logged- by the Temporal service.

The whole point is to incrementally capture the change in state. 
Activities are just a way to do that for unreliable steps in the process. If an Activity fails - any retry starts from initial state. From there you can reason that - of course you don't want to do multiple unreliable steps in a process in a single Activity - unless you absolutely have to - because each time it starts from initial state.

tsurdilo:
@flossypurse just to add regarding any retry starts from initial activity state:
you register a single activity impl with worker so all workflow runs are using the same activity impl the starting from "blank" state is a recommendation (thread safety) but its not enforced by temporal. you can write activity code to keep some state between retries if you wanted.

flossypurse (approved):

flossypurse:
I think we need to work out how we are going to talk about this internally for a bit - 
For now, let us just roll with this - and we will try to address the larger explanation sooner than later.

PR #848, opened 2022-01-27
tctl reference: MDX-free activity, batch, taskqueue, and top-level pages

djmagee:
## What does this PR do?

- Replaces MDX with pure Markdown
- Changes "option" to "modifier" globally
- Adds "required" tag to appropriate modifiers
- Adds required modifiers to all examples

## Notes to reviewers

- This PR changes the presentation significantly. Compare these pages with the current versions.
- Style questions: What is your opinion of...
  - ...using monospace (inline code) for links to first-level and second-level command pages?
  - ...the styling for required modifiers? (I tried adding "(required)" to the H3 headings, but the result was suboptimal.)
- Required modifiers for some `tctl workflow` and `tctl activity` commands are not yet tagged. I'm awaiting guidance from Ruslan.

## Preview

https://deploy-preview-848--mystifying-fermi-1bc096.netlify.app/docs/reference/tctl/

PR #856, opened 2022-01-30
tctl reference: 5 MDX-free workflow pages

djmagee:
## What does this PR do?

- Replaces MDX with pure Markdown
- Changes "option" to "modifier" globally

## Notes to reviewers

- See also "Notes to reviewers" in #848 
- Required modifiers are not yet tagged, and examples do not yet include required modifiers. I'm awaiting guidance from Ruslan.


aarohib (approved):

docs/reference/tctl/workflow.md
```diff
@@ -7,121 +7,28 @@ tags:
...
-<TctlWorkflowReset/>
-
-## `reset-batch`
-
-import TctlWorkflowResetbatch from './workflow/reset-batch.md'
-
-<TctlWorkflowResetbatch/>
+The `tctl workflow` commands enable [Workflow Execution](/docs/content/what-is-a-workflow-execution) operations.
+
+- [`tctl workflow show`](./workflow/show)
```
flossypurse, line 12:
```suggestion
- [`tctl workflow show`](./workflow/show)
```

For consistency and predicability - lets use the full path from the project root, i.e. `/docs/reference/tctl/workflow/show`

docs/reference/tctl/workflow.md
```diff
@@ -7,121 +7,28 @@ tags:
...
+- [`tctl workflow count`](./workflow/count)
+- [`tctl workflow query`](./workflow/query)
+- [`tctl workflow stack`](./workflow/stack)
+- [`tctl workflow describe`](./workflow/describe)
+- [`tctl workflow describeid`](./workflow/describeid)
+- [`tctl workflow observe`](./workflow/observe)
+- [`tctl workflow reset`](./workflow/reset)
+- [`tctl workflow reset-batch`](./workflow/reset-batch)
+
+## Related
```
flossypurse, line 32: My instinct is that we should not be repeating this as a header on every page.

And we should not have to worry about linking back to the root of the reference in each page.
I think we will let the presentation layer handle that.

djmagee: I added these pages to the sidebar, which works nicely. (Ignore the other sub-command pages for workflow; I'll update those in subsequent PRs.)

PR #875, opened 2022-02-07
Workers Tuning Guide

Spikhalskiy:
## What does this PR do?

Adds "How to tune Workers" page with Workers Tuning Guide


docs/reference/metrics.md
```diff
@@ -51,11 +51,12 @@ Note that `Core` includes the Typescript SDK.
...
 | sticky_cache_miss                         | A workflow task did not find a cached workflow                                                                                                                  | Counter     | Core, Go, Java                              | namespace, task_queue                               |
-| sticky_cache_size                         | Current cache size in number of workflows                                                                                                                       | Gauge       | Core, Go, Java                              | namespace, task_queue                               |
+| sticky_cache_size                         | Current cache size in number of workflows                                                                                                                       | Gauge       | Core, Go, Java                              |                                                     |
```
Spikhalskiy, line 54: @Sushisource Could you please help me to verify that the metrics changes are aligned with Core?

This one is especially important, as Workflow Cache Size should be per Workers host, not per namespace and task queue. This metric/limit is very different and exists to limit the total resource consumption between all workers.

Sushisource: This is per namespace/tq. Specifically it is per worker.

Core resolves a problem that exists in Go/Java relating to controlling max cached workflows. Go and Java use one sticky queue for the host. This makes it impossible to accurately constrain the number of workflow slots to be <= the defined maximums (I can dig up where I wrote the explanation why, but, this is true).

So, Core fixes this by using unique sticky queues per normal queue.

Now, this doesn't mean that we can't *also* make a host-wide metric for this, so Core could add that, but cache limits in the config need to stay per-worker in Core.

Spikhalskiy:
I will add more info into the guide based on Spencer’s comment (to highlight the discrepancy in Core vs Go + Java) and will reopen when ready.

docs/reference/metrics.md
Spikhalskiy: As we discussed, there is no problem in Go or Java with controlling max cached workflows. There is a tricky situation with a shared poller (in JavaSDK) and an amount of executing threads, but it's not really related to cached workflow limit.
Because adjustment of maximum cached workflows should be considered when we are out (or have a surplus) of host level resources, like RAM and total system threads count, this metric mostly makes sense per process and the limit also should be enforced per process to avoid underutilization.
But for now, I just document the state as it is.

docs/content/how-to-tune-workers.md
```diff
@@ -0,0 +1,112 @@
...
+3. `sticky_cache_size` and `workflow_active_thread_count` report the size of the Workflow cache and the number of cached Workflow threads.
+
+_Note: To have access to the metrics mentioned above, JavaSDK ≥ 1.8.0 is required._
+
+## Configuration
+
+The following options are defined on `WorkerOptions` and are applicable for each Worker separately:
+
+1. `maxConcurrentWorkflowTaskExecutionSize` and `maxConcurrentActivityExecutionSize` define the number of total available slots for that Worker.
+2. `maxConcurrentWorkflowTaskPollers` (_JavaSDK: `workflowPollThreadCount`_)and `maxConcurrentActivityTaskPollers` (_JavaSDK: `activityPollThreadCount`_) define the number of pollers performing poll requests waiting on Workflow / Activity task queue and delivering the tasks to the executors.
```
Sushisource, line 32: The italics inside the parens are I think unneeded (applies a few places)

docs/content/how-to-tune-workers.md
```diff
@@ -0,0 +1,112 @@
...
+An evicted workflow execution will need to be replayed when it gets any action that may advance the execution.
+
+If
+
+1. The Workflow Cache limits described above are hit, and
+2. Worker hosts have enough free RAM and are not close to reasonable thread limits,
+
+`workflowCacheSize` and `maxWorkflowThreadCount` limits may be increased to decrease the overall latency and cost of the replays in the system. If the opposite occurs, consider decreasing the limits.
+
+_Note: In CoreSDK, this metric works differently and should be monitored and adjusted on a per Worker / Task Queue basis._
```
Sushisource, line 89: Most users don't know Core exists, so probably just say "...CoreSDK based SDKs, like TypeScript,..."

docs/content/how-to-tune-workers.md
```diff
@@ -0,0 +1,112 @@
...
+The main area of tuning should be the number of Workflow slots. If:
+
+1. the Worker hosts are underutilized (there are no bottlenecks on CPU, load average, etc), and
+2. the `worker_task_slots_available` metric from the corresponding Worker type often shows a depleted number of available Worker slots,
+
+then consider increasing the maximum number of working slots by adjusting `maxConcurrentWorkflowTaskExecutionSize` or `maxConcurrentActivityExecutionSize`.
+
+### Poller count
+
+_Note: Adjustments to pollers are rarely needed and rarely make a difference. Please consider this step only after adjusting Worker slots in the previous step. The only scenario in which the pollers’ adjustment makes sense is when there is a significant network latency between the Workers and Temporal Server._
```
Sushisource, line 60: Maybe move this section to be after the others then?

docs/content/how-to-tune-workers.md
```diff
@@ -0,0 +1,112 @@
...
+
+Perform this sanity check after the adjustments to worker settings.
+
+1. `workflowCacheSize` should be ≤ `maxWorkflowThreadCount`. Each Workflow has at least 1 Workflow thread.
+2. `maxConcurrentWorkflowTaskExecutionSize` should be ≤ `maxWorkflowThreadCount`. Having more Worker slots than the Workflow cache size will lead to resource contention/stealing between executors and unpredictable delays. It’s recommended that `maxWorkflowThreadCount` be at least 2x of `maxConcurrentWorkflowTaskExecutionSize`.
+3. `maxConcurrentWorkflowTaskExecutionSize` should be significantly ≤ `maxConcurrentWorkflowTaskPollers`. And `maxConcurrentActivityExecutionSize` should be significantly ≤ `maxConcurrentActivityTaskPollers`. The number of pollers should always be lower than the number of executors.
+
+## Drawbacks of putting just "large values everywhere"
+
+As with any multithreading system, specifying too large values without monitoring with the SDK and system metrics will lead to constant resources contention/stealing, which decreases the total throughput and latency jitter of the system.
```
Sushisource, line 105: "... and increases latency jitter..." ?

docs/reference/metrics.md
```diff
@@ -51,11 +51,12 @@ Note that `Core` includes the Typescript SDK.
...
 | worker_start                              | A worker was registered/created/started                                                                                                                         | Counter     | Core, Go, Java                              | namespace, task_queue, worker_type                  |
-| worker_task_slots_available               | How many free slots for execution remain, type determined by the `worker_type` label                                                                            | Counter     | Go                                          | namespace, task_queue, worker_type                  |
+| worker_task_slots_available               | How many free slots for execution remain, type determined by the `worker_type` label                                                                            | Gauge       | Go, Java                                    | namespace, task_queue, worker_type                  |
```
Sushisource, line 58: I will go add this to core now

docs/content/how-to-tune-workers.md
Spikhalskiy: I added clarifications in the update, I think that the current order makes the most sense. Every next step makes sense only after checking each of the previous ones.

PR #881, opened 2022-02-10
tctl reference: add pages for namespace, data-converter, and cluster

djmagee:
## What does this PR do?

- Create pages for tctl commands `namespace`, `data-converter`, and `cluster`.
- Alphabetizes sidebar


docs/reference/tctl/cluster/list-search-attributes.md
```diff
@@ -0,0 +1,15 @@
...
+id: list-search-attributes
+title: tctl cluster list-search-attributes
+sidebar_label: list-search-attributes
+description: How to list all Search Attributes using tctl.
+tags:
+  - reference
+  - tctl
+---
+
+The `tctl cluster list-search-attributes` command lists all [Search Attributes](/docs/content/what-is-a-search-attribute) that can be used in the `--query` modifier of the [`tctl workflow list`](/docs/reference/tctl/workflow/list) command.
```
aarohib, line 11:
```suggestion
The `tctl cluster list-search-attributes` command lists all [Search Attributes](/docs/content/what-is-a-search-attribute) that can be used in the 
`--query` modifier of the [`tctl workflow list`](/docs/reference/tctl/workflow/list) command.
```

aarohib: because in the UI, the '--query' breaks across two lines such that `-` ends line one and `-query` starts the next line. may not be a big deal, but at first glance it looks strange.

docs/reference/tctl/namespace/register.md
```diff
@@ -0,0 +1,151 @@
...
+sidebar_label: register
+description: How to register a Namespace using tctl.
+tags:
+  - reference
+  - tctl
+---
+
+The `tctl namespace register` command registers a [Namespace](/docs/content/what-is-a-namespace).
+
+`tctl namespace describe`
```
aarohib, line 13: should this be `tctl namespace register` ?

docs/reference/tctl/namespace/register.md
```diff
@@ -0,0 +1,151 @@
...
+
+```
+tctl namespace register --description <value>
+```
+
+### `--global-namespace`
+
+How to specify whether a [Namespace](/docs/content/what-is-a-namespace) is a Global Namespace.
+
+Alias `--gd`
```
aarohib, line 61: Wouldn't the Alias derive to `--gn` for _g_lobal _n_amespace?

docs/reference/tctl/namespace/register.md
```diff
@@ -0,0 +1,151 @@
...
+
+```
+tctl namespace register --history-uri <uri>
+```
+
+### `--namespace-data`
+
+How to specify data for a [Namespace](/docs/content/what-is-a-namespace) in the form of key-value pairs (such as `k1:v1,k2:v2,k3:v3`).
+
+Alias `--dmd`
```
aarohib, line 97: calling out just to verify that this (purely based on the preceding naming of aliases) might be `--nmd`. Or not.

docs/reference/tctl/cluster/list-search-attributes.md
djmagee: Interesting! Which browser are you using, @aarohib?

docs/reference/tctl/namespace/register.md
djmagee: _D'oh!_ Copypasta fail.

docs/reference/tctl/namespace/register.md
djmagee: Makes sense, but the code sez otherwise: https://github.com/temporalio/tctl/blob/316cdd28256dab4c850c56542efb70e426e78031/cli/flags.go#L149

docs/reference/tctl/namespace/register.md
djmagee: Yep—see https://github.com/temporalio/tctl/blob/316cdd28256dab4c850c56542efb70e426e78031/cli/flags.go#L151

docs/reference/tctl/cluster/list-search-attributes.md
djmagee: Behavior confirmed in Safari and Brave on macOS and Edge on Windows. No repro in Firefox on either macOS or Windows. (Also no repro in Internet Explorer, but who cares?)

When repro'd, line break occurs before `--query`, between the hyphens, immediately following the hyphens, or after `--query`, depending on window width.

Because this is a rendering issue and not a source-file issue, I'm gonna stet rather than hard-coding a line break.

This appears to be consistent behavior (based on my extensive sampling during the past five minutes). Let's check with Candace or Shawn to find out whether they know a way to override this behavior in CSS without preventing desired word breaks.

aarohib (approved):

flossypurse (approved):

PR #882, opened 2022-02-10
Update Worker shutdownGraceTime advice

lorensr:
Thanks @mabadir for pointing out the gap!


sw-yx:
i think Mina wanted something more substantial than just the clarification of shutdownGraceTime.

>  yesterday’s discussion around updating production workers, and how best to handle graceful shutdown needs to be surfaced in the documentation…
> It was not very clear to me what happens when a worker dies in the middle of an activity.
> This also might require some documentation around how to architect your code…I think any activity that takes more than 1 min or 2 should be broken down into smaller activities and managed through a workflow…very long running activities will have the risk of dying and all the progress is lost, unlike workflows where the history can be replayed.

flossypurse (approved):

lorensr:
That seems like general non-sdk-specific Temporal concepts. I added links in TS docs to the concept pages. The Activities concept page has info on long-running Activities. I'm not finding something on architecture like "how much should I put in an Activity versus breaking it up into smaller activities"—examples/advice on what type of logic for workflows to maintain state of versus a heartbeat to maintain state of. @flossypurse do we have content on that?

image

flossypurse:
> I'm not finding something on architecture like "how much should I put in an Activity versus breaking it up into smaller activities"—examples/advice on what type of logic for workflows to maintain state of versus a heartbeat to maintain state of

Ya we want this information to live in concepts - and at the moment those concept descriptions are lacking recommendations around patterns for Activity complexity -

lorensr:
Ok thanks, opened a new issue for that piece: https://github.com/temporalio/documentation/issues/890

PR #886, opened 2022-02-11
[chore] update algolia key to new app, enable faceting

sw-yx:
## What does this PR do?

[chore] update algolia key to new app, enable faceting

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -156,7 +156,7 @@ module.exports = {
       apiKey: "cd527863e60d95ebe650cdd21c7a6f3f",
       indexName: "temporal",
       // contextualSearch: true, // Optional, If you different version of docs etc (v1 and v2) doesn't display dup results
-      appId: 'T5D6KNJCQS', // Optional, if you run the DocSearch crawler on your own
+      appId: "T5D6KNJCQS", // Optional, if you run the DocSearch crawler on your own
       // algoliaOptions: {}, // Optional, if provided by Algolia
     },
   },
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -153,10 +153,10 @@ module.exports = {
       anonymizeIP: true, // Should IPs be anonymized?
     },
     algolia: {
-      apiKey: "14805ba2eb682edb2e719df4d5e03c8a",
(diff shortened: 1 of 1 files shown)


sw-yx:
theoretically there is supposed to be faceting enabled:

image

but i cant get the actual facet options to display when i search and the docs aren't very helpful  :( cc @slorber in case you want feedback on faceted search with algolia

resources used: 
- https://docusaurus.io/docs/search#contextual-search
- https://www.algolia.com/doc/guides/managing-results/refine-results/faceting/how-to/declaring-attributes-for-faceting/?client=javascript
- https://discourse.algolia.com/t/how-to-configure-a-new-facet/10776/9

still the results better than before so recommend merging

slorber:
Not sure to understand what you want me to check? 🤔 

FYI we enabled contextualSearch by default recently, maybe you now want to disable it?

sw-yx:
@slorber we were unable to enable faceted search in docusaurus, even though i enabled facet index in algolia, i have not seen any docs or examples or screenshots that show that docusaurus has faceted search... is that something that comes in the docusaurus search component, or do we have to code it ourselves?

slorber:
@sw-yx all sites with `contextualSearch: true` (now the default in beta 15) are using facetFilters search, officially documented as `searchParameters.facetFilters` on the DocSearch/Algolia docs that we reference from our doc.

image

(there's a typo in that example that I'm fixing in https://github.com/facebook/docusaurus/pull/6704)

The way it works: we merge the contextual facetFilters (dynamic, depends on current page) with the ones you provide in config.

If `contextualSearch: true` does not work for you it's likely hiding a deeper problem. Maybe @shortcuts can help figure it out.

Now, when you say "faceted search", are you talking about filtering (ie `facetFilcers`) or do you have something else in mind?

shortcuts:
Hey

@sw-yx you can go to the edit -> `search preview` and see that only `docusaurus_tag` was retrieved during last crawl (feb 15), which makes `contextualSearch` unusable as it's missing `language`.

We hint this issue in our UI with the following error, (we might need to display it somewhere else too 🤔)
Screenshot 2022-02-17 at 12 13 47

As you've edited the `indexSettings` in the Algolia dashboard, restarting the crawl should be enough, but otherwise deleting the index and starting a new crawl will also solve it.

PR #918, opened 2022-02-17
Subscription Walkthrough in PHP

seregazhuk:


docs/php/subscription-tutorial.md
```diff
@@ -0,0 +1,347 @@
...
+Let's build a realistic monthly subscription payments workflow that can be canceled while it runs. 
+
+:::info Prerequisites
+
+We assume that you have gone through our [Hello World tutorial](/docs/php/hello-world) and understood the basics of 
+getting a Temporal PHP SDK project up and running. We don't assume knowledge of the Workflow APIs.
+
+:::
+
+**To skip straight to a fully working example, you can check our [Subscription Workflow repo](https://github.com/temporalio/samples-php/tree/master/app/src/Subscription)**.
```
sw-yx, line 17: remember to leave a link back on the readme when this is merged

docs/php/subscription-tutorial.md
```diff
@@ -0,0 +1,353 @@
...
+
+```php
+public function subscribe(string $userID)
+{
+    yield $this->account->sendWelcomeEmail($userID);
+
+    try {
+        $isTrialPeriod = true;
+        while (true) {
+            yield Workflow::timer(CarbonInterval::days(30));
```
sw-yx, line 257: a month is not 30 days - can you give some code example of "proper" monthly date setting? eg "1st day of each month"

seregazhuk: Yes, I'll fix it 👍

sw-yx: doesnt have to be in this code it self it is too complicated - could just be a comment on the recommended way to do this

sidebars.js
```diff
@@ -236,7 +236,7 @@ module.exports = {
           type: "category",
           label: "Tutorials",
           collapsed: false,
-          items: ["php/hello-world"],
+          items: ["php/hello-world", "php/subscription-tutorial"],
```
flossypurse, line 239: prolly add this to the learning paths sidebar section under a new category "Subscription"

sw-yx (approved):

sidebars.js
sw-yx: i cut out the Tutorials section just to be in line with your goals and moved the mention to the intro for now so that it is not an orphan page.

PR #924, opened 2022-02-18
[TS] recommend option object args in docs

sw-yx:
## What does this PR do?

intentionally matching up with the go docs, teaching the simple form first, then mentioning that we recommend the struct form https://docs.temporal.io/docs/go/how-to-develop-a-workflow-definition-in-go

### DIFF
--- a/docs/typescript/workflows.md
+++ b/docs/typescript/workflows.md
@@ -41,12 +41,12 @@ The snippet above uses `proxyActivities` to create functions that, when called, 
 A Workflow function can have multiple parameters, but we encourage you to use a single object parameter, as that helps with backward compatibility:
 
 ```ts
-export async function example({
-  name,
-}: {
+type ExampleArgs = {
   name: string;
-}): Promise<{ greeting: string }> {
-  const greeting = await greet(name);
+}
+
+export async function example(args: ExampleArgs): Promise<{ greeting: string }> {
+  const greeting = await greet(args.name);
   return { greeting };
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/typescript/workflows.md
```diff
@@ -37,6 +37,14 @@ Workflow Definitions are "just functions", which can store state, and orchestrat
 <!--SNIPEND-->
 
 The snippet above uses `proxyActivities` to create functions that, when called, schedule a `greet` Activity in the system to say "Hello World".
+Workflow functions can support one or more arguments, but we encourage you to use single parameter option objects, as that helps with backward compatibility:
```
lorensr, line 40:
```suggestion

A Workflow function can have multiple parameters, but we encourage you to use a single object parameter, as that helps with backward compatibility:
```

docs/typescript/workflows.md
```diff
@@ -37,6 +37,14 @@ Workflow Definitions are "just functions", which can store state, and orchestrat
...
+
+```ts
+export async function example({ name: string }): Promise<{ name: string }> {
```
lorensr, line 43:
```suggestion
export async function example({ name: string }): Promise<{ greeting: string }> {
```

docs/typescript/workflows.md
```diff
@@ -38,6 +38,19 @@ Workflow Definitions are "just functions", which can store state, and orchestrat
...
+
+```ts
+export async function example({
+  name,
+}: {
+  name: string;
+}): Promise<{ greeting: string }> {
+  const greeting = await greet(name);
+  return { greeting };
+}
```
sw-yx, line 51: maybe more readable?

```suggestion
type argsType = {
  name: string;
}

export async function example(args: argsType): Promise<{ greeting: string }> {
  const greeting = await greet(args.name);
  return { greeting };
}
```

lorensr:
```suggestion
type ExampleArgs = {
  name: string;
}

export async function example(args: ExampleArgs): Promise<{ greeting: string }> {
  const greeting = await greet(args.name);
  return { greeting };
}
```

flossypurse (approved):

PR #930, opened 2022-02-20
update the production dependencies section with correct info

idoshamun:
## What does this PR do?

After some digging, we found the root cause for my previous issue.
You can find more information in my comment here:
https://github.com/temporalio/sdk-typescript/issues/489#issuecomment-1046246332

### DIFF
--- a/docs/typescript/production-deploy.md
+++ b/docs/typescript/production-deploy.md
@@ -130,7 +130,7 @@ Add the following line to your Dockerfile if you use Debian based images:
 RUN apt update && apt install -y ca-certificates
 ```
 
-## Install all dependencies even in production
+## Missing dependencies in production
 
-Make sure to run `npm i` when you build your Docker image.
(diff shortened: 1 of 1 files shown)


sw-yx:
not sure if this is a specific recommendation we want to give as prebuilding workflows is meant to be a performance optimization. cc @lorensr if this is a dependency that we should pull in then lets do that, its just acorn so i imagine pretty light

lorensr:
Yeah, we don't want to force use of `workflowBundle` or `npm i --production=false` in prod. I updated the issue: https://github.com/temporalio/sdk-typescript/issues/489#issuecomment-1048250793

For now, let's continue with the `npm i` recommendation. I updated `--production=false` and added a link to the issue.

idoshamun:
Ok makes sense. @sw-yx should I go ahead and close this PR then?

sw-yx:
no no, Loren updated the doc for you. @flossypurse can you stamp?

flossypurse (approved):

PR #934, opened 2022-02-22
Add Workflow default Retry Policy

lorensr:
I don't actually know all the default values, so someone please edit this. I just know that by default, failed Workflow Executions don't get retried.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ To ensure that each change is relevant and properly peer reviewed, we request th
 This means that if you are outside the Temporal organization, you must fork the repository and create pull requests from branches on your own fork.
 GitHub's [first-contributions repo README](https://github.com/firstcontributions/first-contributions) provides an example of how to do that.
 
-When it comes to crafting content, please follow our [style guidelines](style-guidance).
+When it comes to crafting content, please follow our [style guidance](#style-guidance).
 
 ## Preview changes locally
 
--- a/docs/concepts/what-is-a-retry-policy.md
+++ b/docs/concepts/what-is-a-retry-policy.md
@@ -43,13 +43,25 @@ There are some subtle nuances to how Events are recorded to an Event History whe
   But the first Workflow Task won't be scheduled until the backoff duration is exhausted.
   That duration is recorded as the `firstWorkflowTaskBackoff` field of the new run's `WorkflowExecutionStartedEventAttributes` event.
 
-### Default values for Retry Policy\*\*
+### Default Values
+
+#### Default Activity Retry Policy
 
 ```
 Initial Interval     = 1 second
 Backoff Coefficient  = 2.0
 Maximum Interval     = 100 × Initial Interval
 Maximum Attempts     = ∞
+Non-Retryable Errors = []
+```
+
+#### Default Workflow Retry Policy
+
+```
+Initial Interval     = 1 second
+Backoff Coefficient  = 2.0
+Maximum Interval     = 100 × Initial Interval
+Maximum Attempts     = 1
 Non-Retryable Errors = []
 ```
 
--- a/docs/concepts/what-is-a-retry-policy.md
+++ b/docs/concepts/what-is-a-retry-policy.md
@@ -43,7 +43,7 @@ There are some subtle nuances to how Events are recorded to an Event History whe
   But the first Workflow Task won't be scheduled until the backoff duration is exhausted.
   That duration is recorded as the `firstWorkflowTaskBackoff` field of the new run's `WorkflowExecutionStartedEventAttributes` event.
 
-### Default Values
+### Default values for Retry Policy
 
 #### Activities Retry Policy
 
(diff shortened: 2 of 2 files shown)


tsurdilo:
weird, temporal provides a default retry policy for activities but not for workflow.

flossypurse:
I don't think this is an accurate route to take - but that could be because my understanding is based on the Go SDK.

But essentially these two statements address what you are trying to do here:

> When a Workflow Execution is spawned it is not associated with a default Retry Policy and thus does not retry by default.
  
> When an Activity Execution is spawned, it is associated with a default Retry Policy, and thus Activity Task Executions are retried by default.

A default Retry Policy is a thing by itself.
It doesn't matter whether it is attached to a Workflow Execution or an Activity Execution.
If you supply an empty Retry Policy without specifying anything in Go for example, the SDK overwrites the default values.
The system will just auto supply a default policy to Activity Executions - but won't for Workflow Executions.
So, the values of the default policy are the same regardless.

So I don't think these are good changes to make unless there is a misunderstanding of the above.

lorensr:
Thanks, I didn't read that part of the page. For others who also aren't reading the page and are just trying to lookup the default, I kept the section for them to find, and fixed the content. I also updated Activity default `MaximumIntervalInSeconds` per @tsurdilo's comment (it used to be `100 × Initial Interval`).

lorensr:
image

flossypurse (changes requested):
I am not saying that we can't improve upon the narrative here.
But I don't think these particular changes improve things - 

The fact is that providing an empty Retry Policy (a custom one but with no values specified) means that the default Retry Policy values are used.

By creating these headers, we are making it about whether the system provides default policies to certain execution types - which is not the point here.
The point about when a policy is automatically applied to an execution type is already made above in the page.

I think that "I didn't read it" is not actionable feedback.

I support a better narrative - but this is not it and it needs to be for a different reason.

lorensr:
> The fact is that providing an empty Retry Policy (a custom one but with no values specified) means that the default Retry Policy values are used.

Just for Activities or also for WEs? If for WEs, what are the default values?

flossypurse:
> > The fact is that providing an empty Retry Policy (a custom one but with no values specified) means that the default Retry Policy values are used.
> 
> Just for Activities or also for WEs? If for WEs, what are the default values?

The default values are the same - it doesn't matter where the policy is applied:
https://pkg.go.dev/go.temporal.io/sdk/internal#RetryPolicy

Like I said - I am all for improving the overall narrative.
But I don't see us making progress on that in this PR in this way.

My suggestion is to start by updating the default values to be correct and current.
Then, if we want to improve the overall narrative - lets draft out our proposed changes separately.

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -43,16 +43,22 @@ There are some subtle nuances to how Events are recorded to an Event History whe
...
-### Default values for Retry Policy\*\*
+### Default Values
+
+#### Activities Retry Policy
 
 ```
 Initial Interval     = 1 second
 Backoff Coefficient  = 2.0
-Maximum Interval     = 100 × Initial Interval
+Maximum Interval     = ∞
```
flossypurse, line 53: +1

CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ To ensure that each change is relevant and properly peer reviewed, we request th
 This means that if you are outside the Temporal organization, you must fork the repository and create pull requests from branches on your own fork.
 GitHub's [first-contributions repo README](https://github.com/firstcontributions/first-contributions) provides an example of how to do that.
 
-When it comes to crafting content, please follow our [style guidelines](style-guidance).
+When it comes to crafting content, please follow our [style guidance](#style-guidance).
```
flossypurse, line 9: +1

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -43,16 +43,22 @@ There are some subtle nuances to how Events are recorded to an Event History whe
   But the first Workflow Task won't be scheduled until the backoff duration is exhausted.
   That duration is recorded as the `firstWorkflowTaskBackoff` field of the new run's `WorkflowExecutionStartedEventAttributes` event.
 
-### Default values for Retry Policy\*\*
+### Default Values
```
flossypurse, line 46: Leave header alone, but remove the trailing and escaped asterisks 
```suggestion
### Default values for Retry Policy
```

flossypurse (commented):
OK, so I added "+1" to the changes you can do.
Plus the header suggestion.

For the other stuff - I am open on working in a draft together and look for a better narrative.

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -43,12 +43,12 @@ There are some subtle nuances to how Events are recorded to an Event History whe
   But the first Workflow Task won't be scheduled until the backoff duration is exhausted.
   That duration is recorded as the `firstWorkflowTaskBackoff` field of the new run's `WorkflowExecutionStartedEventAttributes` event.
 
-### Default values for Retry Policy\*\*
+### Default Values for Retry Policy
```
flossypurse, line 46: Docs style guide uses sentence casing.

lorensr: Fixed thanks!

flossypurse (approved):

PR #938, opened 2022-02-23
[TS Docs] fix patching notes

sw-yx:
## What does this PR do?

we seem to be recommending people rename task queues more than rename workflows, so just want to shout it out here.


sw-yx:
tagging in @lorensr to see what he thinks

flossypurse (approved):

PR #943, opened 2022-02-24
Add missing export and make the parameter to getHandle more clear

stephenreddek:
## What does this PR do?
Fixes a missing export of a signal subscription.
Uses the example of `business-meaningful-id` from previous client example to make the `getHandle` arg more clear.


flossypurse (approved):
Pretty sure this is a positive change - @lorensr to confirm?

lorensr:
Thanks for the fix! 😊

PR #945, opened 2022-02-24
add temporal-rest blog post

vkarpov15:
## What does this PR do?

Blog post introducing https://www.npmjs.com/package/temporal-rest :+1:

### DIFF
--- a/blog/2022-03-01-temporal-rest.md
+++ b/blog/2022-03-01-temporal-rest.md
@@ -21,15 +21,15 @@ In this blog post, I'll show how to use temporal-rest to create a couple of REST
 Counter API
 -----------
 
-The key idea of long-lived Workflows is that Workflow functions are _deterministic_, so Temporal can store the state of the Workflow by storing the Workflow's initial state and event history.
-For example, below is a Workflow that stores a single numeric counter.
-The Workflow listens for a Signal `increment` to increase the counter, and a Query `getCount` to get the current state of the counter.
+The key idea of long-lived Workflows is that Workflow functions are _deterministic_, so Temporal can persist the state of the Workflow by storing the Workflow's initial state and event history.
+For example, below is a Workflow that keeps a single numeric counter.
+The Workflow listens for an `increment` Signal to increase the counter, and a `getCount` Query to get the current value of the counter.
 
 ```ts
 import * as wf from '@temporalio/workflow';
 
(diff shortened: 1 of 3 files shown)


lorensr (approved):
Nice! I got lazy after the first suggested edit and committed the rest in https://github.com/temporalio/documentation/pull/945/commits/6cb53fcab3417da2634777382e0e2948fab20b6c —lmk if you'd like me to undo anything ☺️ 

- Changed `exports.` to `export const` to match `export function` and our docs/samples.
- In `{ok: 1}`, 1 means successful? A little confusing here, appearing next to `{ result: 1 }`, and the signal handler not returning anything. Maybe `{ delivered: true }` or something?
- Removed the Worker line—I think it raises questions, and is perhaps not essential to explain in this post? If essential, I'd explain in more depth.

lorensr:
Also, you make the point about not having to do DB calls at the beginning. Perhaps at the end, it would be helpful to more fully close that loop, noting what automatic durability means—like "even if Temporal Server goes down or the process running the Workflow code loses power, the next query after the servers are up will have the correct counter value", even though we never sent an UPDATE or $inc

blog/2022-03-01-temporal-rest.md
```diff
@@ -0,0 +1,239 @@
...
+author: Valeri Karpov
+author_title: Community Member
+author_image_url: https://avatars.githubusercontent.com/u/1620265?v=4
+release_version: V1.15
+---
+
+<!--truncate-->
+
+You can use Temporal's long-lived Workflows to build [REST APIs without a conventional database](https://docs.temporal.io/blog/build-an-ecommerce-app-with-temporal-part-4-rest-api).
+While you can build a REST API on top of Temporal Workflows yourself, we built [`temporal-rest`](https://www.npmjs.com/package/temporal-rest) to make creating a RESTful API for your Workflows a one-liner with [ExpressJS](https://expressjs.com/).
```
vkarpov15: Maybe change "we" to "I" here to avoid implying that this is an official Temporal project?

sw-yx:
```suggestion
While you can build a REST API on top of Temporal Workflows yourself, I built [`temporal-rest`](https://www.npmjs.com/package/temporal-rest) to make creating a RESTful API for your Workflows a one-liner with [ExpressJS](https://expressjs.com/).
```

PR #952, opened 2022-02-28
clarify DST transition for cron

sw-yx:
## What does this PR do?

I find our warnings around DST very scary. two examples help greatly reduce the anxiety around using the new feature.


flossypurse:
👍 
@djmagee for a quick copy edit

docs/concepts/what-is-a-temporal-cron-job.md
```diff
@@ -90,9 +90,11 @@ Consider that using time zones in production introduces a surprising amount of c
 If you need to use time zones, here are a few edge cases to keep in mind:
 
 - **Beware Daylight Savings**: If a Temporal Cron Job is scheduled around the time when daylight saving time begins or ends, e.g. `30 2 * * *`, **it may run zero, one, or two times in a day**! The Cron library that we use does not do any special handling of DST transitions. Avoid schedules that include times that fall within DST transition periods.
```
djmagee, line 92:
```suggestion
- **Beware Daylight Saving Time**: If a Temporal Cron Job is scheduled around the time when daylight saving time (DST) begins or ends (for example, `30 2 * * *`), **it might run zero, one, or two times in a day**! The Cron library that we use does not do any special handling of DST transitions. Avoid schedules that include times that fall within DST transition periods.
```

docs/concepts/what-is-a-temporal-cron-job.md
```diff
@@ -90,9 +90,11 @@ Consider that using time zones in production introduces a surprising amount of c
...
+  - For example, in the US, DST is at the 1AM hour. When you "fall back", the clock will go `1:59 … 1:00 … 1:01 … 1:59 … 2:00 … 2:01 AM` and any Cron jobs that falls in that 1AM hour will be fired again. The inverse happens when clocks "spring forward" for DST, and Cron jobs that fall in the 1AM hour are skipped.
```
djmagee, line 93:
```suggestion
  - For example, in the US, DST begins at 2 AM. When you "fall back," the clock goes `1:59 … 1:00 … 1:01 … 1:59 … 2:00 … 2:01 AM` and any Cron jobs that fall in that 1 AM hour are fired again. The inverse happens when clocks "spring forward" for DST, and Cron jobs that fall in the 2 AM hour are skipped.
```

docs/concepts/what-is-a-temporal-cron-job.md
```diff
@@ -90,9 +90,11 @@ Consider that using time zones in production introduces a surprising amount of c
...
+  - In other timezones like Chile and Iran, DST "spring forward" is at midnight. [11:59PM is followed by 1:00AM, which means `00:00:00` never happens](https://twitter.com/raphaelschaad/status/1425249265643839490?s=20&t=vrMx4ixqcBRp5D7xB4FgNw).
```
djmagee, line 94:
```suggestion
  - In other time zones like Chile and Iran, DST "spring forward" is at midnight. [11:59 PM is followed by 1 AM, which means `00:00:00` never happens](https://twitter.com/raphaelschaad/status/1425249265643839490?s=20&t=vrMx4ixqcBRp5D7xB4FgNw).
```

docs/concepts/what-is-a-temporal-cron-job.md
```diff
@@ -90,9 +90,11 @@ Consider that using time zones in production introduces a surprising amount of c
...
+- **Absolute Time Fixed at Start**: The absolute start time of the next Run is computed and stored in the database when the previous Run completes, and is not recomputed. This means that if you have a Cron Schedule that runs very infrequently, and the definition of the time zone changes in between one Run and the next, the Run might happen at the wrong time. For example, `CRON_TZ=America/Los_Angeles 0 12 11 11 *` means "noon in Los Angeles on November 11" (normally not in DST). If at some point the government makes any changes (e.g. move the end of DST one week later, or stay on permanent DST year-round) the meaning of that specification will change. In that first year, the Run will happen at the wrong time, because it was computed using the older definition.
```
djmagee, line 97:
```suggestion
- **Absolute Time Fixed at Start**: The absolute start time of the next Run is computed and stored in the database when the previous Run completes, and is not recomputed. This means that if you have a Cron Schedule that runs very infrequently, and the definition of the time zone changes between one Run and the next, the Run might happen at the wrong time. For example, `CRON_TZ=America/Los_Angeles 0 12 11 11 *` means "noon in Los Angeles on November 11" (normally not in DST). If at some point the government makes any changes (for example, move the end of DST one week later, or stay on permanent DST year-round), the meaning of that specification changes. In that first year, the Run happens at the wrong time, because it was computed using the older definition.
```

djmagee (approved):

PR #953, opened 2022-02-28
Clarify that all providers use Tally's interface

zaninime:
When reading the sentence, it sounded like only M3 relies on Tally, while the reality is that all the providers use it (as far as I understand the code).

## What does this PR do?
Clarify a sentence, to help other devs navigate Temporal's code faster.


flossypurse (approved):
LGTM - @sw-yx ?

PR #960, opened 2022-03-03
Give structure to what-is-a-retry-policy

lorensr:
> Retry Policies do not apply to Workflow Task Executions, which, by default, retry indefinitely.

I removed `by default`. Is it possible to change that behavior? If no, I'd leave out. Otherwise, I'd say how.


docs/concepts/what-is-a-retry-policy.md
```diff
@@ -7,43 +7,26 @@ tags:
...
 
-A Retry Policy is collection of attributes that instructs the Temporal Server how to retry a failure of a [Workflow Execution](/docs/concepts/what-is-a-workflow-execution) or an [Activity Task Execution](/docs/concepts/what-is-an-activity-task-execution).
-
-- If a custom Retry Policy is to be used, it must be provided as an options parameter when a Workflow Execution or an Activity Execution is spawned.
-
-- The wait time before a retry is the _retry interval_.
-  A retry interval is the smaller of two values:
-  - The [Initial Interval](#initial-interval) multiplied by the [Backoff Coefficient](#backoff-coefficient) raised to the power of the number of retries.
-  - The [Maximum Interval](#maximum-interval).
+A Retry Policy is collection of attributes that instructs the Temporal Server how to retry a failure of a [Workflow Execution](/docs/concepts/what-is-a-workflow-execution) or an [Activity Task Execution](/docs/concepts/what-is-an-activity-task-execution). Retry Policies do not apply to [Workflow Task Executions](/docs/concepts/what-is-a-workflow-task-execution), which retry indefinitely.
```
djmagee, line 10:
```suggestion
A Retry Policy is a collection of attributes that instructs the Temporal Server how to retry a failure of a [Workflow Execution](/docs/concepts/what-is-a-workflow-execution) or an [Activity Task Execution](/docs/concepts/what-is-an-activity-task-execution).
Retry Policies do not apply to [Workflow Task Executions](/docs/concepts/what-is-a-workflow-task-execution), which retry indefinitely.
```

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -7,43 +7,26 @@ tags:
...
 
 <!-- ![Diagram that shows the retry interval and its formula](/img/retry-interval-diagram.png) -->
 
-- When a Workflow Execution is spawned it is not associated with a default Retry Policy and thus does not retry by default.
-  The intention is that a Workflow Definition should be written to never fail due to intermittent issues; an Activity is designed to handle such issues.
+### Default behavior
 
-Retry Policies do not apply to [Workflow Task Executions](/docs/concepts/what-is-a-workflow-task-execution), which, by default, retry indefinitely.
+- **Workflow Execution**: When a Workflow Execution is spawned, it is not associated with a default Retry Policy and thus does not retry by default. The intention is that a Workflow Definition should be written to never fail due to intermittent issues; an Activity is designed to handle such issues.
```
djmagee, line 16:
```suggestion
- **Workflow Execution**: When a Workflow Execution is spawned, it is not associated with a default Retry Policy and thus does not retry by default.
The intention is that a Workflow Definition should be written to never fail due to intermittent issues; an Activity is designed to handle such issues.
```

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -7,43 +7,26 @@ tags:
...
 
-- A Retry Policy can be provided to a Workflow Execution when it is spawned, but only certain scenarios merit doing this, such as the following:
+- **Activity Execution**: When an Activity Execution is spawned, it is associated with a default Retry Policy, and thus Activity Task Executions are retried by default. When an Activity Task Execution is retried, the Cluster places a new [Activity Task](/docs/concepts/what-is-an-activity-task) into its respective [Activity Task Queue](/docs/concepts/what-is-a-task-queue), which results in a new Activity Task Execution.
```
djmagee, line 18:
```suggestion
- **Activity Execution**: When an Activity Execution is spawned, it is associated with a default Retry Policy, and thus Activity Task Executions are retried by default.
When an Activity Task Execution is retried, the Cluster places a new [Activity Task](/docs/concepts/what-is-an-activity-task) into its respective [Activity Task Queue](/docs/concepts/what-is-a-task-queue), which results in a new Activity Task Execution.
```

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -7,43 +7,26 @@ tags:
...
 
-  - A [Temporal Cron Job](/docs/concepts/what-is-a-temporal-cron-job) or some other stateless, always-running Workflow Execution that can benefit from retries.
-  - A file-processing or media-encoding Workflow Execution that downloads files to a host.
+### Custom Retry Policy
 
-- When an [Activity Execution](/docs/concepts/what-is-an-activity-execution) is spawned, it is associated with a default Retry Policy, and thus Activity Task Executions are retried by default.
-  When an Activity Task Execution is retried, the Cluster places a new [Activity Task](/docs/concepts/what-is-an-activity-task) into its respective [Activity Task Queue](/docs/concepts/what-is-a-task-queue), which results in a new Activity Task Execution.
-
-There are some subtle nuances to how Events are recorded to an Event History when a Retry Policy comes into play.
+To use a custom Retry Policy, provide it as an options parameter when starting a Workflow Execution or Activity Execution. Only certain scenarios merit starting a Workflow Execution with a custom Retry Policy, such as:
```
djmagee, line 22:
```suggestion
To use a custom Retry Policy, provide it as an options parameter when starting a Workflow Execution or Activity Execution.
Only certain scenarios merit starting a Workflow Execution with a custom Retry Policy, such as the following:
```

djmagee (commented):
Looks great; thanks!

@flossypurse, what's your opinion on placing Retry Interval and Event History after Properties?

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -85,14 +68,35 @@ Non-Retryable Errors = []
...
   In this case you can specify them such that if they occur, the given execution will not be retried.
 
+### Retry interval
+
+The wait time before a retry is the _retry interval_. A retry interval is the smaller of two values:
+
+- The [Initial Interval](#initial-interval) multiplied by the [Backoff Coefficient](#backoff-coefficient) raised to the power of the number of retries.
+- The [Maximum Interval](#maximum-interval).
+
+### Event History
```
flossypurse, line 86:
```suggestion
### Retry Policies and Event Histories
```

docs/concepts/what-is-a-retry-policy.md
```diff
@@ -85,14 +68,35 @@ Non-Retryable Errors = []
...
+
+- For a Workflow Execution with a Retry Policy, if the Workflow Execution fails, the Workflow Execution will [Continue-As-New](/docs/concepts/what-is-continue-as-new) and the associated Event is written to the Event History.
+  The [WorkflowExecutionContinuedAsNew](/docs/concepts/what-is-an-event#workflowexecutioncontinuedasnew) Event will have an "initiator" field that will specify the Retry Policy as the value and the new Run Id for the next retry attempt.
+  The new Workflow Execution is created immediately.
+  But the first Workflow Task won't be scheduled until the backoff duration is exhausted.
+  That duration is recorded as the `firstWorkflowTaskBackoff` field of the new run's `WorkflowExecutionStartedEventAttributes` event.
+
 **Implementation guides:**
 
 - [How to set a Retry Policy for a Workflow Execution in Go](/docs/go/how-to-set-startworkflowoptions-in-go/#retrypolicy)
```
flossypurse, line 102: I think we should move links to implementation guides to just after the intro paragraph.

flossypurse (commented):
Nice, this improves the narrative IMO.

docs/concepts/what-is-a-retry-policy.md
lorensr: The hard line break seems so weird to me. I imagine this is a standard thing we do? Why? 😄

docs/concepts/what-is-a-retry-policy.md
lorensr: I'm curious why the repetition here and in `Default values for Retry Policy`. My inclination is toward concision (`Default values` and `Event Histories`), with the idea that the reader knows it's in the context of the Retry Policy page they're reading.

flossypurse: For search results.
Event History is a term that could be involved in multiple headers throughout the docs.
I was thinking that this is a great way to differentiate quickly from the results.

lorensr: It looks like if it were just `Default values`, it would still come up in search, since "retry" matches "What is a Retry Policy"

image

flossypurse: If there is an opportunity to be more specific - I think we should take it.
There is only something to be gained here I should think? Nothing is lost.

docs/concepts/what-is-a-retry-policy.md
flossypurse: We adopted this awhile ago.
Each sentence in a Markdown file is on its own line.
It reduces wrapping while editing, and enables the scope of a change to be just the sentence rather than a whole paragraph.

flossypurse:
> Looks great; thanks!
> 
> @flossypurse, what's your opinion on placing Retry Interval and Event History after Properties?

In the end, I think the properties are the more important.

docs/concepts/what-is-a-retry-policy.md
lorensr: > If there is an opportunity to be more specific - I think we should take it.

I'd consider 

`Properties > Default values for Retry Policy` header on the Retry Policy page 

to be no more specific than 

`Properties > Default values` header on the Retry Policy page.

In both cases, it's clear to me the answer to "which default values?" is "retry policy properties".

> There is only something to be gained here I should think? Nothing is lost.

I wonder whether "for Retry Policy" would fit under @djmagee 's definition of "filler language"?

Regardless, there's something lost for me, and I imagine some subset of others. My experience of the "for Retry Policy" version is:

- Takes longer to parse/read/understand, or to come back later and glance at and understand, compared to the more concise version below.
- Takes me out of flow a bit—my brain hiccups and is like, "wait, so if this thing is about retry policy, what is Properties and Initial interval about?" Of course the answer is obvious, but my brain for a moment automatically imagines there must be a reason for the qualifier prepositional phrase—that other things around it are less specific, or apply to something else.


image

image

flossypurse: Ya, I think you definitely made an air tight case here - well done.

flossypurse (approved):

docs/concepts/what-is-a-retry-policy.md
lorensr: Thanks! ☺️

PR #999, opened 2022-03-16
add intro section on learning paths

sw-yx:
## What does this PR do?

add intro section on learning paths

### DIFF
--- a/docs/learning-paths/index.md
+++ b/docs/learning-paths/index.md
@@ -3,12 +3,6 @@ id: index
 title: Learning
 ---
 
-import ReactPlayer from "react-player"
-
-## What are Temporal Learning Paths?
-
-<ResponsivePlayer url='https://www.youtube.com/watch?v=-YYtRwArSQM' />
-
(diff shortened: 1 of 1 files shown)


docs/learning-paths/index.md
```diff
@@ -3,8 +3,16 @@ id: index
 title: Learning
 ---
 
+import ReactPlayer from "react-player"
+
+## What are Temporal Learning Paths?
+
+<ResponsivePlayer url='https://www.youtube.com/watch?v=-YYtRwArSQM' />
```
flossypurse, line 10: This is great context to provide to someone if they desire it.

But I am not sure about embedding it directly here.
At most, let us link to it - maybe without the header.

Long term, I have noticed that "learning paths" is a great internal name for what we are doing but it is not great for external usage. Externally - I think people will respond a lot better to "tutorials" and we won't need to explain what they are.

flossypurse (approved):

rachfop (approved):

PR #1022, opened 2022-03-28
adding commands directly to docs

flossypurse:
The commands were previously written into docs using Snipsync, but the source file was removed.

### DIFF
--- a/docs/clusters/how-to-integrate-elasticsearch-into-a-temporal-cluster.md
+++ b/docs/clusters/how-to-integrate-elasticsearch-into-a-temporal-cluster.md
@@ -68,7 +68,7 @@ persistence:
 Run the following commands to create the index schema and index:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/clusters/how-to-integrate-elasticsearch-into-a-temporal-cluster.md
```diff
@@ -67,8 +67,17 @@ persistence:
 
 Run the following commands to create the index schema and index:
 
-<!--SNIPSTART setup-es-template-commands-->
-<!--SNIPEND-->
+```bash
+# ES_SERVER is the URL of Elasticsearch server i.e. "http://localhost:9200".
```
djmagee, line 71:
```suggestion
# ES_SERVER is the URL of Elasticsearch server; for example, "http://localhost:9200".
```

PR #1029, opened 2022-03-29
Use normal svg files for logos, fixes #1028

### LINKED ISSUE #1028: [ Bug report ] Blurry logo
## Required

**What is the issue?**

Logo doesn't look crisp on my display:

image

I'm on Lg Ultra Hd display on 2160 x 3840

Looks crisp. 

## Solution

Looks like it's a PNG inside an SVG. Exporting as pure SVG from illustrator/etc should fix it (can also run through `svgo` to reduce file size). If that's not sufficient, further possibilities are [designing on pixel boundaries](https://stackoverflow.com/a/44799741/627729) and [`shape-rendering="crispEdges"`](https://stackoverflow.com/a/53794471/627729).

image

lorensr:
Got SVGs without padding from @canvanooo and ran through svgo

Fixes #1028

image


flossypurse:
Hmm... now I get a crinkly effect

Screen Shot 2022-03-29 at 3 41 22 PM

djmagee:
> Hmm... now I get a crinkly effect

Ditto. When I zoom my browser, the logo sharpens nicely, but at 100% the scaling doesn't do the logo any favors. OTOH, the difference between this and the current version is negligible.

lorensr:
Well `shape-rendering="crispEdges"` definitely doesn't help 😄

image

Maybe this is a little better? `shape-rendering="geometricPrecision"`:

image

@canvanooo any ideas on how to smooth our an svg `path`? (think it happens on non-retina displays at full resolution)

lorensr:
Looks like inline svg has better antialiasing, but not supported by docusaurus.

image


https://docusaurus.io/docs/api/themes/configuration#navbar-logo

lorensr:
@GiantRobots Candace said you solved this for temporal.io?

GiantRobots:
> Checking back with @GiantRobots to see what was done for temporal.io logo?

Currently we're using the SVG with the crinkly effect. We have a PR up for using a PNG instead which addresses the crinkliness but it will deform if a user zooms the browser in with CMD + +.

We'll likely throw that in for the marketing website because we expect most users not to be zoomed in and it still looks good enough when it is deformed from zooming.

Unfortunately the aliasing artifacts are hard to deal with and the curvature of the line and the size of the stroke at that size will always have aliasing artifacts and the browser doesn't provide good resources to draw a better line there.

lorensr:
To me, svg with `shape-rendering="geometricPrecision"` (current version of this PR, [preview here](https://deploy-preview-1029--mystifying-fermi-1bc096.netlify.app/)) looks better on both 1x and retina screens than png:

## png 1x

image

## svg 1x

image

## png retina

image

## svg retina

image

GiantRobots:
Yeah I take it back that's not a good idea especially after testing with other things. I did notice if you change the height of the SVG on the docs site it looks significantly better on non retina screens.

lorensr:
Here is png set to [35px height](https://github.com/temporalio/documentation/issues/1028#issuecomment-1104188702). I'm still preferring svg, particularly for the Temporal text.

## png 35px 1x

image

## png 35px retina

image

flossypurse:
@GiantRobots committed the 35px change in this PR where we are rolling forward to a non-swizzled world https://github.com/temporalio/documentation/pull/1082

flossypurse:
Hey @lorensr - Thank you very much for pushing on this 
Looks like the 35px change that @GiantRobots introduced in this PR has worked well enough: https://github.com/temporalio/documentation/pull/1082

Closing as this now appears resolved.

PR #1030, opened 2022-03-29
Remove - (NumpadSubtract) from recommended range

lorensr:
Previous recommendation was this character, which I imagine was unintentional: https://keycode.info/for/-


djmagee:
In this case, the selection of the hyphen was intentional. 

The Google developer documentation style guide recommends using a hyphen and _not_ using an en dash in a range of numbers (https://developers.google.com/style/hyphens#range-of-numbers). 

The Microsoft Writing Style Guide recommends using an en dash (https://docs.microsoft.com/en-us/style-guide/punctuation/dashes-hyphens/enes). 

Our style guide accepts either a hyphen or an en dash.

PR #1036, opened 2022-04-02
(easy) go mod => go mod tidy?

youngrrr:
I'm pretty new to go but when I run 'go mod' as part of the tutorial I get a generic printout since I'm missing a command. Does the tutorial mean to say that 'go mod tidy' should be run?

## What does this PR do?

Updates 'go mod' in the tutorial to 'go mod tidy'


flossypurse (approved):
Indeed. Thanks!

PR #1039, opened 2022-04-03
Add data converter docs

lorensr:
## What does this PR do?

- Expands on existing data converter what-is
- Adds `typescript/data-converter.md`

## Notes to reviewers

@robholland feel free to edit to fix or add anything. I think the main things missing are:

- [`tctl dc web`](https://docs.temporal.io/docs/tctl/data-converter/web): how to create a websocket URL
- [`tctl --data_converter_plugin`](https://docs.temporal.io/docs/tctl/#--data-converter-plugin): how to use
- remote DCs 

though they don't need to be part of this PR.

@rachfop here's my attempt at:
- including all generally-applicable info outside SDK docs
- in SDK docs, referencing general info instead of re-explaining


lorensr:
This has h2 and h3s—is it possible to add h4s?

image

flossypurse:
> The right-side ToC has h2 and h3s—is it possible to add h4s?

Looks like it only supports h2s and h3s https://docusaurus.io/docs/markdown-features/inline-toc

We would have to do an inline TOC to list h4s looks like -

docs/typescript/data-converters.md
```diff
@@ -0,0 +1,208 @@
+---
+id: data-converters
+title: Data Converters
```
flossypurse, line 3: Just a few small changes and this is close to "How to use a custom Data Converter in TypeScript"

rachfop (approved):

PR #1042, opened 2022-04-05
Reference the correct variable name

stephenreddek:
## What does this PR do?
References `client` instead of undefined `workflow`. This clears up that it is not a function defined on the `@temporalio/workflow` package but on the client.


flossypurse (approved):

flossypurse:
Thank you @stephenreddek !

Issue #748, opened 2021-11-16
[Bug] Repository's autoformat action fails to checkout forked repo branches


JoshuaKGoldberg:
### To Reproduce

https://github.com/temporalio/documentation/pull/747 -> https://github.com/temporalio/documentation/runs/4219627104?check_suite_focus=true

### Expected behavior

I'd like to prove that my code is Prettier-approved.

### Screenshots/Terminal output

```
Fetching the repository
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/typescript-typos*:refs/remotes/origin/typescript-typos* +refs/tags/typescript-typos*:refs/tags/typescript-typos*
  The process '/usr/bin/git' failed with exit code 1
```

### Additional context

Looks like https://github.com/actions/checkout/issues/455#issuecomment-792228083 might be a good fix?

flossypurse:
> Looks like [actions/checkout#455 (comment)](https://github.com/actions/checkout/issues/455#issuecomment-792228083) might be a good fix?

The example that he proposes, as I read it, suggests that it does NOT work. He says:

> Not working: Commit to the fork's branch.

And that is what we need, I think - we need it to work when there is a commit to a branch on the fork.
Perhaps I am not looking at the right spot on that thread though?

PR #1049, opened 2022-04-11
Await the results of the promise so it can be caught

stephenreddek:
## What does this PR do?
`await` the promise inside the function so that the `try`/`catch` can handle a thrown cancellation error.


lorensr (approved):
Thanks for the fix! ☺️

rachfop (approved):

lorensr:
Merging despite prettier, which is expected to fail on forks (#748)

PR #1050, opened 2022-04-12
use `randomUUID()` instead of `Math.random()` for generating workflow ids

vkarpov15:
Re: temporalio/samples-typescript#110

## What does this PR do?

A couple of examples in this blog post use `Math.random()` to generate workflow ids. This is bad practice, we don't want people using `Math.random()` to generate workflow ids in production. This PR uses [Node's built-in `randomUUID()` function](https://nodejs.org/api/crypto.html#cryptorandomuuidoptions) instead to provide unique workflow ids without additional dependencies.

### DIFF
--- a/blog/2022-02-28-temporal-rest.md
+++ b/blog/2022-02-28-temporal-rest.md
@@ -162,7 +162,7 @@ To create a new counter with this Workflow, we would normally send an `increment
 ```ts
 const handle = await client.start(counterWorkflow, {
   taskQueue: 'tutorial',
-  workflowId: 'workflow-' + crypto.randomUUID(),
+  workflowId: 'counter-' + crypto.randomUUID(),
 });
 
 // Increment a new counter
--- a/blog/2022-02-28-temporal-rest.md
+++ b/blog/2022-02-28-temporal-rest.md
@@ -55,7 +55,8 @@ async function run() {
 
   const handle = await client.start(counterWorkflow, {
     taskQueue: 'tutorial',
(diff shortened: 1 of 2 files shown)


blog/2022-02-28-temporal-rest.md
```diff
@@ -48,13 +48,14 @@ To execute an instance of `counterWorkflow`, use `WorkflowClient` to start the W
...
 import { counterWorkflow } from './workflows';
+import crypto from 'crypto';
 
 async function run() {
   const client = new WorkflowClient();
 
   const handle = await client.start(counterWorkflow, {
     taskQueue: 'tutorial',
-    workflowId: 'wf-id-' + Math.floor(Math.random() * 1000),
+    workflowId: 'workflow-' + crypto.randomUUID(),
```
mfateev, line 58: I think the workflow id should come from some business id. What is the valid business id in this case?

lorensr: There isn't an existing business ID. This use case is just "start a run of the counter workflow from a command-line script for example purposes". Perhaps:

```suggestion
    // Usually, we use a business ID. In this example, we don't have one, so we generate an ID.
    workflowId: 'counter-' + crypto.randomUUID(),
```

lorensr (approved):
Example IDs can be updated to `counter-*`

image

image

vkarpov15 (commented):
@lorensr thanks for your suggestions :+1:

djmagee (approved):

PR #1056, opened 2022-04-14
Make search attribute Types section

lorensr:
## What does this PR do?

Better surface SA types to address:

> the page here https://docs.temporal.io/docs/go/search-apis/#value-types lists the search-attribute types that temporal support. But I believe its missing information on how the type `text` is different from `keyword`. I have seen a few customers asking about this exact difference.


rachfop (approved):

flossypurse (approved):
Ah, very nice!

PR #1057, opened 2022-04-14
Use team for CODEOWNERS; add loren to TS

lorensr:
- Use team for CODEOWNERS, so that we don't need to update both GH team and the file
- Adding myself to TS docs, so that I get notified about relevant PRs


flossypurse (approved):
dig it.

PR #1069, opened 2022-04-19
added note about safe pointers being permitted.

DJSanti:
Added a couple of notes about Temporal allowing safe pointers alongside regular type variables. Could use a look-over.


aarohib (approved):
Good stuff!

flossypurse (approved):

PR #1070, opened 2022-04-19
Added notes to Go and TS docs about mutating workflows

DJSanti:
Added a note to the Go and TS Workflow docs about not including Side Effects and command generation, and why that would be a bad idea.

### DIFF
--- a/docs/go/how-to-handle-a-query-in-a-workflow-in-go.md
+++ b/docs/go/how-to-handle-a-query-in-a-workflow-in-go.md
@@ -15,7 +15,9 @@ The handler must be a function that returns two values:
 1. A serializable result
 2. An error
 
-The handler function can receive any number of input parameters, but all input parameters must be serializable.
+Do not include Side Effects or command generation within a query handler. This could mutate the Workflow state and lead to unexpected behaviors on subsequent runs.
+
+The handler function can receive any number of input parameters, but all input parameters must be serializable. 
 The following sample code sets up a Query Handler that handles the `current_state` Query type:
 
 ```go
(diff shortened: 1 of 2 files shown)


docs/go/how-to-handle-a-query-in-a-workflow-in-go.md
```diff
@@ -15,6 +15,8 @@ The handler must be a function that returns two values:
 1. A serializable result
 2. An error
 
+Do not include Side Effects or command generation within a query handler. This could mutate the Workflow state and lead to unexpected behaviors on subsequent runs.
```
flossypurse, line 18:
```suggestion
Do not include any logic that causes Command generation within a Query handler, such as Side Effects or local Activities.
This will lead to unexpected behavior.
```

Just a small stylistic note - put each sentence on a newline.
When someone makes a change, the change will be limited to just that sentence.

docs/typescript/workflows.md
```diff
@@ -429,7 +429,7 @@ Temporal guarantees read-after-write consistency of Signals-followed-by-Queries.
 
 #### Notes on Queries
 
-> 🚨 WARNING: NEVER mutate Workflow state inside a query! This would be a source of non-determinism.
+> 🚨 WARNING: NEVER mutate Workflow state inside a query! Generating commands in query handlers can lead to unexpected behaviors on subsequent executions.
```
flossypurse, line 432:
```suggestion
> 🚨 WARNING: NEVER mutate Workflow state inside a query! Generating Commands in Query handlers can lead to unexpected behavior.
```

flossypurse (commented):
I suggest dropping the "on subsequent runs" part from the warning and just say it "will lead to unexpected behavior" - 

If we wanted to be precise, we might say that 
"If Commands are generated in a Query handler, unexpected behavior is possible whenever the Workflow Function is re-executed after having handled the Query"

flossypurse (approved):

PR #1103, opened 2022-04-28
avoid hard coded color styles in retry simulator for better dark mode support

vkarpov15:
## What does this PR do?

Someone pointed out in Slack that the retry simulator had some issues in dark mode. Before:

image

After this PR:

image


rylandg (approved):
Tyvm

PR #1104, opened 2022-04-30
ensure retry simulator converts slider inputs to numbers, add (in ms) to clarify units

vkarpov15:
Re: feedback from community Slack

## What does this PR do? 

Based on feedback from Slack. One person pointed out a bug where the retry input sliders change runtime to a string, which breaks the runtime calculation. Also, I added (in ms) to headers to clarify that all time inputs are in milliseconds.


djmagee:
Looks clean to me. Only one style nit: the lack of contrast in dark mode in the Success/Failure box.

image

rylandg (approved):
LGTM. Didn't review every single line but the overall change.

flossypurse (approved):

PR #1127, opened 2022-05-07
Fix signal example code snippet

dystopiandev:


lorensr (approved):
Thanks for the fix! 😊

PR #1136, opened 2022-05-10
Move root from /docs to /

flossypurse:


DJSanti (commented):
screw it here's my review

PR #1176, opened 2022-05-20
added definitions

DJSanti:
Search Attribute definitions. Could really use a review from an eng.

### DIFF
--- a/docs/concepts/what-is-a-search-attribute.md
+++ b/docs/concepts/what-is-a-search-attribute.md
@@ -23,7 +23,7 @@ These Search Attributes are created when the initial index is created.
 
 | NAME                  | TYPE     | DEFINITION                                                |
 | --------------------- | -------- | --------------------------------------------------------- |
-| BatcherNamespace      | Keyword  | Batcher namespace.                                        |
+| BatcherNamespace      | Keyword  | Batcher Namespace.                                        |
 | BatcherUser           | Keyword  | Batcher username.                                         |
 | BinaryChecksums       | Keyword  | Binary Id of a given Worker.                              |
 | CloseTime             | Datetime | The time at which the Workflow Execution closes.          |
--- a/docs/concepts/what-is-a-search-attribute.md
+++ b/docs/concepts/what-is-a-search-attribute.md
@@ -23,21 +23,21 @@ These Search Attributes are created when the initial index is created.
 
 | NAME                  | TYPE     | DEFINITION                                                |
 | --------------------- | -------- | --------------------------------------------------------- |
-| BatcherNamespace      | Keyword  | Batcher Namespace.                                        |
-| BatcherUser           | Keyword  | Batcher username.                                         |
-| BinaryChecksums       | Keyword  | Binary Id of a given Worker.                              |
-| CloseTime             | Datetime | The time at which the Workflow Execution closes.          |
-| ExecutionDuration     | Int      | The time needed to run the Workflow Execution.            |
-| ExecutionStatus       | Keyword  | The current state of the Execution.                       |
-| ExecutionTime         | Datetime | The time spent running the Execution.                     |
-| HistoryLength         | Int      | Length of the History Host.                               |
-| RunId                 | Keyword  | Identifies the current Run.                               |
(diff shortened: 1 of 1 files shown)


docs/concepts/what-is-a-search-attribute.md
```diff
@@ -21,23 +21,23 @@ When using [Continue-As-New](/docs/concepts/what-is-continue-as-new) or a [Tempo
...
-| RunId                 | Keyword  |
-| StartTime             | Datetime |
-| StateTransitionCount  | Int      |
-| TaskQueue             | Keyword  |
-| TemporalChangeVersion | Keyword  |
-| WorkflowId            | Keyword  |
-| WorkflowType          | Keyword  |
+| NAME                  | TYPE     | DEFINITION                                                |
+| --------------------- | -------- | --------------------------------------------------------- |
+| BatcherNamespace      | Keyword  | Batcher namespace.                                        |
```
flossypurse, line 26:
```suggestion
| BatcherNamespace      | Keyword  | Batcher Namespace.                                        |
```

flossypurse (commented):
Right now, the additional descriptions do not add a lot more info to the situation. 

And when it comes to tables in the docs, I am not a fan of having long descriptions in table columns.

We may want to break out each of these into their own little header like we do with other references.

flossypurse (approved):
@alexshtin Can you verify for us?

docs/concepts/what-is-a-search-attribute.md
```diff
@@ -21,23 +21,23 @@ When using [Continue-As-New](/docs/concepts/what-is-continue-as-new) or a [Tempo
...
+| ExecutionStatus       | Keyword  | The current state of the Execution.                       |
+| ExecutionTime         | Datetime | The time spent running the Execution.                     |
+| HistoryLength         | Int      | Length of the History Host.                               |
+| RunId                 | Keyword  | Identifies the current Run.                               |
+| StartTime             | Datetime | The time at which the Workflow Execution began running.   |
+| StateTransitionCount  | Int      | The number of times that a stage transition has occurred. |
+| TaskQueue             | Keyword  | Task Queue in use.                                        |
+| TemporalChangeVersion | Keyword  | Current version of Temporal.                              |
+| WorkflowId            | Keyword  | Identifies the Workflow.                                  |
+| WorkflowType          | Keyword  | The type of Workflow currently running.                   |
```
alexshtin, line 40: Let's order it not alphabetically but in order of usefulness.

```suggestion
| NAME                  | TYPE     | DEFINITION                                                |
| --------------------- | -------- | --------------------------------------------------------- |
| WorkflowType          | Keyword  | The type of Workflow.                   |
| WorkflowId            | Keyword  | Identifies the Workflow Execution.                                  |
| ExecutionStatus       | Keyword  | The current state of the Workflow Execution.                       |
| StartTime             | Datetime | The time at which the Workflow Execution started.   |
| CloseTime             | Datetime | The time at which the Workflow Execution completed.          |
| ExecutionTime         | Datetime | Same as StartTime for the most cases but different for cron Workflows and retried Workflows. For them it is the time at which the Workflow Execution actually begin running.                     |
| RunId                 | Keyword  | Identifies the current Workflow Execution Run.                               |
| ExecutionDuration     | Int      | The time needed to run the Workflow Execution. Available only for closed Workflows.            |
| HistoryLength         | Int      | The number of events in the history of Workflow Execution. Available only for closed Workflows.                               |
| StateTransitionCount  | Int      | The number of times that Workflow Execution has persisted its state. Available only for closed Workflows. |
| TaskQueue             | Keyword  | Task Queue used by Workflow Execution.                                        |
| TemporalChangeVersion | Keyword  | If workflow versioning is enabled, list of change/version pairs will be stored here.|
| BinaryChecksums       | Keyword  | List of binary Ids of Workers that run the Workflow Execution.                              |
| BatcherNamespace      | Keyword  | Used by internal batcher to indicate the Namespace where batch operation was applied to.                                        |
| BatcherUser           | Keyword  | Used by internal batcher to indicate the user who started the batch operation.                                         |
```

PR #1195, opened 2022-05-25
Style/update UI

GraceGardner:
**PREVIEW**: https://deploy-preview-1195--mystifying-fermi-1bc096.netlify.app/

## What does this PR do?
Update Styles for landing page
## Notes to reviewers
Let me know what you think, happy to update anything that you might need.


djmagee:
This looks great! I have only three bits of feedback, and only the first is significant, IMNSHO:

- The page-title font size likely needs to decrease. See, for example, [tcld namespace update accepted-client-ca set](https://deploy-preview-1195--mystifying-fermi-1bc096.netlify.app/cloud/tcld/namespace/update/accepted-client-ca/set).
- At least on my monitor, the hairline around a code block fades into the background along the top in light mode. (I usually use dark mode, so I care less than others.)
- On the Documentation landing page, the leading between sections feels "spacey." (That's just an opinion; I know that this is a design decision.)

aarohib:
Adding to what Dail said:
- The difference between the content text size and the header size is really affecting readabiility. I had to squint to read the content (I have high astigmatism and myopia though).
- I see the text size in lists is bigger than the text size of paragraphs in some places (notably the app dev guide section though not sure why it isn't uniformly so across)

GraceGardner:
Thanks for all of the feedback! I made a handful of new adjustments. Let me know if you see anything else. Y'all are awesome!

djmagee:
Thanks for the H1 tweaks! I also love the colored path to the current node in the left nav—I don't think I've seen that done before.

Code blocks in light mode still blend into the background: 
image
Dark mode has good contrast.

The baseline of `inline code (monospace)` renders two or three points lower than the baseline of the surrounding text. Is that something that we can control? (And if so, is it worth fussing with?)

GraceGardner:
> Thanks for the H1 tweaks! I also love the colored path to the current node in the left nav—I don't think I've seen that done before.
> 
> Code blocks in light mode still blend into the background: image Dark mode has good contrast.
> 
> The baseline of `inline code (monospace)` renders two or three points lower than the baseline of the surrounding text. Is that something that we can control? (And if so, is it worth fussing with?)

Let me ask Candace about this. She probably has a good suggestion. Not sure if it will happen in this PR or a new one. :) Thanks for the feedback!

flossypurse (approved):
lgtm!

PR #1247, opened 2022-06-16
Dev guide: Fill in TS signals

lorensr:
Not sure why some of the generated files have added newlines @ EOF


flossypurse (approved):
Ya, this is a great addition.

Issue #1220, opened 2022-06-06
[ Bug report ] Allow linkable headers inside Tabs


lorensr:
## Required

**What is the issue?**

Non-Go (default lang) h4s in the app dev guide aren't linkable. 

**What is the expected behavior?**

Linkable. 

### Solutions

Two solutions that come to mind are:

1. Different pages with different default (SSRed) langs, like `docs.temporal.io/typescript/guide` has TypeScript tab default selected.
2. Rendering all langs on the page at once, and an onload script that changes lang if the hash isn't currently visible? The page would be pretty large and not sure how complex the script would be.

### Related issues

Being able to link to an h3 section with a particular lang preselected. (Would be enabled by Solution 1)

## Additional context (optional—the more we know, the faster it goes)

**Screenshots:**

<img width="905" alt="image" src="https://user-images.githubusercontent.com/251288/172197768-029ea580-ae91-4d44-8e06-710fe92d628d.png">

flossypurse:
The app dev guide does support h4s.
You just can't have the h4s in the source content - if the source content is going to be tabulated.
The reason is because Docusaurus builds all of the h4s from each tab into the toc - which then makes some of the links in the toc unusable depending on the selected tab. 

But that feels like a completely separate issue from being able to link to a selected tab - which I think would be a great feature. And it seems like @lorensr is making progress on that one with the framework.

lorensr:
Changed the issue title to `Allow linkable headers inside Tabs`. `h5`s are fine, and don't appear in TOC, but they're not linkable—when I open

http://localhost:3000/application-development-guide/?lang=typescript#testh5-1

the browser stays at the top of the screen, unless the heading is in a Go Tab (the first Tab, which is displayed first, before the `?lang=typescript` is processed and the tab is changed).

So in order for them to be linkable, I think the selected tab would need to SSRed. That may be implemented upstream as part of https://github.com/facebook/docusaurus/issues/7008

PR #1261, opened 2022-06-23
add script to set tab language from query string

vkarpov15:
Re: #1220

## What does this PR do?

Going to https://docs.temporal.io/typescript/workflows?language=js will set the language to js in tabs.

I wasn't able to figure out how to do this in React. For some reason, pulling in `import { useTabGroupChoice } from '@docusaurus/theme-common'` and using `useTabGroupChoice()` in a component was giving me an [invalid hook call warning error](https://reactjs.org/warnings/invalid-hook-call-warning.html). But, luckily, we can just set the correct localStorage key before React hydrates.

### DIFF
--- a/static/scripts/set-tab-language.js
+++ b/static/scripts/set-tab-language.js
@@ -18,4 +18,8 @@
   } else if (entries.language === "ts") {
     window.localStorage.setItem("docusaurus.tab.language", "ts");
   }
+
+  if (entries.lang) {
+    window.localStorage.setItem("docusaurus.tab.site-lang", entries.lang);
+  }
 })();
--- a/static/scripts/set-tab-language.js
+++ b/static/scripts/set-tab-language.js
@@ -1,0 +1,18 @@
+(function () {
+  const querystring = window.location.search;
+  if (!querystring) {
+    return;
+  }
+
+  const entries = querystring.slice(1).split('&').reduce((entries, str) => {
+    const [key, value] = str.split('=').map(v => decodeURIComponent(v));
+    entries[key] = value;
+    return entries;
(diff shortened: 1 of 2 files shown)


lorensr (approved):
Awesome, thanks! I added app dev guide, like `docs.temporal.io/application-development-guide/?lang=java`

static/scripts/set-tab-language.js
```diff
@@ -0,0 +1,25 @@
...
+      entries[key] = value;
+      return entries;
+    }, {});
+
+  if (entries.language === "js") {
+    window.localStorage.setItem("docusaurus.tab.language", "js");
+  } else if (entries.language === "ts") {
+    window.localStorage.setItem("docusaurus.tab.language", "ts");
+  }
+
```
vkarpov15, line 21: I'd check if entries.lang is something valid

lorensr: 👍 done & tested

lorensr:
App dev guide version is ?lang=

vkarpov15:
@lorensr it looks like, in the app dev guide, it always opens the Go tab first and then switches over to the one specified in `?lang`, see below video:

https://www.loom.com/share/952bf39312994447b33cd65e8aede082

Worth fixing?

lorensr:
It's the same behavior without `?lang=`, and just selecting Java and reloading. Same for TS docs. I imagine the first is SSRed and the change happens after reading from LocalStorage. 

Changing the behavior might be important for the linking-to-h4s-inside-Tab issue, but not for this PR.

@rachfop could you approve?

PR #1263, opened 2022-06-24
adding dprint config files

rachfop:
## What does this PR do?
Run `yarn dprint` to run the formatter. 
This is significantly faster than prettier. 

`yarn format`:  4.97s.
`yarn dprint`: 0.15s.

**Versioned docs**
>This PR is now in compliance with the versioned-docs.

`yarn format`: 9.46s.
`yarn dprint`: 0.22s.

Performanced optimized with `"incremental": true,`.
 dprint will only format files that have changed since the last time you formatted the code in order to drastically improve performance.

**Incremental format**
`yarn dprint`: 0.18s.

## Notes to reviewers

I have not updated the `.github/workflows/autoformat.yml` script. 
This is something _we_ should look into when we have bandwidth.
GitHub's action [here](https://github.com/marketplace/actions/dprint-check-action).

Left the following prettier rules alone:
```
.prettierrc.json
.prettierignore
babel.config.js
```
You can view the dprint rules at `dprint.json`.
They use all the same rules as [prettier](https://prettier.io/docs/en/options.html).

The following are dprint's ignore rules in case you want to add it for specific lines of code.

```
  "ignoreDirective": "dprint-ignore",
  "ignoreFileDirective": "dprint-ignore-file",
  "ignoreStartDirective": "dprint-ignore-start",
  "ignoreEndDirective": "dprint-ignore-end"
```

## dprint rules playground
View the rule's playground for additional rules [here](https://dprint.dev/playground/#code/N4AgvkA/language/markdown).

## To install dprint
These instructions are not needed, but keeping here for legacy’s sake.

1. Install dprint:
`brew install dprint`
1. Setup dprint:
`dprint init`
1. Add dprint plugins:
```
dprint config add typescript
dprint config add markdown
dprint config add json  
dprint config add prettier
```
4. Upgrade dprint plugins:
`dprint config update`
1. ~~Configure incremental false:
`dprint fmt --incremental=false`~~
1. Run dprint and output the time it takes:
`dprint output-format-times`

### DIFF
--- a/docs/php/testing.md
+++ b/docs/php/testing.md
@@ -42,7 +42,6 @@ temporal-test-server
 
 ### How it works
 
-
 For testing workflows there is no need to run a full Temporal server (with storage and UI interface).
 Instead, we can use a light-weight test server.
 
@@ -137,7 +136,6 @@ final class WaitWorkflow
     }
 }
 ```
-
 
 The activity will be called immediately. But, there may be cases when you do need to wait. So, to change this behaviour
 you can use `TestService` class:


flossypurse (approved):
Let's do it!

PR #1280, opened 2022-07-07
Describe heartbeat throttling

lorensr:
- Should this content also live in any `/concepts` page? I see that `docs/concepts/what-is-a-heartbeat-timeout.md` links to the app dev guide section I edited, so maybe doesn't need to be duplicated anywhere.
- I see a lot of content in `versioned_docs/version-june-2022/activities.md` that I can't find in the unversioned docs, and I'm wondering if that was preserved & will be added back in?

### DIFF
--- a/docs/app-dev-context/activity-heartbeats.md
+++ b/docs/app-dev-context/activity-heartbeats.md
@@ -1,0 +1,16 @@
+An Activity Heartbeat is a ping from the Worker that is executing the Activity to the Temporal Cluster.
+Each Heartbeat informs the Temporal Cluster that the Activity Execution is making progress and the Worker has not crashed.
+If the Cluster does not receive a Heartbeat within a [Heartbeat Timeout](/concepts/what-is-a-heartbeat-timeout) time period, the Activity will be considered failed and another [Activity Task Execution](/concepts/what-is-an-activity-task-execution) may be scheduled according to the Retry Policy.
+
+Heartbeats may not always be sent to the Cluster—they may be throttled by the Worker.
+The throttle interval is the smaller of:
+
+- `if heartbeatTimeout is provided, heartbeatTimeout * 0.8; else defaultHeartbeatThrottleInterval`
+- `maxHeartbeatThrottleInterval`
+
+`defaultHeartbeatThrottleInterval` is 30 seconds by default, and `maxHeartbeatThrottleInterval` is 60 seconds by default.
+Each can be set in Worker options.
+
+If an Activity is Cancelled, it will receive the Cancellation request when the next heartbeat is sent to the Cluster.
+
+Heartbeats may contain a `details` field describing the Activity's current progress. If an Activity gets retried, the Activity can access the `details` from the last heartbeat that was sent to the Cluster.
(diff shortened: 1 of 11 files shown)


docs/app-dev-context/activity-heartbeats.md
```diff
@@ -0,0 +1,16 @@
+An Activity Heartbeat is a ping from the Worker that is executing the Activity to the Temporal Cluster.
+Each Heartbeat informs the Temporal Cluster that the Activity Execution is making progress and the Worker has not crashed.
+If the Cluster does not receive a Heartbeat within a [Heartbeat Timeout](/concepts/what-is-a-heartbeat-timeout) time period, the Activity will be considered failed and another [Activity Task Execution](/concepts/what-is-an-activity-task-execution) may be scheduled according to the Retry Policy.
+
+Heartbeats may not always be sent to the Cluster—they may be throttled by the Worker.
```
Spikhalskiy, line 5: Let's add a more explicit explanation of how they are throttled. It actually may be important that heartbeats that happen during the throttling period are not discarded and the last details are actually sent at the end.

This is how I tried to highlight it in the original thread:

> After sending a heartbeat we set a timer. We "turn on" throttling and stop sending heartbeats until the timer fires. We still observe and remember the last heartbeat that you sent during the throttling period. When the timer fires, we send the last heartbeat that came in during the throttling period and set the timer again.

docs/app-dev-context/activity-heartbeats.md
```diff
@@ -0,0 +1,16 @@
+An Activity Heartbeat is a ping from the Worker that is executing the Activity to the Temporal Cluster.
+Each Heartbeat informs the Temporal Cluster that the Activity Execution is making progress and the Worker has not crashed.
```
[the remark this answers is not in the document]
lorensr: This is at bottom, after throttling so that reader knows what this part means:

> from the last heartbeat that was sent to the Cluster

docs/app-dev-context/activity-heartbeats.md
lorensr: 👍 Sounds good

docs/app-dev-context/activity-heartbeats.md
```diff
@@ -0,0 +1,16 @@
...
+Heartbeats may not always be sent to the Cluster—they may be throttled by the Worker.
+The throttle interval is the smaller of:
+
+- `if heartbeatTimeout is provided, heartbeatTimeout * 0.8; else defaultHeartbeatThrottleInterval`
+- `maxHeartbeatThrottleInterval`
+
+`defaultHeartbeatThrottleInterval` is 30 seconds by default, and `maxHeartbeatThrottleInterval` is 60 seconds by default.
+Each can be set in Worker options.
+
+If an Activity is Cancelled, it will receive the Cancellation request when the next heartbeat is sent to the Cluster.
```
Spikhalskiy, line 14: Let's more explitly highlight the connection between Activity Cancelation and Throttling. Maybe something like...

> Activity Cancelations are delivered to the Activities from the Server when they heartbeat. Activity that doesn't heartbeat can't receive a cancellation. Heartbeat Throttling may lead to a longer Cancelation delivery than expected.

docs/app-dev-context/activity-heartbeats.md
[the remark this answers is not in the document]
Spikhalskiy: Nevermind, found it later in the text.

Spikhalskiy:
Thank you for picking it up so quickly, Loren! Left some suggestions.

Spikhalskiy (approved):

PR #1295, opened 2022-07-13
Add info about Mocha and Jest to TS docs

vkarpov15:
## What does this PR do?

Re: discussion from https://github.com/temporalio/temporal-pendulum/pull/18, I think it is better to guide devs to use any testing framework other than Jest by default. This PR makes the testing docs suggest Mocha, while still linking to the Jest sample and adding more info on 1) how to get Jest to work, and 2) supported versions of Jest. I took a look and our Jest samples don't work with Jest 26.x.


lorensr (approved):
I like Jest's built-in capabilities, so post setup (what you note and what the SDK had to do internally), I prefer Jest's DX, and would personally recommend it on that basis. So slight pref for keeping default example Jest, but fine with it being Mocha if people feel strongly about it.

My personal initial reaction coming to Temporal and seeing the test sample in Mocha was "oh the creators of this are old-fashioned / not up with the times" (no offense meant, just sharing my knee-jerk reaction that I imagine others might have). And then working with it, trying to remember how to do things in Mocha (which I used to use before Jest), feeling frustrated by things not being built in that I had become accustomed to. There's also the comparative usage. Jest skews FE, so not sure what the majority of people learning Temporal are familiar with. https://npmtrends.com/ava-vs-jest-vs-mocha-vs-vitest

bergundy:
I'd keep jest as the default runner since as I mentioned in the pendulum PR and so has Loren here (with citations) it's by far the most popular test runner.
Up until today we've handled all of the issues with using jest.
@vkarpov15 if you've got some time maybe you can look into this one: https://github.com/temporalio/sdk-typescript/issues/753

bergundy (changes requested):
Thanks for this.
As mentioned before I would like to support jest.
We should consider adding the comments at the bottom of the file though.

docs/typescript/testing.md
```diff
@@ -351,3 +349,12 @@ await worker.runUntil(
   testEnv.workflowClient.execute(functionToTest, workflowOptions) // Throws WorkflowFailedError
 );
 ```
+
+### Testing with Jest
+
+Although these testing docs use [Mocha](https://mochajs.org/), you can also use [Jest](https://jestjs.io/).
+There are a couple of caveats for testing with Jest:
+
+1. The Temporal TypeScript SDK only supports Jest `>= 27.0.0`.
```
bergundy, line 358: We should note why

docs/typescript/testing.md
```diff
@@ -351,3 +349,12 @@ await worker.runUntil(
...
+2. The Temporal TypeScript SDK recommends using ESM. In order to test ESM modules, you need to run Jest with the `NODE_OPTIONS` environment variable set to `--experimental-vm-modules`, for example: `NODE_OPTIONS=--experimental-vm-modules jest`.
```
bergundy, line 359: We don't recommend this, it's up to the users to decide.
ESM is still pretty new and is still not the recommended method in Node.js.

docs/typescript/testing.md
lorensr: @vkarpov15 do you remember why?

bergundy (commented):
Let's not merge this for now, I think we may have fixed the jest segfault issues and I'd rather keep it as default

docs/typescript/testing.md
```diff
@@ -351,3 +354,12 @@ await worker.runUntil(
...
 );
 ```
+
+### Testing with Jest
+
+Although these testing docs use [Mocha](https://mochajs.org/), you can also use [Jest](https://jestjs.io/).
+There are a couple of caveats for testing with Jest:
+
+1. The Temporal TypeScript SDK only supports Jest `>= 27.0.0`.
+2. You need to run Jest with the `NODE_OPTIONS` environment variable set to `--experimental-vm-modules`, for example: `NODE_OPTIONS=--experimental-vm-modules jest`.
```
bergundy, line 364: This won't be relevant in the latest release AFAICT

docs/typescript/testing.md
```diff
@@ -351,3 +354,12 @@ await worker.runUntil(
...
+3. Make sure you run Jest with [`testEnvironment: 'node'`](https://jestjs.io/docs/configuration#testenvironment-string). `testEnvironment: 'jsdom'` is not supported.
```
lorensr, line 365:
```suggestion
1. Make sure you run Jest with [`testEnvironment: 'node'`](https://jestjs.io/docs/configuration#testenvironment-string). `testEnvironment: 'jsdom'` is not supported.
```

docs/typescript/testing.md
lorensr: Removed, and edited PR title to match current state

docs/typescript/testing.md
```diff
@@ -351,3 +354,11 @@ await worker.runUntil(
   testEnv.workflowClient.execute(functionToTest, workflowOptions) // Throws WorkflowFailedError
 );
 ```
+
+### Testing with Jest
+
+Although these testing docs use [Mocha](https://mochajs.org/), you can also use [Jest](https://jestjs.io/).
```
lorensr, line 360:
```suggestion
```

PR #1316, opened 2022-07-22
Fixing custom Archiver link in "Set up Archival"

Jkat:
## What does this PR do?

This fixes the link in the "Set up Archival" doc to point correctly to the "Custom Archiver" doc.


flossypurse:
Changed merged in - thanks for pointing this out @Jkat !

PR #1340, opened 2022-07-27
Fix legacy docs link

lorensr:


flossypurse (approved):
Yes, good call to keep this up for now.

Issue #1328, opened 2022-07-26
Keep anchor when switching Tabs


lorensr:
If you have an anchor selected, and then click a tab - the anchor ref is wiped from the url string.

start with this: `/application-development-guide#develop-workflows`
click a tab and you get this: `/application-development-guide?lang=typescript`

So, for this to work, the tab must be selected first, then the anchor.

Any way to make it append the url param to the end of the url - after the anchor - rather than clearing it?

_Originally posted by @flossypurse in https://github.com/temporalio/documentation/issues/1292#issuecomment-1189213713_

PR #1368, opened 2022-08-03
Upgrading Docusaurus

flossypurse:
- Upgrades Docusuarus to v2
- Adjusts Tabs swizzle component imports to work with upgrade
- Adds handling for https://github.com/temporalio/documentation/issues/1328


djmagee:
Switching from one page of the App dev guide to another keeps the previously selected tab selected, but the tab isn't represented in the URL; for example, when going from `application-development/observability?lang=python#tracing` to `application-development/features`, the Python tab remains selected, but the URL no longer contains `?lang=python`. Is that behavior in scope or out of scope for this PR?

flossypurse:
> Switching from one page of the App dev guide to another keeps the previously selected tab selected, but the tab isn't represented in the URL; for example, when going from `application-development/observability?lang=python#tracing` to `application-development/features`, the Python tab remains selected, but the URL no longer contains `?lang=python`. Is that behavior in scope or out of scope for this PR?

So that was the behavior before this change.

And I think it is out of scope for this PR as I am only affecting things when the tab is selected.

That sort of thing would need to be addressed somewhere else I think.

djmagee (approved):

lorensr (approved):

flossypurse:
Build passed

PR #1388, opened 2022-08-05
Fix typo in tctl workflow cancel documentation

FinnitoProductions:
## What does this PR do?
The documentation for `tctl workflow cancel` provides an example for `--run_id` in which `show` is used instead of `cancel`. Fixes that typo.


lorensr (approved):
Thanks for the fix! 😊

PR #1398, opened 2022-08-09
Update minimum length of a Namespace Name

djmagee:
## What does this PR do?

Per [PR 1359](https://github.com/temporalio/saas-control-plane/pull/1359) in [temporalio/saas-control-plane](https://github.com/temporalio/saas-control-plane/), updates the minimum length of a Namespace Name in "What is a Cloud Namespace Name?"

### DIFF
--- a/docs/concepts/what-is-a-cloud-namespace-name.md
+++ b/docs/concepts/what-is-a-cloud-namespace-name.md
@@ -13,7 +13,7 @@ It cannot be changed after the Namespace is provisioned.
 
 Each Namespace Name must conform to the DNS naming rules:
 
-- A Namespace Name must contain at least 1 character and no more than 34 characters.
+- A Namespace Name must contain at least 2 characters and no more than 34 characters.
 - A Namespace Name must begin or end with a letter or number and can contain only letters, numbers, and the hyphen (-) character.
 - Each hyphen (-) character must be immediately preceded _and_ followed by a letter or number; consecutive hyphens are not permitted.
 - All letters in a Namespace Name must be lowercase.


aarohib (approved):

docs/concepts/what-is-a-cloud-namespace-name.md
```diff
@@ -13,7 +13,7 @@ It cannot be changed after the Namespace is provisioned.
 
 Each Namespace Name must conform to the DNS naming rules:
 
-- A Namespace Name must contain at least 1 character and no more than 34 characters.
+- A Namespace Name must contain at least 2 characters and no more than 34 characters.
 - A Namespace Name must begin or end with a letter or number and can contain only letters, numbers, and the hyphen (-) character.
```
feihuang, line 17: cc @mastermanu right now we require it to begin with a letter (not a number).

docs/concepts/what-is-a-cloud-namespace-name.md
```diff
@@ -13,7 +13,7 @@ It cannot be changed after the Namespace is provisioned.
...
 - Each hyphen (-) character must be immediately preceded _and_ followed by a letter or number; consecutive hyphens are not permitted.
```
feihuang, line 18: good point on consecutive hyphens but we currently don't reject them...

docs/concepts/what-is-a-cloud-namespace-name.md
mastermanu: correct

feihuang (approved):

docs/concepts/what-is-a-cloud-namespace-name.md
djmagee: OK as is for now, or should I do one of the following?

- Remove the list item.
- Change sentence to "For forward compatibility, each hyphen (-) character should be immediately preceded _and_ followed by a letter or number; consecutive hyphens should not be used."

@mastermanu @feihuang

feihuang: I think we should just keep it as it is and on backend we'll just enforce no consecutive dashes before GA.

On a minor point, line 14 talks about DNS naming rules but since 1 character is valid dns name, our rules have deviated from dns rules?

djmagee: Good point. I'll modify line 14 to skip the reference to DNS.

PR #1399, opened 2022-08-09
swapping

rachfop:
## What does this PR do?
Swaps **my** to **your**.

This is a PR to clean up code examples that still use the _**My**Workflow_ pattern. 
We agreed a while back to use _**Your**Workflow_ and _**Your**Activity_ style for function names.

This shouldn't have any effect on code samples as the functions are just placeholders for real values.

Matched previous casing to this PR - so there shouldn't be any issues with camel case or snake casing either.

### DIFF
--- a/docs/application-development/features.md
+++ b/docs/application-development/features.md
@@ -1892,13 +1892,13 @@ Either `scheduleToCloseTimeout` or `scheduleToStartTimeout` must be set.
 Type: time.Duration
 Default: ∞ (infinity – no limit)
 
-In this example, you can set the `ScheduleToStartTimeout` to 60 seconds.
+In this example, you can set the `scheduleToStartTimeout` to 60 seconds.
 
 ```typescript
 // Sample of typical options you can set
 const {greet} = proxyActivities<typeof activities>({
   scheduleToCloseTimeout: "5m",
-  ScheduleToStartTimeout: "60s",
+  scheduleToStartTimeout: "60s",
   retry: {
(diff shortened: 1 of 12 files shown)


docs/application-development/features.md
```diff
@@ -344,7 +344,7 @@ Content is not available
 ```ts
 import {setHandler} from "@temporalio/workflow";
 
-export async function myWorkflow() {
+export async function YourWorkflow() {
```
lorensr, line 347: Lowercase

lorensr (approved):
What was the reason? I remember the "using you instead of we" slack thread, but this seems different—since it's code the user is copying or typing. To me, doesn't feel right for me, the user, to write "function yourWorkflow() {"

docs/application-development/features.md
flossypurse: @rachfop I am not considering function names to be placeholders and so I don't think that style guidance applies here. Though there could be other style guidance I am not following - so we could check up on that.

@lorensr This was my call - I want our app dev guide code samples to be explicit about what is 100% your code - your functions names - your custom data structures - etc.
My thinking on is that it helps to reduce reader fatigue - and clearly illustrate which key words are part of the SDK within the context of the things that the user is creating.

This is a content design choice I implemented in the legacy docs for the Go SDK - which as a whole have been receiving great feedback in our surveys. I can't make 100% correlation there - but my instinct is that is a good habit for us to stick with.

djmagee (approved):

PR #1409, opened 2022-08-12
Update how-to-set-clientoptions-in-go.md

rachfop:
## What does this PR do?

Currently, there isn't docs for this option in the docs. 

[Go API docs](https://github.com/temporalio/sdk-go/blob/v1.16.0/internal/interceptor.go#L37) say:

> Interceptor is a common interface for all interceptors. See documentation in the interceptor package for more details.

But the documenation doesn't provide more details, so I'm hoping this PR will do just that.

The only place in the docs that talk about interceptors ([here](https://docs.temporal.io/application-development/foundations/#run-worker-processes)) links to the [TS docs](https://docs.temporal.io/typescript/interceptors).

## Notes to reviewers

We will want to write the interceptors docs soon - but this PR doesn't address that. It just fills in the gaps in the Go docs.


docs/go/how-to-set-clientoptions-in-go.md
```diff
@@ -136,3 +136,5 @@ Set to induce artificial failures in test scenarios
 - Type: `TrafficController`
 
 ### Interceptors
```
djmagee, line 138: Should this heading also be styled as `inline code`?

docs/go/how-to-set-clientoptions-in-go.md
```diff
@@ -136,3 +136,5 @@ Set to induce artificial failures in test scenarios
...
+
+Grpc interceptors which will be applied to every RPC call performed by this connection. By default, an interceptor will be included which automatically retries retryable errors. If you do not wish to perform automatic retries, set this to an empty list (or a list with your own interceptors).
```
djmagee, line 140:
```suggestion
gRPC interceptors that are applied to every RPC call performed by this connection.
By default, an interceptor is included; it automatically retries retryable errors.
If you do not want to perform automatic retries, set this to an empty list (or a list with your own interceptors).
```

djmagee (approved):

PR #1418, opened 2022-08-16
Change log ➡️ changelog

### LINKED ISSUE #1415: [ Bug report ] Hard to find example applications and tutorials from docs.temporal
## Required

**What is the issue?**
From docs.temporal.io, I cannot easily navigate to learn.temporal.io. As such, I cannot easily find the tutorials.

1. Go to docs.temporal.io.
1. Try to find tutorials and example applications.

There exists a quick way to get between {docs,learn}.temporal.io. E.g., on docs, a link in the header to learn.

lorensr:
## What does this PR do?

- Fixes #1415
- Uses the more familiar-to-developers spelling of "changelog"

Prefer shorter/cleaner "Changelog" over "Docs changelog" in nav. Should still be clear after you click it that it's for docs, given the header:

image

Sidebar:

image


flossypurse (commented):
We have a couple of other things that might get added to the main nav which will be docs site specific.

I would prefer to build out our footer links more and add the learn link there for now.

As for the change log -> changelog change -

@djmagee and I followed up on style guidance and there are plenty of examples going both ways for different reasons - and we settled on keeping it as "Docs change log" specifically - keep "Docs" there so its clear it is not a Product change log.

lorensr:
My guess is a footer link might not meet @afitz0's desired level of discoverability

lorensr:
> @djmagee and I followed up on style guidance and there are plenty of examples going both ways for different reasons - and we settled on keeping it as "Docs change log" specifically - keep "Docs" there so its clear it is not a Product change log.

Did you consider mental load? My sense is changelog is more familiar for developers, and due to that and it being one word, it's easier to see/comprehend in a nav list. 

Re: prefixing with `Docs ` in nav, I think 1 word is easier to see/comprehend than 3 words, and IMO [easier-to-see/comprehend for everyone] is more important than [some people mistakenly click it only to learn it's the Docs changelog].

afitz0:
RE: Learn link. As a user, Documentation and Learning are nearly identical to me, and putting Learn in the collection of footer links feels a little like shipping an org chart. (That is to say, @lorensr is right, adding `Learn` to the footer doesn't eel like increasing discoverability to me.)

lorensr:
If there won't be header real estate, and alternative is what Apollo does: list the tutorials in the left nav of docs, and they take you to their separate Odyssey learning platform:

https://www.apollographql.com/docs/

image

flossypurse:
Picking up the learn link change here: https://github.com/temporalio/documentation/pull/1575

flossypurse:
Down to keep this alive if we go with "Docs changelog" for the header nav label and sidebar label and keep it focused on the changelog 

Basically I want to keep any non docs.temporal.io links in the footer (already updated them per this and other prs) -

LMK -

lorensr:
@flossypurse Updated PR to just `Docs changelog`

PR #1427, opened 2022-08-18
Format and update external resources

lorensr:


flossypurse (approved):
Noice.

In the near future a lot of the content on this page could benefit from Dev rel ownership - in blog posts etc - 
I would also consider retaining some of it in the form of KB articles.

PR #1490, opened 2022-08-22
Remove blog from docs site

flossypurse:


lorensr (approved):
New links don't work for me yet. If the paths are the same, you can add 

`/blog/*  /:splat`

to `_redirects`

PR #1504, opened 2022-08-24
Side Effects and Mutable Side Effects

rachfop:
### Side Effect
Adds Workflow Side Effect for:
- [ ] Go
- [ ] Java

Python and TypeScript do not currently support Side Effects.

https://github.com/temporalio/features/issues/561

### Mutable Side Effect
Adds Mutable Side Effect for:
- [ ] Go
- [ ] Java
- [ ] PHP
- [ ] Python
- [ ] TypeScript

## NOTE

I didn't touch any of the random files in this PR.

It's probably from running `yarn gen` and Update branch.

### DIFF
--- a/docs/application-development/features.md
+++ b/docs/application-development/features.md
@@ -3313,7 +3313,13 @@ values={[{label: 'Go', value: 'go'},{label: 'Java', value: 'java'},{label: 'PHP'
 
 <TabItem value="go">
 
-Content is not available
+To use [`MutableSideEffect()`](https://pkg.go.dev/go.temporal.io/sdk/workflow#MutableSideEffect) in Go, provide the Workflow Id.
+
+```go
+if err := workflow.MutableSideEffect(ctx, "configureNumber", get, eq).Get(&number); err != nil {
+		panic("can't decode number:" + err.Error())
+	}
+```
 
 </TabItem>
 <TabItem value="java">
--- a/docs/go/how-to-set-mutable-side-effects-in-go.md
+++ b/docs/go/how-to-set-mutable-side-effects-in-go.md
@@ -1,0 +1,18 @@
+---
+id: how-to-set-mutable-side-effects-in-go
+title: How to set Mutable Side Effects in Go
+sidebar_label: Set Mutable Side Effects
+description: To use MutableSideEffect() in Go, provide the Workflow Id.
+tags:
+  - developer-guide
+  - sdk
+  - Go
+---
+
+To use [`MutableSideEffect()`](https://pkg.go.dev/go.temporal.io/sdk/workflow#MutableSideEffect) in Go, provide the Workflow Id.
+
+```go
+if err := workflow.MutableSideEffect(ctx, "configureNumber", get, eq).Get(&number); err != nil {
+		panic("can't decode number:" + err.Error())
+	}
+```
--- a/guide-gen/guide-configs/app-dev/features.json
+++ b/guide-gen/guide-configs/app-dev/features.json
@@ -922,7 +922,7 @@
             {
               "lang": "go",
               "title": "Go",
-              "path": "none"
+              "path": "go/how-to-set-mutable-side-effects-in-go"
             },
             {
               "lang": "java",


docs/app-dev-context/side-effects.md
```diff
@@ -0,0 +1,20 @@
+---
+id: side-effects
+title: Side Effects
```
djmagee, line 3: Should we combine this file with `\concepts\what-is-a-side-effect` and then use the one file both here and in the Glossary?

docs/app-dev-context/side-effects.md
```diff
@@ -0,0 +1,20 @@
...
+description: Side Effects are a method of execution to produce nondeterministic code.
+sidebar_label: Side Effects
+tags:
+  - guide-context
+---
+
+A Side Effect is a method of execution to produce nondeterministic code; such as, generating a UUID or a random number.
+By implementing Side Effect into your Workflow Execution, you can execute the provided function once and records its result into the Workflow Execution [Event History](../concepts/what-is-an-event-history).
```
djmagee, line 11:
```suggestion
A Side Effect is a method of execution to produce nondeterministic code, such as generating a UUID or a random number.
By implementing a Side Effect in your Workflow Execution, you can execute the provided function once and record its result into the Workflow Execution [Event History](/workflows/#event-history).
```

docs/go/how-to-use-a-side-effect-in-go.md
```diff
@@ -0,0 +1,44 @@
...
+var random int
+encodedRandom.Get(&random)
+if random < 50 {
+       ....
+} else {
+       ....
+}
+```
+
+The only way to fail Side Effect is to panic, which causes Workflow Task failure. The Workflow Task after Timeout is rescheduled and re-executed giving Side Effect another chance to succeed. Be careful to not return any data from the Side Effect function any other way than through its recorded return value.
```
djmagee, line 26:
```suggestion
The only way to fail a Side Effect is to panic, which causes Workflow Task failure.
The Workflow Task after Timeout is rescheduled and re-executed giving the Side Effect another chance to succeed.
Be careful not to return any data from the Side Effect function any way other than through its recorded return value.
```

docs/go/how-to-use-a-side-effect-in-go.md
```diff
@@ -0,0 +1,44 @@
...
+
+Do not use `SideEffect()` to modify closures. Always retrieve result from Side Effect's encoded return value.
```
djmagee, line 28:
```suggestion
Do not use `SideEffect()` to modify closures.
Always retrieve results from a Side Effect's encoded return value.
```

docs/java/how-to-use-a-side-effect-in-java.md
```diff
@@ -0,0 +1,62 @@
...
+```java
+  int random = Workflow.sideEffect(Integer.class, () -> random.nextInt(100));
+  if random < 50 {
+         ....
+  } else {
+         ....
+  }
+```
+
+Another example using `sideEffect()`.
```
djmagee, line 22:
```suggestion
Here's another example that uses `sideEffect()`.
```

cretz:
How sure are we that we want side effects and mutable side effects documented in the app dev guide? They are Java/Go/PHP only and other languages may never implement them. There are lots of features of specific APIs we don't document in the app dev guide and instead refer to the SDK documentation.

Can we get business/product confirmation that this is something we want to document in our cross-language guide? If we do, can business/product then provide the text making it clear this is not available in TypeScript/Python by intention?

cretz:
Not sure we should document this in the cross-language app dev guide until we're sure we're ever even going to have this feature across languages. If we choose not to have this feature across languages, we need to decide whether this is worth documenting at all and/or what the language would be saying why other languages don't have it.

Spikhalskiy:
It does definitely need to be documented as a side effect is the only safe/correct way of working with external configuration in Workflow code. Using local activities for it is a temporary hack. We should discuss though if we exposed a safer / more user-friendly interface over it, like a "Marker", because side-effect is a little technical and, actually, misleading.

cretz:
@Spikhalskiy - TS impl issue at https://github.com/temporalio/sdk-typescript/issues/144 is where we could potentially discuss different name or approach (no issue for Core or Python yet, it'd come from that). Then we can come back and update the doc stuff. Unsure if we need to leave this doc PR on hold until that is discussed.

flossypurse:
We have already decided that it is perfectly acceptable to document a feature that is not at parity with the other SDKs...
We simply indicate that the feature is not currently available with that SDK.

docs/go/how-to-use-a-side-effect-in-go.md
```diff
@@ -0,0 +1,47 @@
+---
+id: how-to-use-a-side-effect-in-go
```
flossypurse, line 2: Were you aware of the existing file for this?
https://github.com/temporalio/documentation/blob/master/docs/go/how-to-execute-a-side-effect-in-go.md

tbh - I think the existing one is more to the point and clearer atm

cretz:
@flossypurse - :+1: My concern was whether this feature is important enough to document in the guide. Note, the app guide doesn't document a dozen Go SDK features (await, selectors, channels, etc) and every SDK has many features we choose not to document there. If we feel side effects aren't deserving of this kind of documentation and may just add more noise than value, we can leave it up to the API documentation like we do with _lots_ of other features.

flossypurse:
@cretz 
> Note, the app guide doesn't document a dozen Go SDK features (await, selectors, channels, etc)

Valid point are these features covered by the sdk-features repo?

Are there other "concurrency" type of features in other SDKs that are somewhat on parallel?
For example, is threading possible in Java, and do we expose a specific wrapper API to do that in Java like we do with Go channels?

@rachfop can you remind me if this is being tracked anywhere atm or if we made a decision about this yet?

cretz:
> Valid point - are these features covered by the sdk-features repo? [...] Are there other "concurrency" type of features in other SDKs that are somewhat on parallel?

Not all of them, or not yet. But for example, all SDKs provide a "wait for condition" type of API that we don't document. Same with timers/sleep (or at least I can't find the docs). Same for starting a coroutine.

Basically the actual API documents lots of things the docs doesn't (and probably shouldn't). Some are features all SDKs have, some are features only some SDKs have, some are features only one SDK has. The decision to document should just be based on importance vs noise.

> For example, is threading possible in Java, and do we expose a specific wrapper API to do that in Java like we do with Go channels?

For channels, Java does have `Workflow.newQueue` which provides similar behavior to Go's `workflow.NewBufferedChannel`. For threading/coroutines, Java does have `Async.function`/`Async.procedure` which provides similar behavior to Go's `workflow.Go`.

I don't believe we document any of this (and again, maybe we shouldn't, but maybe we should, unsure).

PR #1509, opened 2022-08-26
Adds a note to Run ID topic

rachfop:
## What does this PR do?
[The discussion here](https://github.com/temporalio/api/pull/182) makes it clear that you shouldn't rely on a Run Id when deciding business logic. 

If a Workflow resets, it resets the Workflow History, which can lead to non-determinism issues if that ID is relied on.
## Notes to reviewers

This PR doesn't mention `original_execution_run_id` or `first_execution_run_id` as that goes above the need of the customer. Instructing one not to rely on the Run Id is much better than the logic needed to use the `original_execution_run_id`.

## Note

Used `caution` in the admonitions and not just `note` as this specific issue is notable.

## Note 2

The other random files were probably included from running prettier.

### DIFF
--- a/docs/clusters.md
+++ b/docs/clusters.md
@@ -125,7 +125,6 @@ When implemented, the `frontend` invokes the plugin before executing the request
 
 Temporal offers two plugin interfaces for API call authentication and authorization.
 
-
 - [`ClaimMapper`](#what-is-a-claimmapper-plugin?)
 - [`Authorizer`](#what-is-an-authorizer-plugin?)
 


docs/concepts/what-is-a-run-id.md
```diff
@@ -15,3 +15,9 @@ For example, a Temporal Cron Job is a chain of Workflow Executions that all have
 Each Workflow Execution within the chain is considered a _Run_.
 
 A Run Id uniquely identifies a Workflow Execution even if it shares a Workflow Id with other Workflow Executions.
+
+:::caution
+
+Run Id is unsafe to be used in a Workflow Execution for any logical choices, as a reset operation might lead to non-determinism issues.
```
djmagee, line 21:
```suggestion
Run Id is unsafe to be used in a Workflow Execution for any logical choices because a reset operation can lead to non-determinism.
```

docs/concepts/what-is-a-run-id.md
```diff
@@ -15,3 +15,9 @@ For example, a Temporal Cron Job is a chain of Workflow Executions that all have
 Each Workflow Execution within the chain is considered a _Run_.
 
 A Run Id uniquely identifies a Workflow Execution even if it shares a Workflow Id with other Workflow Executions.
+
+:::caution
+
+Run Id is unsafe to be used in a Workflow Execution for any logical choices because a reset operation can lead to non-determinism.
```
flossypurse, line 21: Unsafe for which logical choices?

Pretty sure it is safe for logical choices when it comes to visibility and listing and signal from a client - etc... 

Also, I can not find the discussion that mentions this - wrong link maybe?

flossypurse, line 21: Unsafe for which logical choices?

Pretty sure it is safe for logical choices when it comes to visibility and listing and signal from a client - etc... 

Also, I can not find the discussion that mentions this - wrong link maybe?

flossypurse (changes requested):

PR #1512, opened 2022-08-29
removing Replay banner

rachfop:
## What does this PR do?
Removes the Replay banner on the docs site.

### DIFF
--- a/docs/components/RetrySimulator/retry-simulator.module.css
+++ b/docs/components/RetrySimulator/retry-simulator.module.css
@@ -95,6 +95,8 @@
 .numberInputLabel {
   float: left;
   max-width: 69%;
+  background-color: var(--ifm-background-color);
+  color: var(--docsearch-text-color);
 }
 
 .numberInput {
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -55,14 +55,7 @@ module.exports = {
         autoCollapseCategories: true,
       },
     },
-    announcementBar: {
```
djmagee, line 58: Should we comment out this section so that we can easily resurrect it for a future announcement?

flossypurse: Yes, I agree - et us comment out rather than remove.

Also - @djmagee are we not converting this to a message about Cloud now?

flossypurse: Yes, I agree - let us comment out rather than remove.

Also - @djmagee are we not converting this to a message about Cloud now?

flossypurse (changes requested):

djmagee:
In this case, not really version control ('cause the next announcement overwrites this one entirely); more a template for the next announcement. (Swap the text and identifier, uncomment, _et voila!_)

djmagee (approved):

flossypurse (approved):

PR #1514, opened 2022-08-29
rewrote some parts for clarity and accuracy

DJSanti:
Using Dominik and Timo's insight in a Slack Thread, I've rewritten reuse policies for clarity and accuracy.

### DIFF
--- a/docs/concepts/what-is-a-workflow-id-reuse-policy.md
+++ b/docs/concepts/what-is-a-workflow-id-reuse-policy.md
@@ -2,27 +2,22 @@
 id: what-is-a-workflow-id-reuse-policy
 title: What is a Workflow Id Reuse Policy?
 sidebar_label: Workflow Id Reuse Policy
-description: A Workflow Id Reuse Policy determines whether a Workflow Execution is allowed to spawn with a particular Workflow Id, if that Workflow Id has been used with a previous, and now Closed, Workflow Execution.
+description: Explanation of the three different Workflow Id Reuse Policies
 tags:
   - explanation
 ---
 
-A Workflow Id Reuse Policy determines whether a Workflow Execution is allowed to spawn with a particular Workflow Id, if that Workflow Id has been used with a previous, and now Closed, Workflow Execution.
+A Workflow Id Reuse Policy determines if new Workflow Executions with a given Workflow Id are allowed.
+The operation fails if the Workflow Id is identical to an Open Workflow Execution.
 
-It is not possible for a new Workflow Execution to spawn with the same Workflow Id as another Open Workflow Execution.
-An attempt to spawn a Workflow Execution with a Workflow Id that is the same as the Id of a currently Open Workflow Execution results in a `Workflow execution already started` error.
-
+These policies apply to the `StartWorkflowExecution` function.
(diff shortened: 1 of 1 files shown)


docs/concepts/what-is-a-workflow-id-reuse-policy.md
```diff
@@ -2,27 +2,22 @@
 id: what-is-a-workflow-id-reuse-policy
 title: What is a Workflow Id Reuse Policy?
 sidebar_label: Workflow Id Reuse Policy
-description: A Workflow Id Reuse Policy determines whether a Workflow Execution is allowed to spawn with a particular Workflow Id, if that Workflow Id has been used with a previous, and now Closed, Workflow Execution.
+description: Explanation of the three different Workflow Id Reuse Policies
```
flossypurse, line 5: This should always be the same as the first sentence in the article.

We only do this for the guides that are generated from these source files.

docs/concepts/what-is-a-workflow-id-reuse-policy.md
```diff
@@ -2,27 +2,22 @@
...
+A Workflow Id Reuse Policy determines if new Workflow Executions with a given Workflow Id are allowed.
+The operation fails if the Workflow Id is identical to an Open Workflow Execution.
 
+These policies apply to the `StartWorkflowExecution` function.
 A Workflow Id Reuse Policy has three possible values:
 
-- **Allow Duplicate** The Workflow Execution is allowed to exist regardless of the Closed status of a previous Workflow Execution with the same Workflow Id.
+- **Allow Duplicate**: allows a new run of the Workflow to exist with the same Workflow Id as a closed Workflow Execution.
   **This is the default policy, if one is not specified.**
-  Use this when it is OK to have a Workflow Execution with the same Workflow Id as a previous, but now Closed, Workflow Execution.
```
flossypurse, line 19: Why is this removed?

flossypurse (changes requested):
This removes too much accurate and important info.

I think the ask here was to:
- add additional explanation around the StartWorkflowExecution operation and that it is at that point which the Workflow Id Reuse Policy comes into play to determine if the new Run can start
- add a diagram.

docs/concepts/what-is-a-workflow-id-reuse-policy.md
```diff
@@ -2,27 +2,22 @@
...
+description: Explanation of the three different Workflow Id Reuse Policies
 tags:
   - explanation
 ---
 
-A Workflow Id Reuse Policy determines whether a Workflow Execution is allowed to spawn with a particular Workflow Id, if that Workflow Id has been used with a previous, and now Closed, Workflow Execution.
-
-It is not possible for a new Workflow Execution to spawn with the same Workflow Id as another Open Workflow Execution.
-An attempt to spawn a Workflow Execution with a Workflow Id that is the same as the Id of a currently Open Workflow Execution results in a `Workflow execution already started` error.
+A Workflow Id Reuse Policy determines if new Workflow Executions with a given Workflow Id are allowed.
```
flossypurse, line 10: This is less accurate and complete than what was here previously.

PR #1523, opened 2022-08-31
Update how-to-send-a-signal-from-a-workflow-in-go.md

AyoWafi:
signalData is not declared, should be signal

## What does this PR do?
Fixes an undeclared variable "signalData" in the example


flossypurse:
Hey @AyoWafi - thanks for pointing this out!
Issue is now fixed!

PR #1537, opened 2022-09-04
docs: correct the web-UI port

thuan1412:
## What does this PR do?
Change the port of web UI from 8080 to 8888
## Notes to reviewers
In docker-compose files except the `docker-compose.yml`, there is no service that exposes to port 8080.

### DIFF
--- a/docs/application-development/foundations.md
+++ b/docs/application-development/foundations.md
@@ -96,7 +96,7 @@ The following steps start and run a Temporal Cluster using the default configura
    docker-compose up
    ```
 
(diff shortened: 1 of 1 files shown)


thuan1412:
@flossypurse 
> Have you updated your Temporal docker-compose repo lately?
Yes, I did.

In this docker-compose file, there is no service that exposes port 8080
https://github.com/temporalio/docker-compose/blob/main/docker-compose-postgres.yml#L55, only 8088. Did I miss something?

flossypurse:
@thuan1412 
Looks like that config is missing the `temporal-ui` stuff like this:
https://github.com/temporalio/docker-compose/blob/main/docker-compose.yml#L64

We will get that added.

PR #1578, opened 2022-09-14
Add more info on how to deploy TS workers in Docker

mjameswh:
## What does this PR do?

Added explanations and examples on how to properly deploy TypeScript SDK based Workers in Docker. These additions answer to questions we regularly get on support channels.


flossypurse (commented):
Oh, we should totally make this a KB article actually.

lorensr:
@flossypurse what do you think about a Deployment section of app dev guide? It includes automated tests and debugging in prod, seems universal & helpful to tell people in guide how to get to prod.

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+COPY . /app
+WORKDIR /app
+
+RUN npm install --only=production \
+    && npm run build
+
+CMD ["build/worker.js"]
+```
+
+For smaller images and/or more secure deployments, it is also possible to use `node:slim` Docker image variants (eg. `node:16-bulleyes-slim`), or `distroless/nodejs` Docker images (ie. `grc.io/distroless/nodejs:16`). See details below.
```
lorensr, line 154:
```suggestion
For smaller images and/or more secure deployments, it is also possible to use `-slim` Docker image variants (like `node:16-bullseye-slim`) or `distroless/nodejs` Docker images (like `grc.io/distroless/nodejs:16`) with the below caveats.
```

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+## Running in Docker
+
+Workers based on TypeScript SDK can be deployed and run as Docker containers.
+
+At this moment, we recommend usage of NodeJS 16 (note that there are known issues with NodeJS 18). Both `amd64` and `arm64` platforms are supported. A glibc-based image is required; musl-based images are _not_ supported (see below).
+
+The easiest way to deploy a TypeScript SDK Worker on Docker is to start with the `node:16-bulleyes` image. For example:
+
+```dockerfile
+FROM node:16-bulleyes
```
lorensr, line 143:
```suggestion
The easiest way to deploy a TypeScript SDK Worker on Docker is to start with the `node:16-bullseye` image. For example:

```dockerfile
FROM node:16-bullseye
```

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+CMD ["build/worker.js"]
+```
+
+For smaller images and/or more secure deployments, it is also possible to use `node:slim` Docker image variants (eg. `node:16-bulleyes-slim`), or `distroless/nodejs` Docker images (ie. `grc.io/distroless/nodejs:16`). See details below.
+
+### Using `node:slim` images
+
+`node:slim` images do not contain some of the common packages found in regular images. This results in significantly smaller images.
+
+Note however that TypeScript SDK requires the presence of root TLS certificates (ie. the `ca-certificates` package), which are not included in `slim` images. This package is required even when connecting to a local Temporal Server, and when using a server connection config that doesn't explicitly use TLS.
```
lorensr, line 160:
```suggestion
### Using `node:slim` images

`node:slim` images do not contain some of the common packages found in regular images. This results in significantly smaller images.

However, TypeScript SDK requires the presence of root TLS certificates (the `ca-certificates` package), which are not included in `slim` images. `ca-certificates` package is required even when connecting to a local Temporal Server or when using a server connection config that doesn't explicitly use TLS.
```

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+FROM node:16-bulleyes-slim
+
+RUN apt-get update \
+    && apt-get install -y ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+# ... same as with regular image
+```
+
+Failure to install this dependency results in `[TransportError: transport error]` runtime error, because the certificates cannot be verified.
```
lorensr, line 174:
```suggestion
Failure to install this dependency results in a `[TransportError: transport error]` runtime error, because the certificates cannot be verified.
```

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+
+### Using `distroless/nodejs` images
+
+`distroless/nodejs` images includes only the files that are strictly required to execute node. This results in even smaller images (approximately half the size of `node:slim` images). It also significantly reduce the surface of potential security issues that could be exploited by an eventual hacker in the resulting Docker images.
```
lorensr, line 178:
```suggestion
`distroless/nodejs` images include only the files that are strictly required to execute `node`. This results in even smaller images (approximately half the size of `node:slim` images). It also significantly reduces the surface of potential security issues that could be exploited by a hacker in the resulting Docker images.
```

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+
+It is generally possible and safe to execute TypeScript SDK Workers using `distroless/nodejs` images (ie. unless your code itself requires dependencies that are not included in `distroless/nodejs`).
```
lorensr, line 180:
```suggestion
It is generally possible and safe to execute TypeScript SDK Workers using `distroless/nodejs` images (unless your code itself requires dependencies that are not included in `distroless/nodejs`).
```

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+
+COPY --from=builder /app /app
+WORKDIR /app
+
+CMD ["build/worker.js"]
+```
+
+### Properly configure Node's memory in Docker
+
+By default, `node` configures its maximum old-gen memory to 25% of the _physical memory_ of the machine on which it is executing, with a maximum of 4 GB. This is very likely innappropriate when running node in a Docker environment and can result in either under usage of available memory (ie. node only uses a fraction of the memory allocated to the container) or overusage (ie. node tries to use more memory than what is allocated to the container, which will eventually lead to the process being killed by the operating system).
```
lorensr, line 209:
```suggestion
By default, `node` configures its maximum old-gen memory to 25% of the _physical memory_ of the machine on which it is executing, with a maximum of 4 GB. This is very likely innappropriate when running node in a Docker environment and can result in either under usage of available memory (`node` only uses a fraction of the memory allocated to the container) or overusage (`node` tries to use more memory than what is allocated to the container, which will eventually lead to the process being killed by the operating system).
```

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+
+It is therefore recommanded that you always explicitly set the `--max-old-space-size` node argument to approximately 80% of the maximum size (in megabytes) that you want to allocate the node process. You might need some experimentation and adjustment to find the most appropriate value based on your specific application.
```
lorensr, line 211:
```suggestion
It is therefore recommanded that you always explicitly set the `--max-old-space-size` `node` argument to approximately 80% of the maximum size (in megabytes) that you want to allocate the `node` process. You might need some experimentation and adjustment to find the most appropriate value based on your specific application.
```

lorensr: maybe also mention workflow cache size worker config?

docs/typescript/production-deploy.md
```diff
@@ -131,11 +131,91 @@ We endeavor to give you good defaults, so you don't have to worry about them, bu
...
+
+In practice, it is generally easier to provide this argument through the `NODE_OPTIONS` environment variable.
```
lorensr, line 213:
```suggestion
In practice, it is generally easier to provide this argument through the [`NODE_OPTIONS` environment variable](https://nodejs.org/api/cli.html#node_optionsoptions).
```

lorensr (approved):

docs/typescript/production-deploy.md
mjameswh: Yeah, I was considering it too, but had too little time to write this at this point.

However, it might be better to move this subsection and the one about configuring workflow cache size to a distinct sections. Would also mention `workflowThreadPoolSize` in this. All of these configuration items applies both to docker and non-docker cases, though being in docker make adjusting these a little bit more important.

PR #1594, opened 2022-09-17
avoid triming TypeScript code samples, use dedent instead

### LINKED ISSUE #1503: [ Bug report ] TypeScript / JavaScript whitespace
## Required

**What is the issue?**
When switching between TypeScript/JavaScript code samples, I've noticed that TypeScript uses 4 spaces `    `, whereas JavaScript only uses 2 spaces `  `.

For example, [on this page](https://docs.temporal.io/application-development/observability/?lang=js#set-custom-search-attributes).
## TypeScript

```typescript
const handle = await client.start(example, {
    taskQueue: 'search-attributes',
    workflowId: 'search-attributes-example-0',
    searchAttributes: {
      CustomIntField: [2],
      CustomKeywordField: ['keywordA', 'keywordB'],
      CustomBoolField: [true],
      CustomDatetimeField: [new Date()],
      CustomStringField: [
        'String field is for text. When queried, it will be tokenized for partial match. StringTypeField cannot be used in Order By',
      ],
    },
  });

  const { searchAttributes } = await handle.describe();
```

## JavaScript
```javascript
const handle = await client.start(example, {
  taskQueue: 'search-attributes',
  workflowId: 'search-attributes-example-0',
  searchAttributes: {
    CustomIntField: [2],
    CustomKeywordField: ['keywordA', 'keywordB'],
    CustomBoolField: [true],
    CustomDatetimeField: [new Date()],
    CustomStringField: [
      'String field is for text. When queried, it will be tokenized for partial match. StringTypeField cannot be used in Order By',
    ],
  },
});

const { searchAttributes } = await handle.describe();
```

Because of that, it appears that TypeScript isn't hightlighting the "keys" or var names and JavaScript does.

## Additional context (optional—the more we know, the faster it goes)

**Screenshots:**

## TypeScript

Screen Shot 2022-08-24 at 8 23 15 AM

## JavaScript

Screen Shot 2022-08-24 at 8 23 22 AM

vkarpov15:
## What does this PR do?

Fixes #1503. The problem is that the current `docusaurus.config.js` runs `trim()` on TypeScript code, which messes up indentation on snippets where every line is indented because the snippet is pulled from a function body. For example, the following screenshot shows the first line of code is flush against the left because of `trim()`.

image

With this PR, the previous snippet is rendered correctly, including trimming leading whitespace:

image

Note that I tried using the [dedent npm package](https://www.npmjs.com/package/dedent), but that package does not handle empty lines.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -241,7 +241,7 @@ module.exports = {
                     if (!/^ts$/.test(node.lang)) {
                       return;
                     }
-                    node.value = "// @ts-nocheck\n" + node.value.trim();
+                    node.value = "// @ts-nocheck\n" + node.value;
                   }
 
                   visit(tree, "code", visitor);
@@ -276,6 +276,9 @@ module.exports = {
                     }
                     if (node.value.startsWith("// @ts-nocheck\n")) {
                       node.value = node.value.slice("// @ts-nocheck\n".length);
+                      if (node.lang === "ts") {
+                        node.value = dedent(node.value);
+                      }
                     }
                     // If TS compiled output is empty, replace it with a more helpful comment
                     if (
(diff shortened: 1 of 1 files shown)


lorensr (approved):
nice, thanks! 😊

Issue #1379, opened 2022-08-05
[ Bug report ] TS snippet format


lorensr:
## Required

**What is the issue?**

At some point this repo lost the spaces between `{}`:

<img width="678" alt="image" src="https://user-images.githubusercontent.com/251288/182986807-920457c2-5f16-496b-b8bf-8d28da550f7d.png">

Note that learn.temporal.io has correct spaces:

<img width="730" alt="image" src="https://user-images.githubusercontent.com/251288/182986862-b280c08c-85c4-4c83-a313-9cd8d14bd6d3.png">

cc @vkarpov15

lorensr:
@rachfop just realized this is only in app dev guide, not old TS docs. Could you look into it?

PR #1596, opened 2022-09-17
correct prettier overrides for TypeScript app dev guide

vkarpov15:
## What does this PR do?

I suspect the issue for #1379 is that we don't run prettier on docs-src. The snippets that have formatting issues are all written inline in the documentation. I manually ran the code snippets through Prettier to make the formatting more consistent, and added an override for `docs-src/typescript` to the Prettier config just in case.

Do you think it's worth it to run prettier on all the app dev guides @lorensr ?


flossypurse (approved):

.prettierrc.json
```diff
@@ -17,11 +17,14 @@
   "useTabs": false,
   "overrides": [
     {
-      "files": ["docs/typescript/**/*.md"],
+      "files": [
+        "docs/typescript/**/*.md",
+        "docs-src/typescript/*.md"
```
flossypurse, line 22: I thought it hit the whole repo by default...
Why wouldn't we want prettier to run on everything in docs-src ?

lorensr: This doesn't change the script (`"format": "prettier --write .",`)—these are just the file paths for the TS prettier rules that are different from the default/general rules.

lorensr:
@flossypurse Is `npm run gen` run by netlify / are updates to the app dev guide reflected in the deploy preview? Or is that waiting on Vercel?

.prettierrc.json
vkarpov15: It isn't that we don't want to run prettier on `docs-src/typescript/**/*.md`, we just want to apply a couple of rule overrides to those directories.

vkarpov15:
Overall I like the idea of merging this into #1608. There isn't anything in this PR that wouldn't be also be accomplished by a correct prettier config.

PR #1620, opened 2022-09-22
Clarify worker_task_slots_available

lorensr:
## What does this PR do?

Answers a question from @maelropars


docs/references/sdk-metrics.md
```diff
@@ -28,10 +28,10 @@ Each metric may have some combination of the following keys attached to them:
   - `workflow_task`
   - `activity_task`
   - `sticky_workflow_task`
-- `worker_type`: One of the following:
-  - `ActivityWorker`
+- `worker_type`: One of the below values. Note that if a single Worker is configured to handle all three types of tasks, it will emit all three metrics.
```
lorensr, line 31: I generally try to avoid "Note that", but didn't think of a better way (parens?) to frame this kind of side-note information inside this bullet. (But do think it belongs in the bullet.)

djmagee: I agree and can't think of a cleaner way to include it.

docs/references/sdk-metrics.md
```diff
@@ -31,10 +31,10 @@ Each metric may have some combination of the following keys attached to them:
...
```
djmagee, line 34:
```suggestion
- `worker_type`: One of the following values. Note that if a single Worker is configured to handle all three types of Tasks, it emits all three metrics.
```

docs/references/sdk-metrics.md
```diff
@@ -296,9 +296,14 @@ A Worker Entity has been registered, created, or started.
 
 ### worker_task_slots_available
 
-The total number of Workflow Task and Activity Task execution slots that are currently available.
-Use the `worker_type` key to differentiate execution slots.
-(Workflow Workers execute Workflow Tasks; Activity Workers execute Activity Tasks.)
+The total number of task execution slots that are currently available.
```
djmagee, line 299:
```suggestion
The total number of Task Execution slots that are currently available.
```

djmagee (approved):

PR #1622, opened 2022-09-23
adding mutable and side effect to docs

rachfop:
## What does this PR do?
This is a cleanuped version [of this PR.](https://github.com/temporalio/documentation/pull/1504)

Updates:
- Took [Cully's request](https://github.com/temporalio/documentation/pull/1504#discussion_r958527387) and replaced the Go example.
- "we're never even going to have this feature across languages."
- Should talk about how to handle 100% missing features in futher doc iterations, but that shouldn't block this PR.
## Notes to reviewers
😎 Added an `na` value for the the id in Assembly workflow.

[Side Effects not supported](https://github.com/temporalio/features/issues/561)
- [x] TS
- [x] Python


assembly/guide-configs/app-dev/features.json
```diff
@@ -585,6 +585,64 @@
...
+          "lang": "php",
+          "id": "none"
+        },
+        {
+          "lang": "python",
+          "id": "none"
+        },
+        {
+          "lang": "typescript",
+          "id": "none"
```
lorensr, line 613:
```suggestion
          "id": "na"
```

and generator fills in with "Not applicable to this SDK." instead of "Content is currently unavailable."

assembly/activities/genMarkdownGuides.js
```diff
@@ -63,6 +63,8 @@ async function generateLangTabs(langtabs) {
     tabStr = `${tabStr}<TabItem value="${tab.lang}">\n\n`;
     if (tab.id == "none") {
       tabStr = `${tabStr}Content is currently unavailable.\n\n`;
+    } else if (tab.id == "na") {
+      tabStr = `${tabStr}The selected SDK does not currently support this feature.\n\n`;
```
lorensr, line 67:
```suggestion
      tabStr = `${tabStr}Not applicable to this SDK.\n\n`;
```

don't want to ~imply that we may/will in future

flossypurse (approved):

docs-src/app-dev-context/mutable-side-effects.md
```diff
@@ -0,0 +1,23 @@
+---
+id: mutable-side-effects
+title: Mutable Side Effects
+description: Mutable Side Effects are a method of execution to produce nondeterministic code.
+sidebar_label: Mutable Side Effects
+tags:
+  - guide-context
+---
+
+Mutable Side Effects execute the provided function once, and then it looks up the History of the value with the given Workflow ID.
```
Spikhalskiy, line 10: Effects multiple, but function singular? Doesn't make sense to me. Probably better to rephrase in singulars.

Spikhalskiy: "then it looks up the History of the value with the given Workflow ID."
I don't understand what it means, it doesn't sound right. It doesn't look up a history of the value. Workflow ID is irrelevant here.

This segment from the javadoc describes it better:

"Executes the provided function once, records its result into the workflow history. The recorded
   * result on history will be returned without executing the provided function during replay. This
   * guarantees the deterministic requirement for workflow as the exact same result will be returned
   * in replay."

Spikhalskiy: > Mutable Side Effects execute the provided function once, and then it looks up the History of the value with the given Workflow ID.

Should be just ID or Side Effect ID, Workflow ID is incorrect.

"execute the provided function once"

This is unclear. Should be just "executes the provided function during original execution (not replay)"

docs-src/app-dev-context/mutable-side-effects.md
```diff
@@ -0,0 +1,23 @@
...
+
+- If there is no existing value, then it records the function result as a value with the given Workflow ID on the History.
+- If there is an existing value, then it compares whether the existing value from the History has changed from the new function results, by calling the equals function.
+  - If the values are equal, then it returns the original value without recording a new History
+  - If the values aren't equal, then it records the new value with the same ID on the History.
+
+:::note
+
+During a Workflow Execution, every new Side Effect call results in a new mark recorded on the Workflow History; whereas Mutable Side Effects only record an entry on the Workflow History if the value changes.
```
Spikhalskiy, line 19:
```suggestion
During a Workflow Execution, every new Side Effect call results in a new marker recorded on the Workflow History; whereas Mutable Side Effects only records a new marker on the Workflow History if the value for the Side Effect ID changes or is set the first time.
```

docs-src/app-dev-context/mutable-side-effects.md
```diff
@@ -0,0 +1,23 @@
...
+sidebar_label: Mutable Side Effects
+tags:
+  - guide-context
+---
+
+Mutable Side Effects execute the provided function once, and then it looks up the History of the value with the given Workflow ID.
+
+- If there is no existing value, then it records the function result as a value with the given Workflow ID on the History.
+- If there is an existing value, then it compares whether the existing value from the History has changed from the new function results, by calling the equals function.
+  - If the values are equal, then it returns the original value without recording a new History
```
Spikhalskiy, line 14:
```suggestion
  - If the values are equal, then it returns the value without recording a new Marker Event
```

docs-src/app-dev-context/side-effects.md
```diff
@@ -0,0 +1,20 @@
+---
+id: side-effects
+title: Side Effects
+description: Side Effects are a method of execution to produce nondeterministic code.
+sidebar_label: Side Effects
+tags:
+  - guide-context
+---
+
+A Side Effect is a method of execution to produce nondeterministic code, such as generating a UUID or a random number.
```
Spikhalskiy, line 10: Not a method of execution. Doesn't produce code.

A decent piece from javadocs:

"Executes the provided function once, records its result into the workflow history. The recorded
   * result on history will be returned without executing the provided function during replay. This
   * guarantees the deterministic requirement for workflow as the exact same result will be returned
   * in replay."

docs-src/app-dev-context/side-effects.md
```diff
@@ -0,0 +1,20 @@
...
+By implementing a Side Effect in your Workflow Execution, you can execute the provided function once and record its result into the Workflow Execution [Event History](/workflows/#event-history).
```
Spikhalskiy, line 11:
```suggestion
Using a Side Effect in your Workflow Execution, you can execute the provided function once and record its result into the Workflow Execution [Event History](/workflows/#event-history).
```

docs-src/app-dev-context/side-effects.md
```diff
@@ -0,0 +1,20 @@
...
+
+A Side Effect does not re-execute during a Replay. Instead, it returns the recorded result from the Workflow Execution Event History.
+Side Effects should not fail, because failure results in the Side Effect function executing more than once. If there’s a chance of failure, use an Activity.
```
Spikhalskiy, line 14:
```suggestion
Side Effects should not fail. An exception that is thrown from the Side Effect causes failure and retry of the current Workflow Task.

An Activity or a Local Activity may also be used instead of a Side effect, as its result is also persisted in Workflow Execution History.
```

docs-src/app-dev-context/side-effects.md
```diff
@@ -0,0 +1,20 @@
...
+
+:::note
+
+You shouldn’t modify the Workflow state inside a Side Effect. Instead, use the return value.
```
Spikhalskiy, line 18:
```suggestion
You shouldn’t modify the Workflow state inside a Side Effect function, because it is not reexecuted during Replay. Side Effect function should be... free of side-effects and only return a value.
```

docs/application-development/features.md
```diff
@@ -3082,6 +3082,196 @@ const handle = await client.start(scheduledWorkflow, {
 </TabItem>
 </Tabs>
 
+## Side Effects
+
+A Side Effect is a method of execution to produce nondeterministic code, such as generating a UUID or a random number.
```
Spikhalskiy, line 3087: Not a method of execution. Doesn't produce code.

Side Effects are a way to execute non-deterministic code and don't compromise the determinism of the workflow by storing the non-deterministic result into workflow history and getting the same exact value during Replay.

docs/application-development/features.md
```diff
@@ -3082,6 +3082,196 @@ const handle = await client.start(scheduledWorkflow, {
...
+
+    if(randomInt % 2 == 0) {
+        // ...
+    } else {
+        // ...
+    }
+}
+```
+
+Java also provides a deterministic method to generate random numbers or random UUIDs.
```
Spikhalskiy, line 3180: Do we really need the piece from here and below? Go snippet doesn't have it.
It's a good note, but it may be generalized in a language-agnostic note like "Temporal Workflow API provides ways to obtain most frequently used non-deterministic values like timestamps and random numbers in a deterministic way without using side-effects".

docs/go/how-to-set-mutable-side-effects-in-go.md
```diff
@@ -0,0 +1,18 @@
...
+title: How to set Mutable Side Effects in Go
+sidebar_label: Set Mutable Side Effects
+description: To use MutableSideEffect() in Go, provide the Workflow Id.
+tags:
+  - developer-guide
+  - sdk
+  - Go
+---
+
+To use [`MutableSideEffect()`](https://pkg.go.dev/go.temporal.io/sdk/workflow#MutableSideEffect) in Go, provide the Workflow Id.
```
Spikhalskiy, line 12: It's not a Workflow ID.

docs/java/how-to-use-a-side-effect-in-java.md
```diff
@@ -0,0 +1,62 @@
...
+
+    if(randomInt % 2 == 0) {
+        // ...
+    } else {
+        // ...
+    }
+}
+```
+
+Java also provides a deterministic method to generate random numbers or random UUIDs.
```
Spikhalskiy, line 42: Same as above, looks slightly irrelevant that we teach basic Workflow.* methods here.

Something like "Temporal Workflow API provides ways to obtain most frequently used non-deterministic values like timestamps and random numbers in a deterministic way without using side-effects" and maybe a link to these methods or the descriptions of these APIs would be more appropriate here.

Spikhalskiy:
We may want to directly compare Mutable Side Effect vs Side Effect to make things a little more clear for users. For example, something like "To make a connection between SE and MSE more clear, Side Effects in a workflow may be replaced with Mutable Side effects 1-1 if every Mutable Side Effect invocation is called with a unique SideEffectID."

PR #1636, opened 2022-09-28
Update command for getting search attributes

bmorton:
## What does this PR do?

I came across `tctl cluster get-search-attr` in documentation, which did not work.  I checked the docs and noticed it should be `tctl cluster get-search-attributes`, so I've updated to reflect that here!


djmagee:
Thanks for spotting these errors, @bmorton! I implemented your fixes (and made a couple of related changes) in #1637.

PR #1638, opened 2022-09-28
fix: 404 link to Temporal Cloud

MarceloAlves:
Clicked a link that ended up at a 404. Quick fix for that


djmagee:
Thanks for the catch, @MarceloAlves! I implemented the change in #1651, which modifies the source file from which the "Connect to a Cluster" section of the Application development guide is generated.

MarceloAlves:
@djmagee ah perfect! didn't know it should be changed there. thanks!!

PR #1643, opened 2022-09-29
feat: add php sdk mTLS

seregazhuk:
Adds a guide how to setup mtls connection with PHP SDK.

Ref: https://github.com/temporalio/samples-php/pull/24

### DIFF
--- a/assembly/guide-configs/app-dev/foundations.json
+++ b/assembly/guide-configs/app-dev/foundations.json
@@ -183,7 +183,7 @@
         },
         {
           "lang": "php",
-          "id": "none"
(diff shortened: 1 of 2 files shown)


flossypurse (commented):
2 things we will want to update here.
1 - new stuff should be going into `/docs-src/php` - that way we can move off of the siloed (legacy) docs set
2 - We recently updated the app dev guide to focus on "How to connect to a Cluster" in general - and those code samples show a client connection with all relevant values being set that are needed to at minimum - connect to Temporal Cloud.

seregazhuk:
@rachfop should I change something in this PR? 🤔  Or you will fix it?

PR #1663, opened 2022-10-08
Let web-ui decide which namespace to redirect to

yawboakye:
## What does this PR do?
currently the redirect assumes that the first namespace created will be the `default` namespace. this isn't always true given that the doc explicit recommends to replace `default` with user's own preferred namespace name. if the user follows the recommendation, then following the link from the docs brings them to an error page (where it says namespace was not found). this happened to me in fact. but i discovered that if i just navigate to http://127.0.0.1:8233/ the web ui is smart enough the redirect to my one and only namespace.

this happened with just `temporalite`, my preferred way to run temporal on my local machine, hence the change only affects that one line. it's possible that it applies to other ways to run temporal but i don't have the experience to recommend the change. regardless, letting web-ui decide where to redirect to sounds good to me.

## Notes to reviewers
all sorts of review welcome. it's my first contribution so i might be overlooking quite a lot of things here.

### DIFF
--- a/docs/application-development/foundations.md
+++ b/docs/application-development/foundations.md
@@ -67,7 +67,7 @@ The following steps start and run a Temporal Cluster.
    ```
    Replace `default` with your [Namespace Name](/cloud/#temporal-cloud-namespace-name).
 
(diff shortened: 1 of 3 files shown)


tsurdilo:
+1 for this, (not to hard-code "default" ns). Thanks for the PR!

yawboakye:
closing in favor of @rachfop's PR.

PR #1665, opened 2022-10-08
Use dprint in CI

lorensr:
## What does this PR do?

- Fixes #1379
- Hopefully fixes #748

## Notes to reviewers

Supercedes #1559

PR #1666, opened 2022-10-08
Use dprint in CI

lorensr:
## What does this PR do?

Version of #1665 on a local branch

### DIFF
--- a/.github/workflows/autoformat.yml
+++ b/.github/workflows/autoformat.yml
@@ -1,34 +1,23 @@
-name: Continuous Integration
+name: Format with dprint and commit changes
 
-# This action works with pull requests and pushes
 on:
-  pull_request:
-  push:
-    branches:
-      - master
+  # runs when a pull request is opened or reopened or when the head branch of the pull request is updated
+  pull_request_target:
 
 jobs:
-  prettier:
+  format:
     runs-on: ubuntu-latest
-
     steps:
-      - name: Checkout on push
-        # github object: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
-        # event names: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
-        if: "github.event_name == 'push'"
-        uses: actions/checkout@v2
+    
+    - uses: actions/checkout@v3
+      with:
+        repository: ${{ github.event.pull_request.head.repo.full_name }}
+        ref: ${{ github.head_ref }}
 
-      - name: Checkout on PR
-        if: "github.event_name == 'pull_request'"
-        uses: actions/checkout@v2
-        with:
-          # Make sure the actual branch is checked out when running on pull requests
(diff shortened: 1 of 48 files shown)


lorensr:
@mindaugasrukas Any idea why having this file in this PR:

https://github.com/temporalio/documentation/blob/dprint-ci/.github/workflows/autoformat.yml#L5

with `on: pull_request_target` 

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

doesn't result in the job being run?

Note that this file:

https://github.com/temporalio/documentation/blob/dprint-ci/.github/workflows/test.yml

with `on: pull_request` does run.

Background: I'm trying to get this Action to work on forks:

https://github.com/stefanzweifel/git-auto-commit-action#advanced-uses

(see first `<details>`)

The Action [isn't able to commit](https://github.com/temporalio/documentation/actions/runs/3211723085/jobs/5250091512) to a fork when triggered by `pull_request`.

lorensr:
Okay, after some trial and error and [discussion](https://github.com/stefanzweifel/git-auto-commit-action/discussions/220#discussioncomment-3845787), I've hit an impass. Here are the current options:

1. CI action runs dprint on PR. Passes if no changes. If changes and local branch, commits changes. If fork, fails in a way that shows it's a formatting failure, so if they run dprint locally and push, action will succeed.
2. CI action runs dprint and commit on merge.

The pro of latter is external contributors don't need to manually run dprint. The con of latter is the presence of "Format" automated commits on master. Thoughts? @flossypurse @rachfop

mindaugasrukas:
> Okay, after some trial and error and [discussion](https://github.com/stefanzweifel/git-auto-commit-action/discussions/220#discussioncomment-3845787), I've hit an impass. Here are the current options:
> 
> 1. CI action runs dprint on PR. Passes if no changes. If changes and local branch, commits changes. If fork, fails in a way that shows it's a formatting failure, so if they run dprint locally and push, action will succeed.
> 2. CI action runs dprint and commit on merge.
> 
> The pro of latter is external contributors don't need to manually run dprint. The con of latter is the presence of "Format" automated commits on master. Thoughts? @flossypurse @rachfop

I have concerns regarding automated commits on the master without the review.

lorensr:
@mindaugasrukas does the idea of limited scope help, or is it a more general security concern? 

I don't think not automatically dprinting on forks is that bad—it's normal IME contributing to see a failed format or lint action and then run a command and push.

flossypurse (approved):

lorensr:
I'll update this to do strat #1, and move from `yarn fmt` to `yarn format`, removing prettier

lorensr:
Confirmed this works as expected with a fork: #1727 

@rachfop I ran `dprint fmt` in this PR. The TS formatting changes are good. Can you verify that the other formatting changes are good?

PR #1668, opened 2022-10-09
Make API reference section first subsection

yawboakye:
## What does this PR do?
it reorders subsections under the "Add your SDK" section so that the API reference section comes first. this is kinda relevant because the API reference section recommends that the reader selects their programming language and following the specific link. given that i had already selected my programming language before i encountered this section it felt a little out of place. i thought it should be moved all the way up so that the reader encounters it before making any language specific decisions.

## Priority
low.

### DIFF
--- a/docs/application-development/foundations.md
+++ b/docs/application-development/foundations.md
@@ -122,6 +122,45 @@ An SDK provides you with the following:
 - APIs to create and manage [Worker Processes](/workers#).
 - APIs to author [Activities](/activities#activity-definition).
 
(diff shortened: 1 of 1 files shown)


flossypurse:
@yawboakye per our [Slack Thread](https://temporalio.slack.com/archives/C023SLH1KML/p1665523520166399) I think we are going to hold off on making this change. Also if we were to make it, this is not source file that needs the edits.
But your suggestion will definitely remain in consideration.

PR #1671, opened 2022-10-10
fix: add PHP search attirbutes

seregazhuk:


docs/concepts/what-is-a-search-attribute.md
```diff
@@ -132,3 +132,4 @@ To actually have results from the use of a [List Filter](/concepts/what-is-a-lis
 How to do this entirely depends on the method by which you spawn the Workflow Execution:
 
 - [How to set Search Attributes as Workflow Execution metadata in Go](/go/startworkflowoptions-reference/#searchattributes)
+- [How to set Search Attributes as Workflow Execution metadata in PHP](/php/how-set-search-attributes-as-workflow-execution-metadata-in-php)
```
flossypurse, line 135: We should just update both of these to:
```suggestion
- [How to set Search Attributes as Workflow Execution metadata](app-dev-context/set-custom-search-attributes)
```

docs/php/how-set-search-attributes-as-workflow-execution-metadata-in-php.md
```diff
@@ -0,0 +1,42 @@
+---
+id: how-set-search-attributes-as-workflow-execution-metadata-in-php
```
flossypurse, line 2: This file should be added to /docs-src/php directory

PR #1756, opened 2022-11-03
tcld Assembly config

flossypurse:


DJSanti (changes requested):
Just had one thing to say concerning a link. Other than that, everything is good to me. Probably not a bad idea to get an SME or Dail on here as well

djmagee:
My review is still in progress.

docs-src/cloud/tcld/index.md
```diff
@@ -0,0 +1,43 @@
...
+
+:::
+
+:::note Join the Temporal Cloud waitlist
+
+Access to Temporal Cloud is currently by invitation only.
+You can [join the waitlist](https://pages.temporal.io/cloud-early-access).
+
+:::
+
```
djmagee, line 23:
```suggestion
```
We can make these admonitions go away entirely.

djmagee (approved):
Just two tweaks from me.

DJSanti (approved):

PR #1766, opened 2022-11-07
remove tctl alias 

DJSanti:
tctl alias has been removed from tctl-next. Ruslan would like the docs to remove any pages about it.


DJSanti:
NOTE: these changes are incoming for tctl cli. It's not in yet; I'm just trying to get some stuff taken care of ahead of time.

PR #1777, opened 2022-11-08
fixing domain stripping from docusaurus

flossypurse:


djmagee (approved):
That was highly intuitive.

Not.

PR #1784, opened 2022-11-11
Fix for TypeScript patching versions to match code snippets

ricmatsui:
## What does this PR do?

The text for TypeScript patching currently uses `v1`, `v1patch`, `v2deprecatedpatch`, and `v2` as the names for the steps, but the code snippets being pulled in from the samples repo use `v1`, `v2`, `v3`, and `vFinal` as the names: 
https://github.com/temporalio/samples-typescript/tree/4812a9e12676393214657c78d7b57cb63599eb56/patching-api/src which makes it hard to follow which step is being referenced.

This PR fixes the text to match the code snippets being used.


lorensr (approved):
Thanks for the fix! ☺️

PR #1785, opened 2022-11-14
Set `public: true` in vercel.json

lorensr:
## What does this PR do?

Makes build logs visible to people outside the temporal vercel team, so that all contributors can see whether/how their PR is breaking the build and fix it.

https://vercel.com/docs/project-configuration#project-configuration/public


flossypurse (approved):
Nice!

PR #1786, opened 2022-11-15
TW-152 adds KB articles to the left nav

rachfop:
## What does this PR do?

Removes the 5 article limit.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -212,7 +212,7 @@ module.exports = {
           sidebarPath: require.resolve("./sidebars.js"),
           routeBasePath: "/",
           exclude: [], // do not render context content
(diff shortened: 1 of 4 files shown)


docusaurus.config.js
```diff
@@ -412,6 +412,7 @@ module.exports = {
         path: "kb",
         blogTitle: "Temporal Platform knowledge base",
         blogSidebarTitle: "Recent KB articles",
```
djmagee, line 414: If we change `blogSidebarCount` to `ALL` instead of a higher number, should we change `blogSidebarTitle` to simply "KB articles"?

flossypurse: Nice catch.

djmagee (approved):

PR #1799, opened 2022-11-17
Add 'disable_write_action' to env. vars. and web ui config

DJSanti:
Request from Ruslan:

Add DISABLE_WRITE_ACTIONS to environmental variables reference and to web ui configuration.


feedmeapples (approved):

flossypurse (approved):
Just run Assembly before merge.

PR #1818, opened 2022-11-26
Update KB verbs & ToC

lorensr:
## What does this PR do?

- Matches verb tense used elsewhere ("Develop Workflows")
- "DeadlineExceeded" looked like an error class. I don't know of an SDK that has one. I would either call it the `deadline exceeded` error or "deadline exceeded" or use its grpc status code `DEADLINE_EXCEEDED` (but that doesn't appear in all SDK's error messages—I think the former is more common).
- As a workflow developer, "client-side" makes me think communication from the Client (and not the Worker)
- ToC before and after:

image

image


kb/deadline-exceeded-troubleshooting.md
```diff
@@ -1,20 +1,20 @@
 ---
 slug: deadline-exceeded-troubleshooting
-title: DeadlineExceeded error troubleshooting
+title: Troubleshoot deadline exceeded error
```
djmagee, line 3:
```suggestion
title: Troubleshoot deadline-exceeded error
```

djmagee (commented):
Just one tweak to the title, and this is good to go.

djmagee (approved):

PR #1819, opened 2022-11-30
Tw 168 create documentation for workflow size constraint violations

DJSanti:
Added information about pending Events and Command failure to:
- Limits
- Workflows
- Commands reference

Please let me know where the information is also needed. I covered the areas mentioned by Cully and Michael at the start.


kb/temporal-platform-limits-sheet.md
```diff
@@ -28,3 +28,10 @@ Here is a list of many hard (error) or soft (warn) limits that you could encount
   - We error at 50,000 Events: [`history size exceeds error limit.`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/workflowExecutionContext.go#L1204)
   - This is configurable with [`HistoryCountLimitError` and `HistoryCountLimitWarn`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/configs/config.go#L382-L383), if you know what you are doing.
 - [Search Attributes maximums](/visibility/#search-attributes-maximums)
+- **Workflow field limits** (leading to Command failure):
```
flossypurse, line 31: What is a "field"here?

Also, looks like what you actually want to do is update and merge this info into the "History total count limit" stuff above...

Because that already calls out the hard limit of 50,000 Events.

docs-src/references/commands.md
```diff
@@ -50,27 +50,55 @@ This Command is triggered by a call to spawn a [Child Workflow Execution](/conce
 - Awaitable: Yes, a Workflow Execution can await on the action resulting from this Command.
 - Corresponding Event: [ChildWorkflowExecutionStarted](/references/events/#childworkflowexecutionstarted)
 
+:::note
+
+OSS users cannot have more than 50000 pending Child Workflows.
```
flossypurse, line 55: Is this saying the following:

Within all my Workflow code that is executing, if I have loops (for example) that spin off x number of Child Workflows - at any given time I should be wary not to call that API too much as I don't want to possibly exceed 5000 pending ChildWorkflows across my Namespace (or is it per Account ?) - and PENDING in this instance means that the Commands have not yet been processed into corresponding Events?

If so, let us clarify that -

docs/workflows.md
```diff
@@ -254,6 +254,24 @@ You can use the <a class="tdlp" href="#continue-as-new">Continue-As-New<span cla
 The Workflow Execution spawned from Continue-As-New has the same Workflow Id, a new Run Id, and a fresh Event History and is passed all the appropriate parameters.
 For example, it may be reasonable to use Continue-As-New once per day for a long-running Workflow Execution that is generating a large Event History.
 
+### Size constraints
```
flossypurse, line 257: Size constraints doesn't feel like the right section title for this.

Maybe just "Limits" ?

We already call out OSS Event hard limit here: https://docs.temporal.io/workflows#event-history

Maybe this limits section could be source of truth and we link to it from Event History ?

docs-src/references/commands.md
```diff
@@ -50,27 +50,55 @@ This Command is triggered by a call to spawn a [Child Workflow Execution](/conce
...
+:::
+
 ### SignalExternalWorkflowExecution
 
 This Command is triggered by a call to [Signal](/concepts/what-is-a-signal) another Workflow Execution.
 
 - Awaitable: Yes, a Workflow Execution can await on the action resulting from this Command.
 - Corresponding Event: [SignalExternalWorkflowExecutionInitiated](/references/events/#signalexternalworkflowexecutioninitiated)
 
+:::note
```
flossypurse, line 67: Do we need to make these notes or is it is just information?

docs-src/references/commands.md
djmagee, line 55:
```suggestion
OSS users cannot have more than 50,000 pending Child Workflows.
```
Comma added per https://developers.google.com/style/numbers#commas-and-decimal-points-in-numbers and https://learn.microsoft.com/en-us/style-guide/numbers#commas-in-numbers.

docs-src/references/commands.md
DJSanti: It's repetitive; will remove.

DJSanti:
Git erased some of my progress. If someone could check to see which of my changes actually made it through, I'd appreciate it. I may have just missed a link.

docs/concepts/what-is-a-workflow-execution.md
```diff
@@ -126,3 +126,19 @@ A Workflow Execution Timeout can be used to limit the duration of Workflow Execu
...
+
+:::note
+
+Cloud users have a hard limit of 2,000 pending Events.
+
+:::
+
+In addition to the hard limits imposed on Event Histories, certain Commands will not be processed when there are already 50,000 pending Events of a given type:
+
+- `StartTimer`
```
MichaelSnowden, line 140: There is no limit for StartTimer

docs/concepts/what-is-a-workflow-execution.md
```diff
@@ -126,3 +126,19 @@ A Workflow Execution Timeout can be used to limit the duration of Workflow Execu
 You can use the [Continue-As-New](/concepts/what-is-continue-as-new) feature to close the current Workflow Execution and create a new Workflow Execution in a single atomic operation.
 The Workflow Execution spawned from Continue-As-New has the same Workflow Id, a new Run Id, and a fresh Event History and is passed all the appropriate parameters.
 For example, it may be reasonable to use Continue-As-New once per day for a long-running Workflow Execution that is generating a large Event History.
+
+### Limits
```
MichaelSnowden, line 130: We should provide the users with more information on what these limits do and why. Something like this:

> When executing Commands, Temporal may consume resources that it only releases later. For example, the ScheduleActivityTask Command adds an entry to the Workflow's pending Activities. Temporal removes this entry only when the activity terminates. As a result, having too many pending Activities may impede progress on the Workflow that created them.
> 
> To avoid this, whenever Temporal executes a Workflow Task that contains a ScheduleActivityTask Command, which would cause the Workflow to exceed its limit of pending activities, Temporal fails that Workflow Task Execution. It then adds a WorkflowTaskFailed event to the Workflow's history, containing a cause of WORKFLOW_TASK_FAILED_CAUSE_PENDING_ACTIVITIES_LIMIT_EXCEEDED within its failure field.
> 
> Temporal retries Workflow Tasks, so if a sufficient number of pending activities terminate, the Workflow Task to schedule more Activities may eventually succeed. However, suppose the Workflow Task contains Commands that exceed this limit even if there are no pending Activities when the Command executes. In that case, the Workflow Task will eventually fail.
> 
> Temporal enforces this constraint similarly for the SignalExternalWorkflowExecution, RequestCancelExternalWorkflowExecution, and StartChildWorkflowExecution Commands. For each Workflow, the number of pending Activities, outgoing Signals, cancel requests to external Workflows, and Child Workflows that it may contain are each limited to 50,000. For Temporal Cloud customers, this limit is 2,000.

docs/concepts/what-is-an-event-history.md
```diff
@@ -15,6 +15,6 @@ An append-log of [Events](/concepts/what-is-an-event) for your application.
 
 **Event History limits**
 The Temporal Cluster stores the complete Event History for the entire lifecycle of a Workflow Execution.
-There is a hard limit of 50,000 Events in a Workflow Execution Event History, as well as a hard limit of 50 MB in terms of size.
+[There is a hard limit of 50,000 Events](/concepts/what-is-a-workflow-execution/#limits) in a Workflow Execution Event History, as well as a hard limit of 50 MB in terms of size.
```
MichaelSnowden, line 18: I'm worried about us stating the exact limit in so many places. Because we may update it soon, I think we should instead say that the limit exists and link to the main limits page.

MichaelSnowden (changes requested):
We should give users more information on what these limits do, how they know they're violating them, and why they exist. I added a comment with a more concrete suggestion.

DJSanti:
> 

Okay, I like this a lot. I'll tweak it some for brevity, but this is great, thank you.

docs-src/references/commands.md
```diff
@@ -50,27 +50,39 @@ This Command is triggered by a call to spawn a [Child Workflow Execution](/conce
 - Awaitable: Yes, a Workflow Execution can await on the action resulting from this Command.
 - Corresponding Event: [ChildWorkflowExecutionStarted](/references/events/#childworkflowexecutionstarted)
 
+OSS users cannot have more than 50,000 pending Child Workflows.
```
MichaelSnowden, line 53: We should use commas everywhere or nowhere

docs/concepts/what-is-a-workflow-execution.md
MichaelSnowden: Your edits look great. Much simpler way of phrasing it, thank you!

docs/workflows.md
```diff
@@ -254,6 +254,22 @@ You can use the <a class="tdlp" href="#continue-as-new">Continue-As-New<span cla
...
+
+:::note
+
+Cloud users have a hard limit of 2,000 pending Events.
+
+:::
+
+In addition to the hard limits imposed on Event Histories, certain Commands will not be processed when there are already 50,000 pending Events of a given type:
+
+- `StartTimer`
```
MichaelSnowden, line 267: Not enforced

kb/temporal-platform-limits-sheet.md
```diff
@@ -26,5 +26,10 @@ Here is a list of many hard (error) or soft (warn) limits that you could encount
 - **History total count limit** (leading to a terminated Workflow Execution):
   - We warn at 10,000 Events: `history size exceeds warn limit.`
   - We error at 50,000 Events: [`history size exceeds error limit.`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/workflowExecutionContext.go#L1204)
+  - We fail the following Commands at 50,000 Events for OSS (Cloud will fail at 2,000 Events):
```
MichaelSnowden, line 29: Commands is more precise than Event here.

MichaelSnowden (commented):
Some minor nits

DJSanti:
> Some minor nits

gotcha. after this, is it good for the stamp?

docs/concepts/what-is-an-event-history.md
```diff
@@ -15,6 +15,6 @@ An append-log of [Events](/concepts/what-is-an-event) for your application.
 
 **Event History limits**
 The Temporal Cluster stores the complete Event History for the entire lifecycle of a Workflow Execution.
-There is a hard limit of 50,000 Events in a Workflow Execution Event History, as well as a hard limit of 50 MB in terms of size.
```
MichaelSnowden, line 18: This should remain unchanged. It's a different limit that happens to use the same number.

docs/workflows.md
```diff
@@ -254,6 +254,22 @@ You can use the <a class="tdlp" href="#continue-as-new">Continue-As-New<span cla
...
+
+For example, each execution of `ScheduleActivityCommand` adds an Activity to the Task Queue, along with the resources needed to execute it.
+However, scheduling too many Activities can take resources away from what's currently running.
+
+To avoid this scenario, Temporal fails Workflow Task Executions that would cause the Workflow to surpass 50,000 pending Activities.
+Similar constraints are enforced for `SignalExternalWorkflowExecution`, `RequestCancelExternalWorkflowExecution`, and `StartChildWorkflowExecution` Commands.
+
+:::note
+
+Cloud users are limited to 2,000 pending Activities, Child Workflows, and external Workflows.
```
MichaelSnowden, line 269: - and Signals
- should say each

docs/workflows.md
```diff
@@ -286,7 +302,7 @@ An append-log of <a class="tdlp" href="#event">Events<span class="tdlpiw"><img s
 
 **Event History limits**
 The Temporal Cluster stores the complete Event History for the entire lifecycle of a Workflow Execution.
-There is a hard limit of 50,000 Events in a Workflow Execution Event History, as well as a hard limit of 50 MB in terms of size.
+[There is a hard limit on Events](/concepts/what-is-a-workflow-execution/#limits) in a Workflow Execution Event History, as well as a hard limit of 50 MB in terms of size.
```
MichaelSnowden, line 305: This should also remain unchanged because it's a separate limit (event history limit vs pending commands limit). They happen to use the same number, though.

kb/temporal-platform-limits-sheet.md
```diff
@@ -26,5 +26,10 @@ Here is a list of many hard (error) or soft (warn) limits that you could encount
 - **History total count limit** (leading to a terminated Workflow Execution):
   - We warn at 10,000 Events: `history size exceeds warn limit.`
   - We error at 50,000 Events: [`history size exceeds error limit.`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/workflowExecutionContext.go#L1204)
+  - We fail the following Commands at 50,000 for OSS (Cloud will fail at 2,000):
+    - `ScheduleActivityTask`
+    - `SignalExternalWorkflowExecution`
+    - `RequestCancelExternalWorkflowExecution`
+    - `StartChildWorkflowExecution`
   - This is configurable with [`HistoryCountLimitError` and `HistoryCountLimitWarn`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/configs/config.go#L382-L383), if you know what you are doing.
```
MichaelSnowden, line 34: This comment should be right below line 28

MichaelSnowden: It's referring to the event history limit, not the pending commands limit

MichaelSnowden (commented):
After comments, this looks good as far as I'm concerned, but I think @yiminc should also take a look.

docs/concepts/what-is-a-workflow-execution.md
```diff
@@ -126,3 +126,19 @@ A Workflow Execution Timeout can be used to limit the duration of Workflow Execu
...
+
+When executing Commands, Temporal allocates resources that get released once a Workflow Task has completed or terminated.
+
+For example, each execution of `ScheduleActivityCommand` adds an Activity to the Task Queue, along with the resources needed to execute it.
+However, scheduling too many Activities can take resources away from what's currently running.
+
+To avoid this scenario, Temporal fails Workflow Task Executions that would cause the Workflow to surpass 50,000 pending Activities.
+Similar constraints are enforced for `SignalExternalWorkflowExecution`, `RequestCancelExternalWorkflowExecution`, and `StartChildWorkflowExecution` Commands.
+
```
yiminc, line 139: This is not accurate.

The correct one would be something like this:
Each pending activity (activity scheduled but not completed yet) will generate metadata entry in workflow mutable state. Too many such entries will result in mutable state become too big. At large scale this will result in persistence become unstable. To protect the system, Temporal enforce max pending activity limit. If incoming command would lead to pending activity exceed that limit, temporal server will reject the command causing the workflow task to fail.  By default the limit is 50000 which is essentially no limit because it is the same as event count limit. But we recommend setting some reasonable limit (for example 1000) to protect the system.

yiminc (changes requested):
The 50K is default OSS limit because we don't want to break any existing use case. But we recommend OSS self-hosted server operator to apply reasonable limit, for example 1000 to protect the system. This could potentially break existing use case, so be cautions when apply the limit.

docs-src/references/commands.md
```diff
@@ -50,27 +50,39 @@ This Command is triggered by a call to spawn a [Child Workflow Execution](/conce
...
+Cloud users can have up to 2,000 pending Child Workflows.
```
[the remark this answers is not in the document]
DJSanti: Removing Cloud limits for now.

DJSanti:
@yiminc about to push my edits. Let me know if it's good to go, and stamp it if you can.

docs-src/references/commands.md
```diff
@@ -50,27 +50,35 @@ This Command is triggered by a call to spawn a [Child Workflow Execution](/conce
 - Awaitable: Yes, a Workflow Execution can await on the action resulting from this Command.
 - Corresponding Event: [ChildWorkflowExecutionStarted](/references/events/#childworkflowexecutionstarted)
 
+By default, OSS users cannot have more than 50,000 pending Child Workflows.
```
yiminc, line 53: We should mention that operator of temporal cluster could adjust this limit via dynamic config, and list the config key that controls this limit.

docs/concepts/what-is-a-workflow-execution.md
```diff
@@ -126,3 +126,21 @@ A Workflow Execution Timeout can be used to limit the duration of Workflow Execu
 You can use the [Continue-As-New](/concepts/what-is-continue-as-new) feature to close the current Workflow Execution and create a new Workflow Execution in a single atomic operation.
 The Workflow Execution spawned from Continue-As-New has the same Workflow Id, a new Run Id, and a fresh Event History and is passed all the appropriate parameters.
 For example, it may be reasonable to use Continue-As-New once per day for a long-running Workflow Execution that is generating a large Event History.
+
+### Limits
+
+When executing Commands, Temporal allocates resources that get released once a Workflow Task has completed or terminated.
```
yiminc, line 132: This is not accurate. Should remove this line.

docs/concepts/what-is-a-workflow-execution.md
```diff
@@ -126,3 +126,21 @@ A Workflow Execution Timeout can be used to limit the duration of Workflow Execu
...
+
+Each pending Activity generates a metadata entry in the Workflow's mutable state.
+Too many entries create a large mutable state, which causes unstable persistence.
+
+To protect the system, Temporal enforces a maximum pending Activity limit.
+By default, Temporal fails Workflow Task Executions that would cause the Workflow to surpass 50,000 pending Activities.
```
yiminc, line 138: should list the dynamic config keys that controls the limits.

yiminc (approved):

docs/concepts/what-is-a-workflow-execution.md
```diff
@@ -126,3 +126,19 @@ A Workflow Execution Timeout can be used to limit the duration of Workflow Execu
...
+
+Similar constraints are enforced for `SignalExternalWorkflowExecution`, `RequestCancelExternalWorkflowExecution`, and `StartChildWorkflowExecution` Commands.
+
+:::note
+
+Cloud users are limited to 2,000 each of: pending Activities, Child Workflows, external Workflows, and Signals.
```
djmagee, line 142:
```suggestion
Cloud users are limited to 2,000 each of pending Activities, Child Workflows, external Workflows, and Signals.
```

docs/concepts/what-is-an-event-history.md
```diff
@@ -14,7 +14,10 @@ An append-log of [Events](/concepts/what-is-an-event) for your application.
 - It also serves as an audit log for debugging.
 
 **Event History limits**
+
 The Temporal Cluster stores the complete Event History for the entire lifecycle of a Workflow Execution.
-There is a hard limit of 50,000 Events in a Workflow Execution Event History, as well as a hard limit of 50 MB in terms of size.
+
+[There is a hard limit of 50,000 Events](/concepts/what-is-a-workflow-execution/#limits) in a Workflow Execution Event History, as well as a hard limit of 50 MB in terms of size.
```
djmagee, line 20:
```suggestion
A Workflow Execution Event History has [a hard limit of 50,000 Events](/concepts/what-is-a-workflow-execution/#limits), as well as a hard limit of 50 MB in terms of size.
```

kb/temporal-platform-limits-sheet.md
```diff
@@ -24,7 +24,16 @@ Here is a list of many hard (error) or soft (warn) limits that you could encount
   - We error at 50 MB: [`history size exceeds error limit.`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/workflowExecutionContext.go#L1204)
   - This is configurable with [`HistorySizeLimitError` and `HistorySizeLimitWarn`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/configs/config.go#L380-L381), if you know what you are doing.
 - **History total count limit** (leading to a terminated Workflow Execution):
+
   - We warn at 10,000 Events: `history size exceeds warn limit.`
   - We error at 50,000 Events: [`history size exceeds error limit.`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/workflowExecutionContext.go#L1204)
+
   - This is configurable with [`HistoryCountLimitError` and `HistoryCountLimitWarn`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/configs/config.go#L382-L383), if you know what you are doing.
+
+  - We fail the following Commands at 50,000 for OSS (Cloud will fail at 2,000):
```
djmagee, line 33:
```suggestion
  - We fail the following Commands at 50,000 for OSS and at 2,000 for Temporal Cloud:
```

docs-src/references/commands.md
MichaelSnowden: Yes, and you can find the key names here https://www.notion.so/temporalio/Workflow-Size-Limits-877a2b2d4f74457d9533f0f976ca442a in the table under "Dynamic Config Key". Don't mention proto field name, though.

MichaelSnowden (approved):
LGTM after Yimin's comment. We need to make operators aware of how they can configure these limits, and that they should do so.

docs/concepts/what-is-a-workflow-execution.md
```diff
@@ -126,3 +126,25 @@ A Workflow Execution Timeout can be used to limit the duration of Workflow Execu
...
 The Workflow Execution spawned from Continue-As-New has the same Workflow Id, a new Run Id, and a fresh Event History and is passed all the appropriate parameters.
 For example, it may be reasonable to use Continue-As-New once per day for a long-running Workflow Execution that is generating a large Event History.
+
+### Limits
+
+Each pending Activity generates a metadata entry in the Workflow's mutable state.
+Too many entries create a large mutable state, which causes unstable persistence.
+
+To protect the system, Temporal enforces a maximum of 50,000 pending Activities, Child Workflows, external Workflows and Signals.
+These limits are set with the following [dynamic configuration keys](https://github.com/temporalio/temporal/blob/53f46157728224a4f62b66d474724a346d7c8405/service/history/configs/config.go):
```
djmagee, line 136:
```suggestion
These limits are set with the following [dynamic configuration keys](https://github.com/temporalio/temporal/blob/master/service/history/configs/config.go):
```

djmagee (commented):
Two tweaks in what-is-a-workflow-execution.md and you're good to go.

djmagee (approved):

PR #1820, opened 2022-11-30
Clarify workflow versioning

laniehei:
## What does this PR do?
There has been questions while making PRs from several reviewers, citing this document and stating that the docs say to start with "1" as the default. The Versioning schema starts at "-1" and is 0 by default. Simplifying the docs, because the currently imply that users should start with 1 as the default version, and to 2, instead of 0 to 1. 

To clarify, this is not a bug in the documentation, but this does cause unnecessary confusion.

### DIFF
--- a/docs/go/versioning.md
+++ b/docs/go/versioning.md
@@ -66,7 +66,7 @@ If you make an additional change, such as replacing ActivityC with ActivityD, yo
 add some additional code:
 
 ```go
-v := workflow.GetVersion(ctx, "Step1", workflow.DefaultVersion, 2)
+v := workflow.GetVersion(ctx, "Step1", workflow.DefaultVersion, 1)
 if v == workflow.DefaultVersion {
         err = workflow.ExecuteActivity(ctx, ActivityA, data).Get(ctx, &result1)
 } else if v == 1 {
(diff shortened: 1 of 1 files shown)


flossypurse (approved):
I would also make these changes to the file in this repo: https://github.com/temporalio/legacy-documentation-sdks

These are duplicate legacy pages.
The ones in this repo will stop being used in the near future - but the ones in the legacy repo will stay published.

So, if you update here - then update there too , or these updates will be lost when we stop indexing this file.

cretz:
Can also update the docs at https://pkg.go.dev/go.temporal.io/sdk/workflow#GetVersion

lorensr:
Moved this to:

https://github.com/temporalio/legacy-documentation-sdks/pull/4/
https://github.com/temporalio/sdk-go/pull/987

@flossypurse are the files ready to be deleted from this repo?

flossypurse:
> @flossypurse are the files ready to be deleted from this repo?

Awesome - @lorensr we won't actually do a hard delete on these files in the docs repo (at least for now). But we will stop indexing them so that you can not access them anymore from site: https://github.com/temporalio/documentation/pull/1846

And that change is pretty much ready to go, waiting on a final stamp from @bergundy

PR #1822, opened 2022-12-01
TW 186: TLS 1.3 not supported

DJSanti:
Added a remark about not supporting TLS v 1.3


djmagee:
@DJSanti: I'm strongly inclined to move the admonition to the top of `certificates-requirements.md` to join the rest of the requirements (and yet still be above the fold on most screens).

djmagee (approved):

PR #1835, opened 2022-12-08
TW-201 Corrects Parent Close Policy default

aarohib:
## What does this PR do?
Fixes the incorrect default listed on the [how to set a parent close policy in Go](https://docs.temporal.io/go/how-to-set-a-parent-close-policy-in-go) and updates the default in the Java topic. Also updated the app-dev-context for parent close policy.


docs-src/app-dev-context/parent-close-policy.md
```diff
@@ -8,3 +8,5 @@ tags:
 ---
 
 A [Parent Close Policy](/concepts/what-is-a-parent-close-policy) determines what happens to a Child Workflow Execution if its Parent changes to a Closed status (Completed, Failed, or Timed Out).
+
+The default Parent Close Policy option is set to `TERMINATE`.
```
djmagee, line 12: Because the option values differ across the SDKs, I recommend using prose—perhaps something like the following:

The default Parent Close Policy is to terminate the Child Workflow Execution.

aarohib: Updated to "The default Parent Close Policy option is set to terminate the Child Workflow Execution."

djmagee (approved):

flossypurse (approved):

PR #1844, opened 2022-12-14
rename Typescript to Go

martin-daisy:
Wrongly title said Typescript where it should've been Go

That's it 😃


flossypurse:
Looks like someone else caught this and put up a duplicate - https://github.com/temporalio/documentation/pull/1870

We just happened to deal with the other one first.

PR #1847, opened 2022-12-16
TS: Add hints about recurrent support questions

mjameswh:
## What changed

- We still recommend Node 16 for production deployments, but there are no longer any outstanding issues with Node 18 (we're adding to test matrix at this moment)
- When bundling TS SDK workers with docker, add `node_modules` to `.gitignore`
- Debugger breakpoints in workflow code doesn't work correctly when running in Jest.
- How to add support for "TS paths aliases" in workflow bundler


lorensr:
These source files are no longer used—the URLs are redirected to a different domain. I moved this PR to that domain's repo: https://github.com/temporalio/legacy-documentation-sdks/pull/3

PR #1848, opened 2022-12-22
Use `npm init` for TS SDK

lorensr:


flossypurse (approved):

bergundy (commented):
I'm debating whether we want to recommend this method at this point as it requires npm >= 8.
It's nicer syntax but I'm afraid of breaking the getting started experience.

flossypurse (changes requested):
Per @bergundy comment - need to assess.

PR #1851, opened 2022-12-25
Fix blog redirect

lorensr:


lorensr:
Rules are evaluated in order, so this now redirects correctly:

https://temporal-documentation-k31gz0s8c.preview.thundergun.io/blog/typescript-1.0.0

to 

https://temporal.io/blog/typescript-1-0-0

flossypurse (approved):

PR #1852, opened 2022-12-27
Fix register in the "Registering multiple Types"

phmx:
## What does this PR do?

Fix register in methods' names in the "Registering multiple Types" code sample.


flossypurse (approved):
Nice catch! Thank you!

Issue #1369, opened 2022-08-04
[ Documentation request ] Document Workflow Id constraints


rachfop:
# Brief description
Customer asked:
> is there a some restrictions on how a workflow id could be named? (special characters, UTF-8 symbols, etc) 

Provide a description of your documentation request.
Answer:
> If you are using Cassandra, the max length limit is 1000. If you are using SQL, the max limit is 255.
The string must be valid utf-8.

Also, [`MaxIDLengthLimit`](https://github.com/temporalio/temporal/blob/7f480149b213b649cf19ca7b8a48ffbe132d829a/service/frontend/service.go#L178) == 1000
#  Your recommended content

[Place it here?](https://github.com/temporalio/documentation/blob/master/docs/concepts/what-is-a-workflow-id.md)

> The Workflow Id is limited to 1000 characters for Cassandra databases and 255 characters for SQL databases.
> The string must be valid UTF-8.

Add your recommended content here. Add the content directly to this page or add links to drafts.
Documentation requests that are made without any recommendations could take longer to process.

rachfop:
Closing as it is tracked internally or completed.

PR #1859, opened 2022-12-30
Add Identifiers to limits sheet

lorensr:
I'm inclined to also add a sentence to the bottom of each relevant glossary page? see #1369


kb/temporal-platform-limits-sheet.md
```diff
@@ -13,11 +13,14 @@ Here is a list of many hard (error) or soft (warn) limits that you could encount
 
 <!-- truncate -->
 
+- **Identifiers**: By default, the maximum length for identifiers (like Workflow Id, Workflow Type, and Task Queue name) is 1000 characters.
+  - This is configurable with the `limit.maxIDLength` dynamic config, which is set to 255 in [our SQL example](https://github.com/temporalio/docker-compose/blob/93d382ef9133e4cde8ce311de5153cd0cc9fbd0c/dynamicconfig/development-sql.yaml#L1-L2).
```
flossypurse, line 17: Not fan of linking to lines in source - seen these change a lot.
Maybe link to overall file?
In the not so distant future we should be sporting a Dynamic Config ref will some of these knobs on there.

flossypurse (approved):
overall lgtm

kb/temporal-platform-limits-sheet.md
lorensr: This link is to a specific commit so that it won't change.

PR #1861, opened 2023-01-03
Fix and add links between metrics pages

lorensr:


docs-src/references/sdk-metrics.md
```diff
@@ -7,10 +7,12 @@ tags:
   - reference
 ---
 
-> For Cluster metrics, see [Cluster ▶️ Production deployment ▶️ Scaling and Metrics](/server/production-deployment/#scaling-and-metrics).
```
lorensr, line 10: Do we have all this info elsewhere, or should I link to this in addition? 

https://docs.temporal.io/server/production-deployment#scaling-and-metrics

flossypurse: Keep that link for now.
But moving this into and expanding it in the Deployment guide is a top priority for us these next 2 Quarters.

flossypurse (approved):

docs-src/references/sdk-metrics.md
```diff
@@ -7,10 +7,12 @@ tags:
...
-> For Cloud metrics, see [Temporal Cloud ▶️ Cloud metrics](/cloud/how-to-monitor-temporal-cloud-metrics).
+> See also: [References ▶️ Cluster metrics](/references/cluster-metrics).
 
 The Temporal SDKs emit a set of metrics from Temporal Client usage and Worker Processes.
+To set up metrics, see [Dev guide ▶️ Observability ▶️ Metrics](/application-development/observability#metrics).
```
flossypurse, line 13:
```suggestion
To set up metrics, see the [Metrics section of the Dev guide](/app-dev-context/metrics).
```

That section is built by Assembly so we can do this above and it will rewrite with a link preview

docs-src/references/sdk-metrics.md
flossypurse: Actually on closer inspection, much of this is captured in the Cluster metrics reference - 
And if it is not - then it should be captured there - or in the Cluster deployment guide
We want to disperse the info in that file into the appropriate places in the new guides.

PR #1862, opened 2023-01-03
Update TS connection snippet

lorensr:


docs-src/typescript/how-to-connect-to-a-cluster-in-typescript.md
```diff
@@ -21,33 +21,40 @@ Use a new `WorkflowClient()` with the requisite gRPC [`Connection`](https://type
 Use the [`connectionOptions`](https://typescript.temporal.io/api/interfaces/client.TLSConfig) API to connect a Client with mTLS.
 
 ```typescript
-import fs from "fs-extra";
-import {Connection, WorkflowClient} from "@temporalio/client";
-import path = from "path";
+import { Connection, WorkflowClient } from '@temporalio/client';
```
flossypurse, line 24: Can we put this into the samples-typescript repo as part of a sample and snipsync it in?

bergundy: We should also use `Client` now instead of `WorkflowClient`

PR #1863, opened 2023-01-03
Add shortlinks to API references

lorensr:


flossypurse:
Hmm... not sure how I feel about this tbh.
Screen Shot 2023-01-03 at 2 03 32 PM

lorensr:
I think they're helpful to know / people appreciate knowing them, and this seemed a place most likely to be seen. A possible alternative for all but Java is displaying the short link at the index of the API reference sites

flossypurse (changes requested):
The general consensus is that this clutters the UX.
How about we put these in each of the lang api-reference nodes like this one: https://github.com/temporalio/documentation/blob/main/docs-src/go/api-reference-go.md

And then those will publish to the dev guide here: https://temporal-documentation-93but6zbt.preview.thundergun.io/application-development/foundations#api-reference

flossypurse (approved):

PR #1875, opened 2023-01-09
Adds Replays

rachfop:
## What does this PR do?

Adds the Concept Replays to WFE and Glossary page.

## Notes to reviewers

Added SVG for now - but it isn't being used. Will create a seperate PR to add the SVG, add an example, and write about it.

### DIFF
--- a/docs/application-development/foundations.md
+++ b/docs/application-development/foundations.md
@@ -218,7 +218,7 @@ pip install temporalio
 
 This project requires Node.js 14.18 or later.
 
-**Create a new project**
+**Create a project**
 
 ```bash
 npx @temporalio/create@latest ./your-app


docs-src/concepts/what-is-a-workflow-execution.md
```diff
@@ -38,6 +38,16 @@ Scalability is responsiveness in the presence of load.
 A single Workflow Execution is limited in size and throughput but is scalable because it can [Continue-As-New](/concepts/what-is-continue-as-new) in response to load.
 A Temporal Application is scalable because the Temporal Platform is capable of supporting millions to billions of Workflow Executions executing concurrently, which is realized by the design and nature of the [Temporal Cluster](/concepts/what-is-a-temporal-cluster) and [Worker Processes](/concepts/what-is-a-worker-process).
 
+### Replays
+
+A Replay is effectively a retry of the Workflow Execution that checks against an existing Event History. Replays are necessary and often happen to give the effect that Workflow Executions are resumable, reliable, and durable.
```
flossypurse, line 43: I would open with:

> A Replay is the method by which a Workflow Execution resumes making progress.
> During a Replay the Commands that are generated are checked against an existing Event History..

And then here https://docs.temporal.io/workflows#deterministic-constraints at the end of the sentence:

>The execution semantics of a Workflow Execution include the re-execution of a Workflow Function.

Add ", this is called a Replay." and link to the Replay section.

docs-src/concepts/what-is-a-workflow-execution.md
```diff
@@ -38,6 +38,16 @@ Scalability is responsiveness in the presence of load.
...
+
+If a failure occurs, the Workflow Execution picks up where the last recorded event occurred in the Event History.
```
flossypurse, line 45: Interestingly, the explanation of determinism basically explains Replays:

> The execution semantics of a Workflow Execution include the re-execution of a Workflow Function. The use of Workflow APIs in the function is what generates [Commands](https://docs.temporal.io/workflows#command)Link preview icon. Commands tell the Cluster which Events to create and add to the Workflow Execution's Event History. When a Workflow Function executes, the Commands that are emitted are compared with the existing Event History. If a corresponding Event already exists within the Event History that maps to the generation of that Command in the same sequence, and some specific metadata of that Command matches with some specific metadata of the Event, then the Function Execution progresses.

docs-src/concepts/what-is-a-workflow-execution.md
```diff
@@ -38,6 +38,16 @@ Scalability is responsiveness in the presence of load.
 A single Workflow Execution is limited in size and throughput but is scalable because it can [Continue-As-New](/concepts/what-is-continue-as-new) in response to load.
 A Temporal Application is scalable because the Temporal Platform is capable of supporting millions to billions of Workflow Executions executing concurrently, which is realized by the design and nature of the [Temporal Cluster](/concepts/what-is-a-temporal-cluster) and [Worker Processes](/concepts/what-is-a-worker-process).
 
+### Replays
```
cretz, line 41: This is a bit confusing wording.

The term "replay" in Temporal means a workflow execution replay, but how you're using "replays" here is specifically related to "replayers".

It is important to differentiate these two. Replays happen all the time in Temporal whether a user sees it or not. The part that you're referring to here is the "Replayer" which is a manually triggered replay I think.

If you are meaning the general concept of a "replay", then I don't think this is a good definition:

> A Replay is effectively a retry of the Workflow Execution that checks against an existing Event History.

Specifically, "retry" is a bad term here. Maybe "rerun" of "code". Also, that it checks against history is true but not necessarily the definition of a replay. Maybe there's some other way to make it clear that the effects of the code are not applied if those effects are already present in history and we check to make sure the effects match.

static/diagrams/replays.svg
```diff
@@ -0,0 +1 @@
...2)"><path fill="black" stroke="none" paint-order="stroke fill markers" d=" M 584.0072014775002 645.6174965800002 L 568.3002752275002 653.4709597050002 L 584.0072014775002 661.3244228300001 Z"/></g></g><g><g><rect fill="white" stroke="none" x="422.97500314312515" y="681.7434269550001" width="81.32099735625" height="24.50280495"/></g><text fill="black" stroke="none" font-family="sans-serif" font-size="11pt" font-style="normal" font-weight="normal" text-decoration="normal" x="425.80224986812516" y="698.7069073050002" text-anchor="start" dominant-baseline="alphabetic" xml:space="preserve">return("ok")</text></g><g><path fill="none" stroke="black" paint-order="fill stroke markers" d=" M 568.3002752275002 706.2462319050002 L 374.4891715500001 706.2462319050002" stroke-miterlimit="10" stroke-width="1.5706926250000002" stroke-dasharray="7.5393246000000005"/><g transform="translate(358.9707284150001,706.2462319050002) translate(-358.9707284150001,-706.2462319050002)"><path fill="black" stroke="none" paint-order="stroke fill markers" d=" M 374.6776546650001 698.3927687800002 L 358.9707284150001 706.2462319050002 L 374.6776546650001 714.0996950300001 Z"/></g></g></g><g/><g/><g/><g/></g></svg>
```
cretz, line 1: Where is this used?

cretz: Agreed version control for the win, but you can just keep in a branch on your fork

docs-src/typescript/clients.md
```diff
@@ -41,13 +41,15 @@ import { example } from './workflows';
 
 async function run() {
   // Connect to the default Server location (localhost:7233)
+  console.log(Date.now());
```
cretz, line 44: Third PR where this same auto-sync'd stuff is coming in. Maybe you need another PR that puts this stuff in main instead of each of these?

docs-src/app-dev-context/workflow-run-timeout.md
```diff
@@ -7,4 +7,4 @@ tags:
   - guide-context
 ---
 
-Use the [Workflow Run Timeout](/concepts/what-is-a-workflow-run-timeout) to restrict the maximum amount of time that a single [Workflow Run](/concepts/what-is-a-workflow-execution#workflow-execution-chain) can last.
+Use the [Workflow Run Timeout](/concepts/what-is-a-workflow-run-timeout) [Workflow Run](/concepts/what-is-a-workflow-execution#workflow-execution-chain) can last.
```
flossypurse, line 10: This looks like an unintended change?

docs-src/concepts/what-is-a-workflow-definition.md
```diff
@@ -23,7 +23,7 @@ We strongly recommend that you write a Workflow Definition in a language that ha
 
 A critical aspect of developing Workflow Definitions is ensuring they exhibit certain deterministic traits – that is, making sure that the same Commands are emitted in the same sequence, whenever a corresponding Workflow Function Execution (instance of the Function Definition) is re-executed.
 
-The execution semantics of a Workflow Execution include the re-execution of a Workflow Function.
+The execution semantics of a Workflow Execution include the re-execution of a Workflow Function, which is called a [Replay](/concepts/what-is-a-workflow-execution#replays).
```
flossypurse, line 26: 👍

docs-src/concepts/what-is-a-workflow-execution.md
```diff
@@ -38,6 +38,16 @@ Scalability is responsiveness in the presence of load.
...
+
+A Replay is the method by which a Workflow Execution resumes making progress. During a Replay the Commands that are generated are checked against an existing Event History. Replays are necessary and often happen to give the effect that Workflow Executions are resumable, reliable, and durable.
+
+If a failure occurs, the Workflow Execution picks up where the last recorded event occurred in the Event History.
+
+To accomplish this, the Replay retries the Workflow Execution function and expects a response. If the corresponding response was recorded in the Event History, the Replay process continues to the next function. If a response wasn't recorded in the Event History, the Replay _retries_ that function and records the response in the Event History. This process repeats often.
+
+You can use Replay APIs from any of the Temporal SDKs to control this behavior.
```
flossypurse, line 49:
```suggestion
- [How to use Replay APIs to test Workflow Definitions](/app-dev-context/replays).
```

I think you can drop this explanation and focus on linking off to the "how to".

I think the way "response" is used to describe things here is too generic and so it doesn't explain much.
The "deterministic constraints" section does this anyways - and so linking to it above should be enough.

flossypurse (approved):
maybe remove the svg if it is not being used?

otherwise - this is looking good to me -

Issue #1740, opened 2022-10-28
[ Bug report ] Jumping anchor


lorensr:
## Required

**What is the issue?**

When I visit 

https://docs.temporal.io/application-development/foundations#connect-to-a-cluster

the correct heading flashes, but then I see a lower down section of the page.

**What is the expected behavior?**

It stays on the anchor heading

## Additional context (optional—the more we know, the faster it goes)

**Screenshots:**

https://user-images.githubusercontent.com/251288/198422871-d3344656-fa3e-46a9-b4bf-5be4c96ca234.mov

**Browser and version:**

Latest Mac Chrome

djmagee:
Confirmed on Firefox 106.0.2 and Edge 106.0.1370.52, both on Windows 10.0.19045.2130.

feedmeapples:
same for https://docs.temporal.io/references/web-ui-environment-variables#temporal_notify_on_new_version

vkarpov15:
It looks like this is due to link previews. The link previews show up for an instance before disappearing, and that's causing the layout shift. For example, you can see that https://docs.temporal.io/temporal#temporal-platform shows up correctly, but with https://docs.temporal.io/temporal#temporal-application the header shifts up a bit out of view. I haven't been able to figure out why this is happening or how to fix yet, but I'm working on it.

axfelix:
This was fixed without being closed (see above links); closing.

PR #1891, opened 2023-01-16
Proof of concept for non-jumping anchors re: #1740

vkarpov15:
## What does this PR do?

A quick example of a potential fix for the jumping anchors mentioned in #1740.

Before: https://www.loom.com/share/e64a6f15472b4b24b42c044e683ec8cf
After: https://www.loom.com/share/fd31788c541e4839ab8e9e79bd7c2bf4

This was an extremely tricky issue to debug. I tried just about everything I could find to get Chrome to not render the link preview on initial load: `content-visibility: hidden`, `height: 0px; width: 0px;`, `position: absolute` off-screen, etc. Through a bunch of trial and error, I found this issue is likely caused by the fact that putting a `<div>` and `<p>` in a `<span>` is considered invalid HTML, see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#conceptual_differences), [StackOverflow 1](https://stackoverflow.com/questions/3921228/link-inside-span-with-display-none-style-is-visible), [StackOverflow 2](https://stackoverflow.com/questions/42705927/why-is-it-causing-an-error-when-i-put-a-p-inside-of-a-span).

With a couple of minor semantic changes (replacing `<p>` and `<div>` with `<span>`, using `<br>` for line breaks) we can get the same visual effect, but without causing layout shift due to invalid HTML. Here's what the link preview looks like now:

image

Here's what the link preview looks like with the suggested changes:

image

Note that I made this change for only 2 link previews, and in a generated file. That is for the purpose of soliciting feedback - if you're OK with the general direction of this change, I'm happy to apply it throughout the codebase, I'll just need a little guidance for running all the linkMagic code.


lorensr (approved):
Nice find! Lgtm. Where are classes from? Wondering if you can use css over br

vkarpov15:
I'll close this for now in favor of #1892 . That PR applies the changes in this PR to all link previews.

PR #1892, opened 2023-01-17
converting divs and ps to spans to create valid html for previews

flossypurse:
Per @vkarpov15 research and debugging on https://github.com/temporalio/documentation/pull/1891 - this PR updates link magic to use spans and newline breaks to adhere to valid HTML rules.

This could potentially fix issues with anchors jumping after page loads.


vkarpov15 (approved):
Nice :+1: I did some tinkering in the Vercel preview, and it looks like anchor links work properly again. Here's a couple of examples:

image

image

Thanks for implementing this!

flossypurse:
@vkarpov15 
> Thanks for implementing this!

R u kidding me!? Thank YOU for spending the time tracking down this very annoying and obscure problem that I introduced!

It does appear to be a fix - so let's roll it out and see!!!

djmagee (approved):

PR #1900, opened 2023-01-20
Add shared logger interface example to TypeScript logging section

vkarpov15:
## What does this PR do?

Pulls in example from https://github.com/temporalio/samples-typescript/pull/228 to demonstrate how to use a shared logger interface, so you can import the same logger file from both Workflows and Activities.

**PREVIEW:** https://temporal-documentation-c487w9adt.preview.thundergun.io/application-development/observability?lang=ts#logging

### DIFF
--- a/docs-src/typescript/logging.md
+++ b/docs-src/typescript/logging.md
@@ -229,6 +229,65 @@ Some important features of the [InjectedSinkFunction](https://typescript.tempora
 - **Limited arguments types**: The remaining Sink function arguments are copied between the sandbox and the Node.js environment using the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
 - **No return value**: To prevent breaking determinism, Sink functions cannot return values to the Workflow.
 
+#### Shared logger interface
+
+Instead of explicitly calling `proxySinks()` to create a logger sink in your Workflow, you can also create a `sharedLogger.ts` file that handles calling `proxySinks()` for you.
+
+<!--SNIPSTART typescript-shared-logger-->
+
+[logger-shared/src/sharedLogger.ts](https://github.com/temporalio/samples-typescript/blob/master/logger-shared/src/sharedLogger.ts)
+
+```ts
+import { proxySinks, inWorkflowContext } from '@temporalio/workflow';
+import logger from './logger';
+
+const sharedLogger = inWorkflowContext() ?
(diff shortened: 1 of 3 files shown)


docs-src/typescript/logging.md
```diff
@@ -229,6 +229,64 @@ Some important features of the [InjectedSinkFunction](https://typescript.tempora
 - **Limited arguments types**: The remaining Sink function arguments are copied between the sandbox and the Node.js environment using the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
 - **No return value**: To prevent breaking determinism, Sink functions cannot return values to the Workflow.
 
+#### Shared logger interface
```
flossypurse, line 232: I think this might be targeting the legacy docs set?
https://legacy-documentation-sdks.temporal.io/typescript/logging

that content is now published through another repo -
Can add it here if you want but it won't show up published.

lorensr: Could you add Val to that repo? (or make public)

docs/application-development/observability.md
```diff
@@ -502,6 +502,64 @@ Some important features of the [InjectedSinkFunction](https://typescript.tempora
 - **Limited arguments types**: The remaining Sink function arguments are copied between the sandbox and the Node.js environment using the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).
 - **No return value**: To prevent breaking determinism, Sink functions cannot return values to the Workflow.
 
+**Shared logger interface**
+
+Instead of explicitly calling `proxySinks()` to create a logger sink in your Workflow, you can also create a `sharedLogger.ts` file that handles calling `proxySinks()` for you.
```
lorensr, line 507:
```suggestion
If you want to call the same logger in your Workflow and Activity code, then you can only use `proxySinks()` if the code is running in Workflow context:
```

lorensr (dismissed):

docs-src/typescript/logging.md
vkarpov15: It looks like I do have access to that repo, I'll make the changes there.

vkarpov15:
Merging this PR is blocked on https://github.com/temporalio/samples-typescript/pull/232

flossypurse:
Implemented in https://github.com/temporalio/documentation/pull/2545

PR #1901, opened 2023-01-21
TW-148: Update certificate requirements to specify encryption key types

djmagee:
## What does this PR do?

Specifies RSA and ECDSA as the valid algorithms for signing certificates.


djmagee:
Null comment to test Jira integration

PR #1908, opened 2023-01-27
Adds link to WIP

rachfop:
## What does this PR do?

All text that starts with, "Content is currently unavailable." will be linked to: `https://docs.temporal.io/application-development#under-development`.

This is due to users being confused about why some content is missing.

## Notes to reviewers

We have:
`none`: Content is currently unavailable.
`na`: Not applicable to this SDK.

This is to clarify that `none` is just a work in progress.


djmagee:
The link target doesn't exclusively show the admonition, so the intent might be lost on the user. Perhaps we should simply rewrite the text—something like "Content is planned but not yet available."

assembly/activities/genMarkdownGuides.js
```diff
@@ -71,15 +71,15 @@ function ssdi(ssdi) {
...
   tabStr = `${tabStr}defaultValue="go"\n`;
   tabStr = `${tabStr}groupId="site-lang"\n`;
   tabStr = `${tabStr}values={[{label: 'Go', value: 'go'},{label: 'Java', value: 'java'},{label: 'PHP', value: 'php'},{label: 'Python', value: 'python'},{label: 'TypeScript', value: 'typescript'},]}>\n\n`;
   for (const tab of langtabs) {
     tabStr = `${tabStr}<TabItem value="${tab.lang}">\n\n`;
-    if (tab.id == "none") {
-      tabStr = `${tabStr}Content is currently unavailable.\n\n`;
-    } else if (tab.id == "na") {
+    if (tab.id == 'none') {
+      tabStr = `${tabStr}Content is planned but not yet available.\n\n`;
```
flossypurse, line 81: Could take this just a bit farther and add a link to the legacy docs site with something like "the information you are looking for may be found in the [legacy docs](https://legacy-documentation-sdks.temporal.io/)"

Could create a `const` at the top of this file and make this into a string literal would be the easiest i -

```
const plannedText = `Content is planned but not yet available.

The information you are looking for may be found in the [legacy docs](https://legacy-documentation-sdks.temporal.io/).`
```
And then right here you would just do:

```
tabStr = `${tabStr}${plannedText}\n\n`;
```

flossypurse (approved):
Nice!

PR #1909, opened 2023-01-28
Added links to Go installation prerequisite

efortuna:
## What does this PR do?
Makes the mention of Go version required to be a link so that the user can click on it and download Go.

## Notes to reviewers

### DIFF
--- a/docs-src/clusters/quick-install.md
+++ b/docs-src/clusters/quick-install.md
@@ -22,7 +22,7 @@ It supports persistence to disk and in-memory mode through SQLite.
 
 **Prerequisites**
 
-Temporalite requires Go 1.18 or later.
+Temporalite requires [Go 1.18](https://go.dev/dl/) or later.
 
 **Build and start Temporalite**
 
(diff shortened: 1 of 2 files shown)


docs-src/clusters/quick-install.md
```diff
@@ -22,7 +22,7 @@ It supports persistence to disk and in-memory mode through SQLite.
 
 **Prerequisites**
 
-Temporalite requires Go 1.18 or later.
+Temporalite requires [Go 1.18](https://go.dev/dl/) or later.
```
djmagee, line 25:
```suggestion
Temporalite requires [Go 1.18 or later](https://go.dev/dl/).
```
Because Go is already up to 1.19.5, I recommend including "or later" in the link text to set expectations appropriately (and to simplify maintenance for us).

djmagee (commented):
Excellent idea. And props for noting the separate source and generated files!

djmagee (approved):

PR #1926, opened 2023-02-07
Advanced Visibility with SQL updates for v1.20

aarohib:
## What does this PR do?
Updates docs on Advanced Visibility support for SQL dbs with server v 1.20

Previews:
- [Visibility](https://temporal-documentation-l537m6y64.preview.thundergun.io/visibility)
- [Setting up your Visibility store](https://temporal-documentation-l537m6y64.preview.thundergun.io/cluster-deployment-guide#visibility-store)
- [Upgrading your DB schema for MySQL and PostgreSQL](https://temporal-documentation-l537m6y64.preview.thundergun.io/cluster-deployment-guide#upgrade-postgresql-or-mysql-schema)

PR #1963, opened 2023-03-14
SQLite is supported

lorensr:


docs-src/concepts/what-is-a-temporal-cluster.md
```diff
@@ -50,8 +50,7 @@ The release notes of each Temporal Server declare when we plan to drop support f
   Temporal has no opinions on database upgrade paths; as long as you can upgrade your database according to each project's specifications, Temporal should work with any version within supported ranges.
 - We do not run tests with vendors like Vitess and CockroachDB, so you rely on their compatibility claims if you use them.
   Feel free to discuss them with fellow users [in our forum](https://community.temporal.io/).
-- Temporal is [working on official SQLite v3.x persistence](https://github.com/temporalio/temporal/pulls?q=is%3Apr+sort%3Aupdated-desc+sqlite), but this is meant only for development and testing, not production usage.
-  Cassandra, MySQL, and PostgreSQL schemas are supported and thus can be used as the Server's database.
```
lorensr, line 54: Don't understand what this means

dnr (approved):
Yeah, that reads funny. The new text is fine.

Actually, can you delete the line "Because Temporal Server primarily relies on core database functionality, we do not expect compatibility to break often."? Advanced visibility in SQL uses much fancier stuff so I would say that's no longer true

flossypurse:
We have a fairly large update going out very soon that has info on how to setup and use SQLite 
https://github.com/temporalio/documentation/pull/1926

Let's wait on 1963 until it 1926 lands.

aarohib (approved):

PR #1967, opened 2023-03-15
CLI release messaging

flossypurse:


docs-src/concepts/what-is-the-temporal-cli.md
```diff
@@ -1,17 +1,17 @@
 ---
-id: what-is-temporal-cli
-title: What is Temporal CLI?
-description: Temporal CLI is the most recent version of Temporal's command-line tool.
+id: what-is-the-temporal-cli
+title: What is the Temporal CLI?
+description: The Temporal CLI is the most recent version of Temporal's command-line tool.
 sidebar_label: cli
```
djmagee, line 5: Should this topic now appear in the Glossary as "Temporal CLI" instead of "cli" (which appears at the bottom of the list)?

PR #1997, opened 2023-04-13
TW 389: deadline exceeded command link

DJSanti:
## What does this PR do?
Adds a link to the grpc repository when the command is mentioned.

## Notes to reviewers
L38 should be where it is.


docs/cli/index.md
tomwheeler, line 1: This change seems to be unrelated to TW-389, although I agree with the change since Homebrew is available for Linux as well as macOS.

tomwheeler (approved):
I looked at the deployment on Vercel and can confirm that the link I requested in the issue is now present and working properly. 

As noted in the review, there seems to be another change unrelated to TW-389, but it corrects a minor inaccurate detail, so I support publishing it too.

PR #2008, opened 2023-04-28
EDU-179 Adds ORDER BY limit to Temporal Cloud section of limits sheet

flossypurse:
This PR adds MVP information to the Temporal Platform limits sheet that indicates that the ORDER BY clause is not supported in List Filters in Temporal Cloud.


docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -29,6 +29,7 @@ A List Filter contains [Search Attribute](/concepts/what-is-a-search-attribute)
 - **ORDER BY**
 
 The **ORDER BY** operator is supported only when Elasticsearch is used as the Visibility store.
+This means that the **ORDER BY** operator is currently not supported in Temporal Cloud.
```
cretz, line 32: Not needed for this PR, but I wonder if we want to just remove (or bury into a spot at the bottom of this page) anything about order by if it's not supported in cloud.

cretz (approved):

kb/temporal-platform-limits-sheet.md
```diff
@@ -45,9 +45,11 @@ Hard limits fail with an error; soft limits produce a warning log on the server
     - `limit.numPendingSignals.error`
     - `limit.numPendingCancelRequests.error`
     - `limit.numPendingChildExecutions.error`
-- [Search Attributes maximums](/visibility/#search-attributes-maximums)
+- [Search Attributes maximums](visibility/#custom-search-attributes-limits)
```
djmagee, line 48:
```suggestion
- [Search Attributes maximums](/visibility/#custom-search-attributes-limits)
```
Without the leading slash, the link renders as `/kb/visibility/#custom-search-attributes-limits`.

kb/temporal-platform-limits-sheet.md
```diff
@@ -64,3 +66,8 @@ Hard limits fail with an error; soft limits produce a warning log on the server
     - keyword: 20
     - text: 5
   - Retention Period: 30 days (configurable; range of 1–90 days)
+- **List Filters**
+  - The **ORDER BY** operator is not supported in List Filters in Temporal Cloud.
+    This means that you can not apply custom ordering of Workflows with Cloud Visibility features.
```
djmagee, line 71:
```suggestion
    This means that you cannot apply custom ordering of Workflows with Cloud Visibility features.
```

kb/temporal-platform-limits-sheet.md
```diff
@@ -64,3 +66,8 @@ Hard limits fail with an error; soft limits produce a warning log on the server
...
+    Lists of Workflows are still ordered by a default ordering rule.
+    Though, this rule may change in the future.
```
djmagee, line 73:
```suggestion
    Be aware that this rule might change.
```
All change that hasn't yet happened occurs in the future. —Spock*

*Absolutely not Spock. Created by DailGPT.

docs-src/concepts/what-is-a-list-filter.md
flossypurse: @cretz Ya, I have been thinking about how to more broadly indicate which features / things are supported across OSS vs Cloud.

If you have any suggestions - lets talk!

cretz: No suggestions in particular beyond burying/de-emphasizing the non-cloud features rather than burying/de-emphasizing a statement that a feature is non-cloud. But that's more general, PR LGTM but may still want a server team approver.

lorensr, line 32: This implies that Cloud doesn't use ES

flossypurse: > This implies that Cloud doesn't use ES

Good catch.

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -29,7 +29,10 @@ A List Filter contains [Search Attribute](/concepts/what-is-a-search-attribute)
 - **ORDER BY**
 
 The **ORDER BY** operator is supported only when Elasticsearch is used as the Visibility store.
```
rodrigozhou, line 31: For the next minor release, we're gonna disable this feature by default in OSS. It's controlled via dynamic config.
cc: @yiminc

rodrigozhou (approved):

kb/temporal-platform-limits-sheet.md
```diff
@@ -64,3 +66,8 @@ Hard limits fail with an error; soft limits produce a warning log on the server
     - keyword: 20
     - text: 5
   - Retention Period: 30 days (configurable; range of 1–90 days)
+- **List Filters**
+  - The **ORDER BY** operator is not supported in List Filters in Temporal Cloud.
+    This means that you cannot apply custom ordering of Workflows with Cloud Visibility features.
+    Lists of Workflows are still ordered by a default ordering rule.
```
rodrigozhou, line 72: Should we specify what's the order?

flossypurse: @rodrigozhou I don't think it is described anywhere. But I also understand that it is subject to change at any time... which is called in below.
Shall we cap it here and add it if it is seems relevant to do so in the future?

docs-src/concepts/what-is-a-list-filter.md
flossypurse: @rodrigozhou @yiminc Can you make a note wherever you are tracking that release to update the docs on this point when it becomes relevant to do so?

kb/temporal-platform-limits-sheet.md
rodrigozhou: If it's not described anywhere, I think it's fine to not to say it here.
Not sure if it's good to mention the specific order or not given that it's subject to change.

aarohib (approved):

PR #2012, opened 2023-05-03
Fix incorrect doc about NDE wrt timer's duration

mjameswh:
## What does this PR do?

- Provide more exact details on changes to a Timer's duration would cause NDE.


lorensr (approved):

docs-src/concepts/what-is-a-workflow-definition.md
```diff
@@ -67,7 +67,9 @@ The Workflow Execution would fail and return a non-deterministic error.
 
 The following are examples of minor changes that would not result in non-determinism errors when re-executing a History which already contain the Events:
 
-- Changing the duration of a Timer (unless changing from a duration of 0).
+- Changing the duration of a Timer, with the following exceptions:
+  - In Java, Python and Go, changing a timer’s duration from or to 0 is a non-deterministic behaviour;
+  - In .Net, changing a timer’s duration from or to -1 (which means ‘infinite’) is a non-deterministic behaviour.
```
cretz, line 72:
```suggestion
  - In .NET, changing a timer’s duration from or to -1 (which means ‘infinite’) is a non-deterministic behaviour.
```

cretz, line 72:
```suggestion
  - In Java, Python and Go, changing a timer’s duration from or to 0 is a non-deterministic behavior;
  - In .Net, changing a timer’s duration from or to -1 (which means ‘infinite’) is a non-deterministic behavior.
```

Too Canadian! (for reals though, did a quick check on docs and we seem to prefer the American spelling)

cretz (approved):

docs-src/concepts/what-is-a-workflow-definition.md
mjameswh: Lol! Yeah, I know. OSX's spell checker is unfortunately fixing those automatically to Canadian spelling, so I regularly get "colour", "centre", "behaviour"... I need to fix that.

flossypurse (approved):

PR #2015, opened 2023-05-03
EDU-412: Define Action

djmagee:
## What does this PR do?

Adds the definition of Action and surfaces it on the Temporal Cloud documentation landing page.


docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+- Signal-With-Start Workflow Execution.
+- Continue-As-New Workflow Execution.
+- Start Child Workflow Execution.
+- Upsert Workflow Search Attributes.
+
+**Activity Actions**
+
+- Schedule Activity Task.
+- Activity Retry. Each retry is counted as a separate Action.
+- Record Marker. A Workflow [Worker](/concepts/what-is-a-worker) executed a [Local Activity](/concepts/what-is-a-local-activity) or a [Side Effect](/concepts/what-is-a-side-effect) or is using [Workflow versioning](/workflows#workflow-versioning).
```
flossypurse, line 28: I think having Workflow Worker in the description is a bit too verbose here.

Record Marker Events are associated with a Local Activity Execution and/or a Side Effect - I get that - but how does it have to do with Workflow Versioning?
And if I go to that link... where is that aspect described?
I don't see anything related to Record Markers in the versioning info.

flossypurse (commented):
Looks like we need to add Cloud index generated file to this list:
https://github.com/temporalio/documentation/blob/main/.gitattributes

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Signal Actions**
```
cretz, line 30: Need a section for update.

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+- Signal Workflow Execution.
+- Signal External Workflow Execution.
+
+**Query Actions**
+
+- Query Workflow.
+
+**Timer Actions**
+
+- Start Timer.
```
cretz, line 41: May need to clarify for users that things that use timeouts (e.g. `AwaitWithTimeout` in Go, giving a timeout for `condition` in TS, etc), those start a timer internally.

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+Each Action is an [Event](/concepts/what-is-an-event) in the [Event History](/concepts/what-is-an-event-history) of a [Workflow Execution](/concepts/what-is-a-workflow-execution).
+(Be aware that not all Events are Actions.)
+
+**Workflow Actions**
+
+- Start Workflow Execution.
+- Reset Workflow Execution.
+- Signal-With-Start Workflow Execution.
+- Continue-As-New Workflow Execution.
+- Start Child Workflow Execution.
```
cretz, line 21: This can be confusing to readers to include with "Start Workflow Execution". Don't we just charge for "Start Workflow Execution" regardless of whether it's a child or not?

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+- Upsert Workflow Search Attributes.
+
+**Activity Actions**
+
+- Schedule Activity Task.
+- Activity Retry. Each retry is counted as a separate Action.
```
cretz, line 27: I am a bit confused here. So just scheduling an activity that may never run is an action? But it's first attempt is not an action but each attempt after the first is? One might expect for simplicity each activity attempt is an action.

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+- Record Marker. A Workflow [Worker](/concepts/what-is-a-worker) executed a [Local Activity](/concepts/what-is-a-local-activity) or a [Side Effect](/concepts/what-is-a-side-effect) or is using [Workflow versioning](/workflows#workflow-versioning).
+
+**Signal Actions**
+
+- Signal Workflow Execution.
+- Signal External Workflow Execution.
```
cretz, line 33: Should they really have to care about the genesis of the signal? Can't we just say all signals charge?

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Query Actions**
+
+- Query Workflow.
```
cretz, line 37: Are we providing a way to tell users today how many queries they have sent to a workflow? All the other actions I can validate via history, but not this one.

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+---
+
+An Action is the fundamental pricing unit in [Temporal Cloud](/concepts/what-is-temporal-cloud).
+Each Action is an [Event](/concepts/what-is-an-event) in the [Event History](/concepts/what-is-an-event-history) of a [Workflow Execution](/concepts/what-is-a-workflow-execution).
+(Be aware that not all Events are Actions.)
+
+**Workflow Actions**
+
+- Start Workflow Execution.
+- Reset Workflow Execution.
```
cretz, line 18: Probably obvious, but can I get confirmation that the actions that happened before reset are still counted? This is another one that's hard to quantify afterwards because the actions are gone.

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+- Signal-With-Start Workflow Execution.
+- Continue-As-New Workflow Execution.
```
cretz, line 20: So is continue as new a double charge with start workflow execution? Maybe to make things clear it should be "Start Workflow Execution - can come from client as start, from client as signal with start, from continue as new, or from child workflow start".

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+
+**Activity Actions**
+
+- Schedule Activity Task.
+- Activity Retry. Each retry is counted as a separate Action.
+- Record Marker. A Workflow [Worker](/concepts/what-is-a-worker) executed a [Local Activity](/concepts/what-is-a-local-activity) or a [Side Effect](/concepts/what-is-a-side-effect) or is using [Workflow versioning](/workflows#workflow-versioning).
+
+**Signal Actions**
+
+- Signal Workflow Execution.
```
cretz, line 32: To confirm, this is an intentional separate charge from signal with start?

docs-src/concepts/what-is-an-action.md
cretz, line 28: So if I want to save some money on retries, I should use local activities since we only record marker on complete? @Sushisource - I thought there was some logic where we're also counting/recording local activity retries, correct?

docs-src/concepts/what-is-an-action.md
cretz, line 37: It can be unclear to users that if they do so much as click a stack trace tab in the UI they will be charged. Or are we excluding stack trace queries?

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,46 @@
...
+  - explanation
+---
+
+An Action is the fundamental pricing unit in [Temporal Cloud](/concepts/what-is-temporal-cloud).
+Each Action is an [Event](/concepts/what-is-an-event) in the [Event History](/concepts/what-is-an-event-history) of a [Workflow Execution](/concepts/what-is-a-workflow-execution).
+(Be aware that not all Events are Actions.)
+
+**Workflow Actions**
+
+- Start Workflow Execution.
```
cretz, line 17: It can be unclear to users using sessions in Go that we run session handling as an activity

cretz (commented):
I am making my comments off-PR

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,33 @@
...
+  - explanation
+---
+
+An Action is the fundamental pricing unit in [Temporal Cloud](/concepts/what-is-temporal-cloud).
+
+**Workflows**
+
+- **Workflow started:** Occurs via client start, client Signal-With-Start, [Continue-As-New](/concepts/what-is-continue-as-new), [Child Workflow](/concepts/what-is-a-child-workflow-execution) start, or [Schedule](/concepts/what-is-a-schedule).
+  If a Workflow fails to start, an Action is not recorded.
+- **Workflow reset:** Actions that occurred before a [Reset](/concepts/what-is-a-reset) are counted even if they are no longer visible in [Event History](/concepts/what-is-an-event-history).
```
lorensr, line 17: I didn't realize "Workflow reset" was a type of action here, because the first bullet primed me to think:

- [context]: list of actions

and this feels like:

- [context]: no listed actions, just an FYI note.

I think my issue might be fixed if all bullets had a period > colon. Then I'm more likely to read as:

- [action]. explanatory text / notes

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,33 @@
...
+- **Workflow cancellation requested:** Includes requests from a client or from within a Workflow to another Workflow.
+- **Timer started:** Includes implicit Timers that are started by a Temporal SDK when timeouts are set, such as `AwaitWithTimeout` in Go or `condition` in TypeScript.
+- **Search Attribute updated:** Occurs after a Workflow has started.
+- **Signal sent:** Includes sending a [Signal](/concepts/what-is-a-signal) from a client or from within a [Workflow](/concepts/what-is-a-workflow) to another Workflow.
+- **Query received:** [Queries](/concepts/what-is-a-query) aren't recorded in Event History.
+  A step such as viewing the stack trace in the Temporal Cloud UI results in a Query.
+- **Side Effect recorded:** For a mutable [Side Effect](/concepts/what-is-a-side-effect), an Action occurs only when the value changes.
+  (Some SDKs don't support Side Effects.)
```
lorensr, line 25:
```suggestion
  (Note that some SDKs don't support Side Effects.)
```

as written, I feel like it's trying to explain what came before, which doesn't make sense

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,33 @@
...
+An Action is the fundamental pricing unit in [Temporal Cloud](/concepts/what-is-temporal-cloud).
+
+**Workflows**
+
+- **Workflow started:** Occurs via client start, client Signal-With-Start, [Continue-As-New](/concepts/what-is-continue-as-new), [Child Workflow](/concepts/what-is-a-child-workflow-execution) start, or [Schedule](/concepts/what-is-a-schedule).
+  If a Workflow fails to start, an Action is not recorded.
+- **Workflow reset:** Actions that occurred before a [Reset](/concepts/what-is-a-reset) are counted even if they are no longer visible in [Event History](/concepts/what-is-an-event-history).
+- **Workflow cancellation requested:** Includes requests from a client or from within a Workflow to another Workflow.
+- **Timer started:** Includes implicit Timers that are started by a Temporal SDK when timeouts are set, such as `AwaitWithTimeout` in Go or `condition` in TypeScript.
+- **Search Attribute updated:** Occurs after a Workflow has started.
```
lorensr, line 20:
```suggestion
- **Search Attribute updated:** Occurs after a Workflow has started, when Workflow code calls `UpsertSearchAttributes`.
```

docs-src/concepts/what-is-an-action.md
```diff
@@ -0,0 +1,33 @@
...
+- **Query received:** [Queries](/concepts/what-is-a-query) aren't recorded in Event History.
+  A step such as viewing the stack trace in the Temporal Cloud UI results in a Query.
+- **Side Effect recorded:** For a mutable [Side Effect](/concepts/what-is-a-side-effect), an Action occurs only when the value changes.
+  (Some SDKs don't support Side Effects.)
+
+**Activities**
+
+- **Activity attempted:** Each [Activity](/concepts/what-is-an-activity) retry is an attempt and thus an Action.
+- **Activity Heartbeat recorded:** Not every SDK call [Activity Heartbeat](/concepts/what-is-an-activity-heartbeat) counts as an Action.
+  Temporal SDKs throttle Heartbeats.
```
lorensr, line 31:
```suggestion
- **Activity Heartbeat recorded:** Not every heartbeat call from Activity code counts as an Action—only those that reach the Server.
  Temporal SDKs throttle [Activity Heartbeats](/concepts/what-is-an-activity-heartbeat).
```

lorensr (approved):

docs-src/concepts/what-is-an-action.md
djmagee: Excellent point. I'll make that change.

DJSanti (approved):
Looks good to me

PR #2018, opened 2023-05-04
Amend ActivityTaskFailed description

rachfop:
## What does this PR do?

Change `completed` to `failed`.
- EDU 421


cretz (approved):

tomwheeler (commented):
I agree with the corrected sentence, but would like to suggest a change to the one that follows it, which begins with "The SDK picked up the Activity Task..." 

I think it would be more clear to say "The SDK client picked up the Activity Task" just as the explanation above this one does.

djmagee (approved):

PR #2026, opened 2023-05-08
Updates Worker Task with TS

rachfop:
## What does this PR do?

Adds TypeScript as reported by user.
- EDU-592


docs-src/references/sdk-metrics.md
```diff
@@ -316,7 +316,7 @@ Use the `worker_type` key to differentiate execution slots.
 (Workflow Workers execute Workflow Tasks; Activity Workers execute Activity Tasks.)
 
 - Type: Gauge
-- Available in: Go, PHP, Java
+- Available in: Go, Java, PHP, TypeScript
```
mjameswh, line 319: That metrics come from Core, so it will also be present in Python and .NET.

PR #2027, opened 2023-05-08
Updates lang for Core based SDKs

rachfop:
## What does this PR do?

- EDU-593

### DIFF
--- a/docs-src/concepts/what-is-a-workflow-id-reuse-policy.md
+++ b/docs-src/concepts/what-is-a-workflow-id-reuse-policy.md
@@ -10,8 +10,17 @@ tags:
 
 A Workflow Id Reuse Policy determines whether a Workflow Execution is allowed to spawn with a particular Workflow Id, if that Workflow Id has been used with a previous, and now Closed, Workflow Execution.
 
-It is not possible for a new Workflow Execution to spawn with the same Workflow Id as another Open Workflow Execution.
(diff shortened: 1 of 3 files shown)


mjameswh (approved):

docs-src/references/sdk-metrics.md
```diff
@@ -26,8 +26,11 @@ All metrics are prefixed with `temporal_` before being exported to their configu
 (The prefix has been removed in the following reference.)
 Currently, some metrics are specific to certain SDKs.
 
+TypeScript, Python, .NET, and Ruby SDKs metrics are defined in the Core SDK.
```
flossypurse, line 29:
```suggestion
TypeScript, Python, .NET, and Ruby SDKs metrics are defined in the Rust Core SDK.
```

djmagee: We don't. This is more of a general grammar question...and a bit philosophical. Using "defined by" overlaps with "defined as," such as "defined by its compact size," whereas "defined in" implies that the direct object is where you can find the definition.

The Rust Core SDK defines a lot more than just the SDK metrics, so in this instance using "defined by" seems to be overly broad. I recommend stetting (keeping "defined in").

Having said all that, the lines are fuzzy enough that I'm not gonna advocate strongly either way.

djmagee (approved):

flossypurse (approved):

PR #2035, opened 2023-05-13
EDU-417: define State Transition

djmagee:
## What does this PR do?

- Adds the definition of State Transition
- Links two existing instances of the term to the definition


docs-src/concepts/what-is-a-state-transition.md
```diff
@@ -0,0 +1,16 @@
...
+  - term
+  - explanation
+---
+
+A State Transition is a unit of progress made by a [Workflow Execution](/concepts/what-is-a-workflow-execution).
+Each State Transition is recorded in a persistence store.
+
+Some operations, such as [Activity Heartbeats](/concepts/what-is-an-activity-heartbeat), require only one State Transition.
+Most operations require multiple State Transitions.
+For example, a simple Workflow with two [Activity Tasks](/concepts/what-is-an-activity-task) (and no retries) might need as few as 11 State Transitions.
```
lorensr, line 16:
```suggestion
For example, a simple Workflow with two [Activity Tasks](/concepts/what-is-an-activity-task) (and no retries) might need as many as 11 State Transitions.
```

"as few" feels like trying to say it's just a small amount. coming from "only one per ST", I expect "as many"

djmagee: I was thinking of retries, but because I specified no retries, the count should indeed be 11. Thanks for the catch!

docs-src/concepts/what-is-a-state-transition.md
```diff
@@ -0,0 +1,16 @@
...
+  - term
+  - explanation
+---
+
+A State Transition is a unit of progress made by a [Workflow Execution](/concepts/what-is-a-workflow-execution).
+Each State Transition is recorded in a persistence store.
+
+Some operations, such as [Activity Heartbeats](/concepts/what-is-an-activity-heartbeat), require only one State Transition.
+Most operations require multiple State Transitions.
+For example, a simple Workflow with two [Activity Tasks](/concepts/what-is-an-activity-task) (and no retries) produces 11 State Transitions.
```
wxing1292, line 16: `two [Activity Tasks]` -> `two sequential [Activity Tasks]`

wxing1292 (approved):

docs-src/concepts/what-is-a-worker-process.md
```diff
@@ -21,7 +21,7 @@ More formally, a Worker Process is any process that implements the Task Queue Pr
 
 **Worker Processes are external to a Temporal Cluster.**
 Temporal Application developers are responsible for developing [Worker Programs](/concepts/what-is-a-worker-program) and operating Worker Processes.
-Said another way, the [Temporal Cluster](/concepts/what-is-a-temporal-cluster) (including the Temporal Cloud) doesn't execute any of your code (Workflow and Activity Definitions) on Temporal Cluster machines. The Cluster is solely responsible for orchestrating state transitions and providing Tasks to the next available Worker Entity.
+Said another way, the [Temporal Cluster](/concepts/what-is-a-temporal-cluster) (including the Temporal Cloud) doesn't execute any of your code (Workflow and Activity Definitions) on Temporal Cluster machines. The Cluster is solely responsible for orchestrating [State Transitions](/concepts/what-is-a-state-transition) and providing Tasks to the next available Worker Entity.
```
flossypurse, line 24:
```suggestion
Said another way, the [Temporal Cluster](/concepts/what-is-a-temporal-cluster) (including the Temporal Cloud) doesn't execute any of your code (Workflow and Activity Definitions) on Temporal Cluster machines. The Cluster is solely responsible for orchestrating [State Transitions](/concepts/what-is-a-state-transition) and providing Tasks to the next available Worker.
```
Can maybe drop "Entity"?

flossypurse (approved):
Overall, lgtm!

docs-src/concepts/what-is-a-worker-process.md
djmagee: @flossypurse: Given the number of instances of "Worker Entit[y|ies]" in this topic, I'm inclined to link this first instance to `/concepts/what-is-a-worker-entity`.

PR #2036, opened 2023-05-14
Adds space for Query received

rachfop:
## What does this PR do?

This is a stylistic change for the [rendered output](https://docs.temporal.io/cloud#action).

Changes:
```
Query received.[Queries](http://localhost:3000/workflows#query)
```

To:

```
Query received. [Queries](http://localhost:3000/workflows#query)
```
## Notes to reviewers

I had to break the format of:

```
- **definition.**
Setence
```

To have the output rendered correctly.  

- EDU-645


docs-src/concepts/what-is-an-action.md
```diff
@@ -26,8 +26,7 @@ The following operations result in Actions, which are billed monthly.
   Occurs after a Workflow starts and invokes `UpsertSearchAttributes`.
 - **Signal sent.**
   Includes sending a [Signal](/concepts/what-is-a-signal) from a client or from within a Workflow to another Workflow.
-- **Query received.**
-  [Queries](/concepts/what-is-a-query) aren't recorded in Event History.
+- **Query received.** [Queries](/concepts/what-is-a-query) aren't recorded in Event History.
```
flossypurse, line 29: Normally, we would want to see each sentence on its own line. That is part of our style guidance.
This puts both sentences on the same line.
Looks like there is an issue with the link preview css that causes that lack of a gap when it is the first word in a sentence.
This fixes the output, but breaks our style guide.
I think we can put this through, but lets track an Issue for trying to fix the link preview css.

djmagee (approved):
Nice catch. We'll hafta figure out why `/concepts/what-is-...` links at the beginning or end of a line have this behavior.

flossypurse (approved):

PR #2037, opened 2023-05-15
testing redirect

flossypurse:


vercel.json
```diff
@@ -50,243 +50,63 @@
...
-      "source": "/application-development/features:path*",
-      "has": [
-        {
-          "type": "query",
-          "key": "lang",
-          "value": "typescript"
-        }
-      ],
-      "destination": "/dev-guide/typescript/features:path*",
+      "destination": "/dev-guide/python:path*",
```
lorensr, line 97: deployment preview:

/application-development/testing#replay?lang=python

takes me to:

/dev-guide#replay?lang=python

as if it's not matching this rule

djmagee (approved):

vercel.json
flossypurse: Pretty sure that is because the query param never worked when it was after the anchor?
For example, if I recall, it would have had to have been `/application-development/testing?lang=python#replay` -
Otherwise the query param would have been ignored.
You can go to this docs page and test it out: https://docusaurus.io/docs/markdown-features/tabs#query-string
Just move the query param to the end - and you will see it doesn't work.

lorensr: Ah right, thanks. Here's anchor last, same behavior:


```
visiting:
https://docs.temporal.io/application-development/testing?lang=python#replay
goes to:
https://docs.temporal.io/dev-guide?lang=python#replay
desired:
https://docs.temporal.io/dev-guide/python/testing#replay
```

flossypurse: Ya, I swear these worked when I tested, but I may have been looking at the wrong preview.
In any case: https://github.com/temporalio/documentation/pull/2044
This should do all of it now.

PR #2043, opened 2023-05-15
Remove direction to select lang

lorensr:


flossypurse (changes requested):
Let's keep the description fields, but update them. 
We don't want to remove the description meta data because other parts of the build process rely on it (or will in the future)
We can collapse nodes (make a single getting started info node for example) so that it has a richer description if that makes it easier too.

djmagee (approved):

PR #2044, opened 2023-05-16
trying to fix redirects

flossypurse:
Actually adds and fixes the redirects.


lorensr (approved):
Works! Liked the last commit message 😄

PR #2049, opened 2023-05-18
Adding to list of generated files

flossypurse:
Most of the files in the docs folder are generated.
Making sure they are all listed so they don't open up on PRs.


aarohib (approved):

.gitattributes
```diff
@@ -1,4 +1,30 @@
 docs/cloud/index.md linguist-generated=true
+docs/cloud/how-to-create-a-ticket-for-temporal-support.md linguist-generated=true
```
djmagee, line 2: This is authored directly in `docs/cloud`. We might want to revisit that decision.

PR #2053, opened 2023-05-19
EDU-680 Glossary & tagging updates

flossypurse:
- Converts each term into an h4
- Adds descriptions and tags to terms
- Builds tags into guide front matter so that tag links work

Screenshot 2023-05-30 at 9 38 03 AM


DJSanti:
I like where this is going. The only thing that bothers me is having "term" and "explanation" all over. Is there any way to preview these tags or make them more descriptive?

djmagee:
Re: tag links. A few possibilities to consider: using CSS to surround each tag with a capsule-like background; shrinking the font size; adding an inline header (`Tags:`).

flossypurse:
> I like where this is going. The only thing that bothers me is having "term" and "explanation" all over. Is there any way to preview these tags or make them more descriptive?

@DJSanti we need to go through and update our tags across all the docs-src files.

flossypurse:
> Re: tag links. A few possibilities to consider: using CSS to surround each tag with a capsule-like background; shrinking the font size; adding an inline header (`Tags:`).

@djmagee took your advice - but used italic markdown instead to make the font different - 
LMK what you think?

djmagee:
> > Re: tag links. A few possibilities to consider: using CSS to surround each tag with a capsule-like background; shrinking the font size; adding an inline header (`Tags:`).
> 
> @djmagee took your advice - but used italic markdown instead to make the font different - LMK what you think?

Nice! Visually distinguishes the element in a lightweight manner.

docs/dev-guide/javalang/foundations.md
```diff
@@ -6,6 +6,13 @@ sidebar_position: 1
 description: The Foundations section of the Temporal Developer's guide covers the minimum set of concepts and implementation details needed to build and run a Temporal Application – that is, all the relevant steps to start a Workflow Execution that executes an Activity.
 slug: /dev-guide/java/foundations
 toc_max_heading_level: 4
+tags:
+- guide-context
+- developer-guide
+- Java
+- java
```
djmagee, line 13: Need both casings here?

flossypurse: weird.

flossypurse: actually that is because this is generated -
we should make sure all are lowercase.
but this isn't a big blocked for the moment.

djmagee (approved):

PR #2054, opened 2023-05-19
Adds Signals DACX to Docs

rachfop:
## What does this PR do?

Adds the codified version of Signals to docs.


docs-src/python/how-to-define-a-signal-in-python.md
```diff
@@ -2,54 +2,38 @@
 id: how-to-define-a-signal-in-python
 title: How to define a Signal in Python
 sidebar_label: Define a Signal
-description: Define a Signal
-tags:
-  - developer-guide
-  - sdk
-  - python
+description: Use the `@workflow.signal` decorator to define a Signal.
 ---
```
DJSanti, line 7: Maybe link to Signal concept page?

DJSanti: It feels weird to just jump in here (unless this is part of a generated document where it doesn't come first). Maybe have some transition.

docs-src/python/how-to-handle-a-signal-in-a-workflow-in-python.md
```diff
@@ -1,16 +1,19 @@
...
-  - developer-guide
-  - sdk
-  - python
+title: How to handle a Signal in Python
+sidebar_label: Handle a Signal
+description: Set the Signal on the Workflow Handle.
 ---
 
-To send a Signal to the Workflow, use the [`signal`](https://python.temporal.io/temporalio.client.workflowhandle#signal) method from the [`WorkflowHandle`](https://python.temporal.io/temporalio.client.workflowhandle) class.
+Workflows listen for Signals by the Signal's name.
```
DJSanti, line 8: "Signal Type" instead of "name" to keep with defining a Signal.

docs-src/python/how-to-send-a-signal-with-start-in-python.md
```diff
@@ -2,24 +2,24 @@
...
-description: Signal-With-Start
-tags:
-  - developer-guide
-  - sdk
-  - python
+description: To Signal-With-Start use the `start_workflow()` method and pass the `start_signal` argument with the name of your Signal.
 ---
 
-To send a Signal-With-Start in Python, use the [`start_workflow()`](https://python.temporal.io/temporalio.client.Client.html#start_workflow) method and pass the `start_signal` argument with the name of your Signal, instead of using a traditional Workflow start.
+To send a Signal-With-Start in Python, use the [`start_workflow()`](https://python.temporal.io/temporalio.client.Client.html#start_workflow) method and pass the `start_signal` argument with the name of your Signal.
```
DJSanti, line 8: The description is very similar to the start of this document. I would provide a brief explanation of why we'd do this and then get into how to do it.

DJSanti (commented):
Only have a few suggestions for transitional sentences. Give it a look-over and then ping me for approval.

docs-src/python/how-to-define-a-signal-in-python.md
```diff
@@ -2,54 +2,38 @@
...
-Your method parameters must be `self`, a string Signal name, and a `*args` variable argument parameter.
-
-```python
-@workflow.signal(dynamic=True)
-def signal_dynamic(self, name: str, *args: Any) -> None:
-    self._last_event = f"signal_dynamic {name}: {args[0]}"
-```
+- The name, also called a Signal type, is a string.
+- The arguments must be serializable.
+  To define a Signal, set the Signal decorator `@workflow.signal` on the Signal function inside your Workflow.
```
napcs, line 12: This version doesn't link to the API reference anymore. Should it?

docs-src/python/how-to-send-a-signal-with-start-in-python.md
napcs: agree.

PR #2055, opened 2023-05-22
Filter branch

rachfop:
## What does this PR do?

Need help executing the JS code to add filters.


assembly/activities/genGlossary.js
```diff
@@ -53,12 +65,24 @@ async function getTerms(config, sourceNodes) {
   }
   return terms;
 }
-
 async function genGlossString(terms) {
-  let glossStr = `${glossFrontmatter}\n\n`;
+  let glossStr = `${glossFrontmatter}\n\n<ul id="glossary-entries">`;
```
flossypurse, line 69: This doesn't necessarily need to be an ul list.
Could be h4s?
But if it is, can we remove the bullets?
I still think we should keep the tags linked and listed horizontally under the description - can style them differently though, make it smaller.
After seeing this, I don't see why we wouldn't want them there. The filter doesn't give you those details - just filters against the list.

static/scripts/filter-glossary.js
```diff
@@ -0,0 +1,23 @@
+function addFilterEvent() {
```
flossypurse, line 1: This should work if you make this into a component here: https://github.com/temporalio/documentation/tree/main/docs/components

Then import it into the glossary.md file the same we were importing Tabs for example.

PR #2029, opened 2023-05-09
Cluster configuration updates

aarohib:
## What does this PR do?
- [Updates the Clusters concept structuring](https://temporal-documentation-nu7245fhb.preview.thundergun.io/clusters)
- [Adds Cluster configuration detail to the Clusters concept](https://temporal-documentation-nu7245fhb.preview.thundergun.io/clusters#cluster-configuration)
- [Adds dynamic configuration reference](https://temporal-documentation-nu7245fhb.preview.thundergun.io/references/dynamic-configuration)
- [Fills out Cluster configuration (static) reference](https://temporal-documentation-nu7245fhb.preview.thundergun.io/references/configuration) (See `archival`, `namespaceDefaults`, `dcReplication`, `dynamicConfigClient`)
- Links between concepts and how-tos and a few other related changes to help connect the information.
- Also updates [no. of Signals limit on Temporal Cloud](https://temporal-documentation-nu7245fhb.preview.thundergun.io/kb/temporal-platform-limits-sheet#default-limits-for-temporal-cloud) ([EDU-726](https://temporalio.atlassian.net/browse/EDU-726)). Context: https://temporaltechnologies.slack.com/archives/C03AGLU5JEQ/p1686277047099249?thread_ts=1686154547.368759&cid=C03AGLU5JEQ

## Notes to reviewers
Thank you for taking the time to review this PR!
- All files under the `docs` folder are generated guides and not specific to the updates made in this PR. You can skip reviewing them. 
- Changed files that need a review in this PR are under `docs-src/clusters`, `docs-src/concepts`, `docs-src/references`, and `kb`.
- _concepts/what-is-cluster-security-configuration_ and _concepts/what-is-cluster-obervability_ are just new files that rehome currently published content and were created to reorganize the Cluster concept so the info flows better. Changes in these files are not a priority in this update (but I'm happy to make any pressing updates to incorrect info).

[EDU-726]: https://temporalio.atlassian.net/browse/EDU-726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2069, opened 2023-05-30
Use tally in sample config

yiminc:
## What does this PR do?
Use tally in the config sample for metrics.

## Notes to reviewers
Tally is still the default metrics framework. We do plan to switch to otel once the otel metrics go sdk is stable.
Later in this same page it does mention that the `tally` is the default. So I think we should use it in the config sample.


aarohib (approved):

aarohib:
Will check for other instances and update them in [PR 2029](https://github.com/temporalio/documentation/pull/2029)

PR #2070, opened 2023-05-30
EDU 264: How to version Workflows in Go

DJSanti:
## What does this PR do?
This PR takes existing information in docs-src and uses it as a source node for Go Developer Guide content.

## Notes to reviewers
This is an unchanged port of legacy content. The current focus is to migrate first, then edit later.


docs/dev-guide/golang/versioning.md
```diff
@@ -0,0 +1,168 @@
...
+Because we design for potentially long running Workflows at scale, versioning with Temporal works differently. We explain more in this optional 30 minute introduction: [https://www.youtube.com/watch?v=kkP899WxgzY](https://www.youtube.com/watch?v=kkP899WxgzY)
+
+## workflow.GetVersion()
+
+Consider the following Workflow Definition:
+
+```go
+func YourWorkflow(ctx workflow.Context, data string) (string, error) {
+        ao := workflow.ActivityOptions{
+                ScheduleToStartTimeout: time.Minute,
```
mfateev, line 29: Please remove ScheduleToStart. It the majority of cases it is harmful.

docs/dev-guide/golang/versioning.md
```diff
@@ -0,0 +1,168 @@
...
+- workflow.NewTimer()
+- workflow.RequestCancelWorkflow()
+- workflow.SideEffect()
+- workflow.SignalExternalWorkflow()
+- workflow.Sleep()
+
+Adding, removing, or reordering any of the above methods triggers the sanity check and results in
+a non-deterministic error.
+
+The sanity check does not perform a thorough check. For example, it does not check on the Activity's
```
mfateev, line 163: It sounds like activity input arguments change is non deterministic, but the check is not through. 

I think we want to say that certain changes are not considered incompatible and are safe to make without versioning. So it is about types of changes and not about checker.

docs/dev-guide/golang/versioning.md
```diff
@@ -0,0 +1,168 @@
...
+input arguments or the timer duration. If the check is enforced on every property, then it becomes
```
mfateev, line 164: Timer duration change is safe unless it changes from 0 to non zero or vice-versa.

docs/dev-guide/golang/versioning.md
```diff
@@ -0,0 +1,168 @@
...
+too restricted and harder to maintain the Workflow code. For example, if you move your Activity code
+from one package to another package, that changes the `ActivityType`, which technically becomes a different
+Activity. But, we don't want to fail on that change, so we only check the function name part of the
+`ActivityType`.
```
mfateev, line 168: Moving activity code form one package to another doesn't change the activity type as we just use the function name. So the checker passes as the name is the same.

DJSanti:
Grabbed screenshots for what to do next. Merging now.

PR #2092, opened 2023-06-08
Use Ki over K for Event warning/limit

lorensr:
Wondering what yall think. @djmagee @afitz0 

GPT4 says `k` is correct for an abbreviation for 1,000, `K` is Kelvin, and `Ki` is 1024 (albeit usually used as a prefix).

image


djmagee:
Agreed. Google style dictates lowercase "k" for the metric prefix _kilo_ (https://developers.google.com/style/units-of-measure#use-k-to-indicate-thousands), so the current usage wouldn't be correct by that measure anyway. 

And here we're speaking of the "binary thousand" (_kibi_), which, as you say, is properly abbreviated "Ki", per the IEC (https://www.electropedia.org/iev/iev.nsf/display?openform&ievref=112-01-27).

Nice catch. Thanks!

djmagee (approved):

PR #2097, opened 2023-06-11
Async signals docs

edmondop:
## What does this PR do?

Provide hints on non-blocking signal handling

### DIFF
--- a/docs/dev-guide/golang/features.md
+++ b/docs/dev-guide/golang/features.md
@@ -78,9 +78,11 @@ MySignal struct {
 
 ### Handle Signal
 
-Workflows listen for Signals by the Signal's name.
+Workflows listen for Signals by the Signal's name. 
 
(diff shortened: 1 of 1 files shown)


cretz:
This is essentially "how to read from a channel using a coroutine". It's not specific to signals, and many would be better off using a selector to read from channel while waiting for other things instead of a coroutine. I am not sure this is the best fit for docs.

edmondop:
@cretz imho, signals have two possible usages:
1. they can be used as a primitive to create wait points in the workflow, using "blocking" selectors call from the workflow functions
2. they can be used as a primitive to listen to outside changes while progressing, and by passing the workflow state to the following activity, you ensure that you take this new knowledge about the outside world into account

The way to do (2) seems to use a Temporal GoRoutine to listen for the channel, and keep this goroutine running for the entire duration of the workflow.  I am not sure whether you think that this second pattern is wrong, or not worth being in the docs (maybe just in the examples?). As a developer, I was expecting to find this in the docs

PR #2098, opened 2023-06-11
Append-only log

lorensr:
I'm not familiar with "append log" and "append-log". I believe append-only log is more commonly used and understood

image


cretz (approved):

flossypurse (approved):
TY. This change makes sense imo.

PR #2100, opened 2023-06-11
Add Temporal diagram

lorensr:
I'm suggesting it goes at top of glossary, since it has so many of the glossary terms. An alternative is in the [Temporal Platform](https://docs.temporal.io/temporal#temporal-platform) concept section, but I'm concerned it would detract from approachability so early on in the docs.

Thanks to @stephanos for creating!


static/img/temporal.png
cretz, line 1: Can you provide a preview URL? I suspect it does not look good in dark mode.

Also, you might consider an image map or links.

cretz (approved):

flossypurse:
This diagram would be great in a blog!
I would highly encourage @stephanos or you @lorensr to add this to publish this somewhere - but I don't think we are going to accept it into the docs.
Reach out to me if you want to discuss in detail.

PR #2143, opened 2023-06-27
Update to node 18, add license to package.json

lorensr:
image


.nvmrc
```diff
@@ -1 +1 @@
-16
+18
```
flossypurse, line 1: Maybe align it with what we have as a Env variable in vercel?
Not sure which overwrites the other tho -

Screenshot 2023-07-26 at 4 39 19 PM

flossypurse (approved):

.nvmrc
flossypurse: nvm - this keeps us on the latest version of 18 - nice.

flossypurse (approved):

Issue #1789, opened 2022-11-15
[ php ] Add code samples for missing content


rachfop:
The following sections need code samples.

Provide a code sample per section by using the comment section to provide a code snippet.
Mark the checkbox complete as sample code is added.

If the language does not support a feature, write “NA” or link/code to a related concept: for an example, see Go’s [Remove Search Attribute](https://docs.temporal.io/application-development/observability?lang=go#remove-search-attribute).

## Testing

- [ ] [Assert In Workflow](https://docs.temporal.io/application-development/testing?lang=php#assert-in-workflow)
- [ ] [Automatic Method](https://docs.temporal.io/application-development/testing?lang=php#automatic-method)
- [ ] [Cancel An Activity](https://docs.temporal.io/application-development/testing?lang=php#cancel-an-activity)
- [ ] [Listen To Heartbeats](https://docs.temporal.io/application-development/testing?lang=php#listen-to-heartbeats)
- [ ] [Manual Method](https://docs.temporal.io/application-development/testing?lang=php#manual-method)
- [ ] [Replay](https://docs.temporal.io/application-development/testing?lang=php#replay)
- [ ] [Run An Activity](https://docs.temporal.io/application-development/testing?lang=php#run-an-activity)
- [ ] [Setting Up](https://docs.temporal.io/application-development/testing?lang=php#setting-up)
- [ ] [Skip Time In Activities](https://docs.temporal.io/application-development/testing?lang=php#skip-time-in-activities)
- [ ] [Test Activities](https://docs.temporal.io/application-development/testing?lang=php#test-activities)
- [ ] [Test Frameworks](https://docs.temporal.io/application-development/testing?lang=php#test-frameworks)
- [ ] [Workflow Context](https://docs.temporal.io/application-development/testing?lang=php#workflow-context)

## Observability

- [ ] [Custom Logger](https://docs.temporal.io/application-development/observability#custom-logger)
- [ ] [Logging](https://docs.temporal.io/application-development/observability#logging)
- [ ] [Metrics](https://docs.temporal.io/application-development/observability#metrics)
- [ ] [Remove Search Attribute](https://docs.temporal.io/application-development/observability#remove-search-attribute)
- [ ] [Search Attributes](https://docs.temporal.io/application-development/observability#search-attributes)
- [ ] [Tracing](https://docs.temporal.io/application-development/observability#tracing)
- [ ] [Visibility](https://docs.temporal.io/application-development/observability#visibility)

## Features

[...]

PR #2145, opened 2023-06-27
SDK-PHP: Add `Features/Side Effects` block

roxblnfk:
## What does this PR do?

The PR adds `Features / Side Effects` block with code sample (see #1789)


docs/dev-guide/phplang/features.md
```diff
@@ -923,3 +923,34 @@ The following example sets a Cron Schedule in PHP:
...
+
+A Side Effect does not re-execute during a Replay. Instead, it returns the recorded result from the Workflow Execution Event History.
+
+Side Effects should not fail. An exception that is thrown from the Side Effect causes failure and retry of the current Workflow Task.
+
+An Activity or a Local Activity may also be used instead of a Side effect, as its result is also persisted in Workflow Execution History.
+
+:::note
+
+You shouldn’t modify the Workflow state inside a Side Effect function, because it is not reexecuted during Replay. Side Effect function should be used to return a value.
```
Sushisource, line 939:
```suggestion
You shouldn’t modify the Workflow state inside a Side Effect, because they are not reexecuted during Replay. Side Effect functions should only return a value, and that value may be used in Workflow code to alter state.
```

docs/dev-guide/phplang/features.md
```diff
@@ -923,3 +923,34 @@ The following example sets a Cron Schedule in PHP:
...
+
+:::
+
+To use a Side Effect in PHP, set the `Workflow::sideEffect()` function in your Workflow Execution and return the non-deterministic code.
```
Sushisource, line 943:
```suggestion
To use a Side Effect in PHP, use the `Workflow::sideEffect()` function in your Workflow Definition to run non-deterministic code and return a value.
```

Sushisource (commented):
A few wording changes. Overall structure is good :+1:

roxblnfk:
> A few wording changes

@Sushisource I copied that text from Java documentation. Should I fix the same there in another PR?

Sushisource:
@roxblnfk That would be great, or the docs folks can do it too, I'll get them to comment here.

PR #2147, opened 2023-06-27
EDU 413: Search Attribute Limits in Cloud

DJSanti:
## What does this PR do?
This PR adds a max character limit for Search Attributes.


docs-src/concepts/what-is-a-search-attribute.md
```diff
@@ -149,6 +149,8 @@ Default total maximum number of Search Attribute **keys** per Temporal Cluster i
 
 <!-- TODO - [How to configure total Search Attribute size limite](#) -->
 
+- The maximum total characters per Search Attribute is 255.
```
flossypurse, line 152: Search Attribute key or Search Attribute value?

DJSanti:
Got it checked with Candace. The limit was for values

djmagee (approved):

PR #2159, opened 2023-06-30
Add Worker Versioning TS guide

Sushisource:
Note this does not include the bit about using `VersioningIntent` with commands because I forgot to add that to the TS SDK :face_with_spiral_eyes:. I will make another PR once that exists.


docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
```diff
@@ -0,0 +1,83 @@
...
+### Assign a Build ID to your Worker
+
+Let's say you've chosen `deadbeef` as your Build ID, which might be a short git commit hash (a reasonable choice as Build ID).
+To assign it in your Worker code, assign the following Worker Options:
+
+```typescript
+// ...
+const worker = await Worker.create({
+  taskQueue: 'your_task_queue_name',
+  buildId: '1.0',
```
lorensr, line 25:
```suggestion
  buildId: 'deadbeef',
```

although i'd prefer something more pleasant like `c00fee00` or `g00df00d` or `c00lc0de` or `b0e318d2` 😄

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
```diff
@@ -0,0 +1,83 @@
...
+id: how-to-use-worker-versioning-in-typescript
+title: How to use Worker Versioning in TypeScript
+sidebar_label: Worker Versioning
+description: Version your TypeScript Workers by using build ID-based versioning
+tags:
+  - typescript
+  - how-to
+---
+
+To use Worker Versioning in TypeScript, you need to do the following:
```
lorensr, line 11:
```suggestion
To use [Worker Versioning](/workers#worker-versioning) in TypeScript, you need to do the following:
```

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
```diff
@@ -0,0 +1,83 @@
...
+// ...
+```
+
+That's all you need to do in your Worker code.
+Importantly, if you start this Worker, it won't receive any tasks.
+That's because you need to tell the Task Queue about your Worker's Build ID first.
+
+### Tell the Task Queue about your Worker's Build ID
+
+Now you can use the SDK (or the Temporal CLI) to tell the Task Queue about your Worker's Build ID.
```
lorensr, line 38: I think I'm more likely to use CLI in CI, could you show that example as well?

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Assign a Build ID to your Worker
+
+Let's say you've chosen `deadbeef` as your Build ID, which might be a short git commit hash (a reasonable choice as Build ID).
+To assign it in your Worker code, assign the following Worker Options:
+
+```typescript
+// ...
+const worker = await Worker.create({
+  taskQueue: 'your_task_queue_name',
```
lorensr, line 24:
```suggestion
  taskQueue: 'your-task-queue-name',
```

underscores feels weird in TS to me, i think all our samples use `-`

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
```diff
@@ -0,0 +1,83 @@
...
+
+```typescript
+// ...
+await client.taskQueue.updateBuildIdCompatibility('your_task_queue_name', {
+  operation: 'promoteBuildIdWithinSet',
+  buildId: 'deadbeef',
+});
+```
+
+You can also promote an entire set to become the default set for the queue. New Workflows will start using that set's default.
```
lorensr, line 75:
```suggestion
You can also promote an entire set to become the default set for the queue. New Workflows will start using that set's default build.
```

lorensr (approved):

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
Sushisource: Problem is I don't want to repeat that in each language guide. Given this contextually is about using the language SDK parts, I want to stick to that.

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
Sushisource: Gonna keep this to be consistent with the other very similar guides

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
lorensr: Could you link to a new section in the CLI docs from each lang guide that has the equivalent examples & explanatory text? 

https://docs.temporal.io/cli/task-queue/

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
```diff
@@ -0,0 +1,83 @@
...
+---
+
+To use [Worker Versioning](/workers#worker-versioning) in TypeScript, you need to do the following:
+
+1. Determine and assign a Build ID to your built Worker code, and opt in to versioning.
+2. Tell the Task Queue your Worker is listening on about that Build ID, and whether its compatible with an existing Build ID.
+
+### Assign a Build ID to your Worker
+
+Let's say you've chosen `deadbeef` as your Build ID, which might be a short git commit hash (a reasonable choice as Build ID).
```
lorensr, line 18: raises the q in my mind of how do you get the commit hash of the commit that this code is in 😄 or is the hash always the last commit before the buildId is updated & code is deployed? wonder if that pattern is worth noting

Sushisource: Yeah, it's a bit of a chicken-egg thing. You could save the hash into a file, and load that file when setting the Build ID.

docs-src/typescript/how-to-use-worker-versioning-in-typescript.md
Sushisource: Sure, but are you fine with me doing that in a subsequent PR?

lorensr: sure!

mjameswh:
This PR will need to be updated for UPPEERCASE keywords (see TS #[1156](https://github.com/temporalio/sdk-typescript/pull/1156))

lorensr (approved):

PR #2160, opened 2023-06-30
Remove unused files

lorensr:
- Went through most of the static assets (all but diagrams dir), did a text search for them in the repo, and deleted if there were no references
- Moved diagrams to diagrams folder

@flossypurse lmk if there's anything that should be added back, eg feedback/fullstory.js


flossypurse (dismissed):
woah, very nice clean up hear @lorensr !

flossypurse (approved):

PR #2161, opened 2023-07-03
fix outdated tcld and cli Schedule flags

taonic:
## What does this PR do?

Fix outdated flags as reported at:
- https://github.com/temporalio/cli/issues/156
- https://temporalio.slack.com/archives/CTRCR8RBP/p1688247842869809

## Notes to reviewers

Re: shorthand vs full flag name, I'm opting to the full flag name for clarity but happy to be convinced either way.


lorensr:
Did you do `npm run assemble`? Need to do `npm run gen` which includes `npm run format`

taonic:
> Did you do `npm run assemble`? Need to do `npm run gen` which includes `npm run format`

Thanks for the reminder @lorensr. The issue is gone after rerunning the [build](https://github.com/temporalio/documentation/actions/runs/5440719498/jobs/9893926691) that was given error below:

```
Run npx dprint fmt
npx: installed 1 in 3.879s
Cannot find module '@dprint/linux-x64-glibc/package.json'
Error: Process completed with exit code 1.
```

It may have something to do with https://github.com/dprint/dprint/pull/685

tomwheeler (approved):
I verified that all of these changes are accurate, based on the output from `temporal schedule create --help` using both versions 0.9.0 and 0.10.0 of that CLI. 

Thank you for the contribution!

PR #2163, opened 2023-07-03
Edit and add to List Filter partial string matching

lorensr:
## What does this PR do?

- Updated Text matching to be ES-agnostic (works in SQL adv vis)
- I find it easier to read these types of examples as code blocks
- Add BETWEEN option


flossypurse (approved):
Nice improvement!

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
 
 ### Partial string match
 
-The `=` operator works like **CONTAINS** to find Workflows with Search Attributes that contain a specific word.
+#### Text
```
flossypurse, line 40: Usually we try not to put a header right after another header.
Perhaps some contextualizing copy between them?

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
...
 
-<!-- note: advanced vis features will be supported in SQL upon the release of v1.20.-->
+Search Attributes of type `Text` are broken up into words that match with the `=` operator.
 
-For example, if you have a custom Search Attribute named `Description` of `Text` type with the value of "The quick brown fox jumps over the lazy dog", searching for `Description='quick'` or `Description='fox'` will successfully return the Workflow.
-However, partial word searches such as `Description='qui'` or `Description='laz'` will not return the Workflow.
-This is because [Elasticsearch's tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html) is configured to return complete words as tokens.
```
flossypurse, line 46: do we want to completely drop this example as well?
Having 2 might be good - and a link to the tokenizer is OK I think?

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
...
+For example, if you have a custom `Text` Search Attribute named `Description` with either of the following values:
```
djmagee, line 44:
```suggestion
For example, if you have a custom `Text` Search Attribute named `Description` with either of the following values—
```
This and the related following edits are an informal approach to the grammar. Another approach is to make each intro line a complete sentence—entirely valid but does add words.

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
...
+
+```
+my-business-id-foobar
+my business id foobar
+```
+
+then the following List Filter will match:
```
djmagee, line 51:
```suggestion
—then the following List Filter matches—
```

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
...
+
+```
+Description = 'foobar'
+```
+
+but a partial word will not:
```
djmagee, line 57:
```suggestion
—but a partial word does not:
```

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
...
+
+```
+// Doesn't match
+Description = 'foo'
+```
+
+#### Keyword
+
+For Search Attributes of type `Keyword` like `WorkflowId`, the only type of partial string matching that works is using BETWEEN for suffixes.
+For example:
```
djmagee, line 67:
```suggestion
For Search Attributes of type `Keyword` like `WorkflowId`, the only kind of partial string matching that works is using BETWEEN for suffixes.
```

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
...
+
+```
+WorkflowId BETWEEN "order-" AND "order-~"
+```
+
+would match WorkflowIds that have characters after `order-` with lower ASCII values than `~` (126, the highest-value printable character), such as the following:
```
djmagee, line 73:
```suggestion
The preceding example matches WorkflowIds that have characters after `order-` with ASCII values lower than `~` (126, the highest-value printable character), such as the following:
```

docs-src/concepts/what-is-a-list-filter.md
```diff
@@ -37,13 +37,48 @@ Custom Search Attributes of the `Text` type cannot be used in **ORDER BY** claus
...
+
+```
+order-
+order-1234
+order-abracadabra
+```
+
+It would not match `order-~~`.
```
djmagee, line 81:
```suggestion
It does not match `order-~~`.
```

docs-src/concepts/what-is-a-list-filter.md
lorensr: Adding the tokenizer link. I think this example is the same as my example, I just used `my-business-id-foobar` and `my business id foobar` over `The quick brown fox jumps over the lazy dog`, since it seemed more common and demonstrates tokenization more

docs-src/concepts/what-is-a-list-filter.md
lorensr: what about "This example matches"? "preceding" feels strange to me and makes me think it's referring to the one before the current example.

PR #2090, opened 2023-06-08
Update Schedules support box

lorensr:

PR #2167, opened 2023-07-05
Remove Schedules not supported in SDKs

rachfop:
## What does this PR do?

edu-893

### DIFF
--- a/docs-src/concepts/what-is-a-schedule.md
+++ b/docs-src/concepts/what-is-a-schedule.md
@@ -183,7 +183,7 @@ A Workflow started by a Schedule can obtain the details of the failure of the mo
 
 ### Limitations
 
-:::info Experimental
+:::info
 
 The Scheduled Workflows feature is available in Temporal Server version 1.18.
 
--- a/ASSEMBLY_REPORT.md
+++ b/ASSEMBLY_REPORT.md
@@ -1,6 +1,6 @@
 # Docs Assembly Workflow report
 
-Last assembled: Wednesday July 05 2023 08:39:52 AM -0700
+Last assembled: Wednesday July 05 2023 08:43:53 AM -0700
 
 Assembly Workflow Id: docs-full-assembly-rachfop-123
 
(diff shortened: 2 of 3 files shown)


docs-src/concepts/what-is-a-schedule.md
```diff
@@ -183,7 +183,7 @@ A Workflow started by a Schedule can obtain the details of the failure of the mo
 
 ### Limitations
 
-:::info Experimental
+:::info
 
 The Scheduled Workflows feature is available in Temporal Server version 1.18.
```
flossypurse, line 188: The version info is also captured by the ssdi meta data at the top of this file.

flossypurse (approved):

docs-src/concepts/what-is-a-schedule.md
mbernier: Can we have it say "Public Preview" - since that is the official status

cretz: > We need to add:

I don't think we should get in the habit of documenting SDK versions of when we made features available. That they are available on the latest SDK version should be enough info IMO.

lorensr:
https://github.com/temporalio/documentation/pull/2090/files

docs-src/concepts/what-is-a-schedule.md
lorensr: @mbernier do we have a definition of "Public Preview" we should link to?

lorensr: @cretz I can imagine if it's not there, I might read the docs and try to use a feature, and not understand why it's not working. For that case, it would be helpful to see beforehand that I need to upgrade, and to which version.

cretz: > I can imagine if it's not there, I might read the docs and try to use a feature, and not understand why it's not working.

How have you survived with any other feature Temporal has ever released ever until now then? If we want a new policy of documenting when features become available in SDKs and their versions in this non-changelog, we need to apply it generally and not here as the first application of this new policy.

> For that case, it would be helpful to see beforehand that I need to upgrade, and to which version.

You always need to upgrade. You always have needed to upgrade for new things. Schedules is not new here and we should have a general discussion about documentation of SDK versions in these docs if we're going to make a new policy here.

lorensr: 👍 will start general discussion in slack

PR #2170, opened 2023-07-05
Pr draft versioning

rachfop:
## What does this PR do?

This is just a proto-type for versioning docs.


docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+In the context of a Workflow with Version Sets, a Build ID can be assigned as the default for a Version Set or as the overall default for the entire Workflow.
+
+A Build ID can contain a maximum of 255 characters.
+
+#### What is a default Build ID?
+
+A Default Build ID is the specific Build ID that a Worker refers to when executing Tasks.
+While this is typically the most recent Build ID, you can manually specify a different Build ID as the default when registering the Workflow with a Worker.
+This option is beneficial for situations where you prefer an older Build ID to be the default rather than the latest one.
+A unique Default Build ID can be assigned to both the Task Queue and each individual Compatible Version Set.
```
Sushisource, line 60: This whole section isn't quite right. 

"you can manually specify a different Build ID as the default when registering the Workflow with a Worker." -- where did this idea come from? No such option exists. I just want to make sure we don't have misleading content somewhere else that made you think this.

There is a default for each compatible set, and there is the overall default, which is just the default of the default set - there is nothing else to consider.

When starting an activity/child/continue-as-new, you can decide if you want to stick to the compatible set of the invoking workflow, or the queue's default - that's the only place you have some kind of choice related to what version is used, and even then it's indirect.

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+If you have one Build ID, it's known as the Default Build ID.
+
+### What is a Version Set?
+
+A Version Set groups multiple Build IDs for compatibility, management, and display purposes within a Workflow.
+Each Version Set contains a designated Default Build ID, and the entire Workflow has one overall Default Build ID.
```
Sushisource, line 67: Is is not within a workflow. Sets are established on a per _Task Queue_ basis.

I would strike the word management, and display is ancillary. All that really matters is their compatibility.

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### What is a Compatible Version Set?
+
+A Compatible Version Set is a user-defined collection of Build IDs that indicates which versions of the Workflow code are mutually compatible.
+Each Compatible Version Set has a Default Build ID.
```
Sushisource, line 72: Again, not really just Workflow code (though that's the most important part). It's the whole set of code which constitutes how a Worker will handle tasks, workflow and activity.

That means workflow code, activity code, data converters, and possibly interceptors (if they mutate anything).

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+A Task Queue has a default maximum of 10 Compatible Version Sets and a maximum of 100 Build IDs for a single Task Queue across all Compatible Version Sets.
+
+You can visualize Compatible Version Sets as a tree-like structure, where a branch represents the progressive evolution of each build. Each branch in the tree is a series of versions that are compatible with each other, but not necessarily with versions in other branches.
```
Sushisource, line 76: I think this came from something I wrote in an earlier version. While it's still kinda accurate, I don't think it's really useful any longer.

A list of lists is really the simplest visualization, as in the tables in https://docs.temporal.io/workers#worker-versioning

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Task Queue-based Versioning
+
+While Task Queue-based Versioning is available to use, it's recommended to use the [Worker Versioning](#worker-versioning) approach instead, as it simplifies the process of versioning your Task Queues.
+
+Task Queue-based Versioning is the process of creating a new Task Queue for each version of your Workflow, and then modify the Task Queue the Client points to.
+
+Workers pull from Task Queues, and because of that you start Workflows on Task Queues, and each Task Queues have a unique identifier.
+
+Rather than creating a new Workflow Type for each version of your Workflow, you can create a new Task Queue for each version of your Workflow.
```
Sushisource, line 88:
```suggestion
Rather than creating a new Workflow Type for each version of your Workflow, you can create a new Task Queue for each incompatible version of your Workflow.
```

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+For example, Build ID of `1.0` might start a new branch and therefore isn't compatible with Build ID of `0.4` from an earlier branch. However, Build ID of `1.1` is compatible with Build ID of `1.0` because it's a newer version of the same branch.
+
+## Task Queue-based Versioning
+
+While Task Queue-based Versioning is available to use, it's recommended to use the [Worker Versioning](#worker-versioning) approach instead, as it simplifies the process of versioning your Task Queues.
+
+Task Queue-based Versioning is the process of creating a new Task Queue for each version of your Workflow, and then modify the Task Queue the Client points to.
+
+Workers pull from Task Queues, and because of that you start Workflows on Task Queues, and each Task Queues have a unique identifier.
```
Sushisource, line 86: This is a somewhat inverted way to look at things IMO. Workflows are tied to task queues, and workers pull from a task queue. The "because" language can probably just be dropped.

Sushisource, line 86:
```suggestion
Workers pull from Task Queues, and because of that you start Workflows on Task Queues, and each Task Queue has a unique identifier.
```

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+- **Conceptually easy**: Task Queues are similar to Workflow Type versioning, but with Task Queues, so you don't need to manually update all starters to use the latest Workflow Type.
+- **Easy maintenance**: In comparison to Patching, Task Queue-based Versioning requires less maintenance because you don't need to track which Workflows are running on which version of the Workflow Type.
+
+### Disadvantages of Task Queue-based Versioning
+
+Task Queue Versioning has several disadvantages:
+
+- **Operationally complex**: To run multiple versions of a Task Queue, you need to keep old Workers alive and change which Task Queue Clients point to. This can result in a number of sets of old Workers, which operationally can be complex to manage.
+- **Bug fixes**: This approach can't be used to fix a bug in a currently running or open Workflow. Because the Workflow is already running, you are unable to change the Task Queue it's running on, and thus you can't make changes or fixes to the Workflow.
```
Sushisource, line 106: True, but, you're not precluding from using patching too. Same with Worker Versioning.

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Patching
+
+Patching allows for the update of Workflow Definitions without affecting currently running Workflows.
```
Sushisource, line 110:
```suggestion
Patching allows for the update of Workflow Definitions without breaking currently running Workflows.
```

It'd be pretty useless if it didn't affect them somehow :smile: - what's important is it doesn't cause nondeterminism in them.

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+### What is Patching
+
+Patching allows you to update your Workflow code without terminating any running Workflows and retaining compatibility with your Workflow Event History, as a way to manage changes or upgrades to your Workflow’s logic.
+
+### Advantages of Patching
+
+Patching is valuable in several respects:
+
+- **No Downtime**: Patching allows you to update your Workflow code without terminating any running Workflows.
```
Sushisource, line 130: With this wording, this is true of _all_ the solutions. I'd just strike this line.

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+- **Quick Migration**: Patching allows you to change the yet-to-be-executed behavior of currently open Workflows while remaining compatible with existing Histories. This is because behavior of new Workflows always takes the _newest_ code path.
```
Sushisource, line 131: New workflows yes; and very importantly, workflows which have not yet reached the code where the patch has been added.

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,170 @@
...
+
+- **Simplicity:** Workflow Type versioning provides a straightforward way to version your Workflows, making it an excellent choice for uncomplicated or small-scale projects.
+- **Single Worker Fleet:** With this approach, you're not required to manage separate Worker fleets for each version of your Workflow, potentially minimizing infrastructure complexity and cost.
+
+### Disadvantages of Workflow Type Versioning
+
+Workflow Type versioning has several drawbacks, including:
+
+- **Version Tracking:** All starters need to be informed about each new Workflow Type and its corresponding version, which might introduce communication overhead and room for error.
+
```
Sushisource, line 167: Lots of code duplication is probably the biggest disadvantage and we should mention that.

Sushisource (commented):
What happens to https://docs.temporal.io/workers#worker-versioning ?

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,191 @@
+---
+id: what-is-versioning
+title: What is Versioning?
+sidebar_label: Versioning
+description: Versioning lets you more easily deploy changes to Workflow Definitions.
```
lorensr, line 5:
```suggestion
description: Versioning lets you deploy incompatible changes to Workflow Definitions.
```

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,191 @@
...
+  - explanation
+---
+
+<!-- What is Versioning -->
+
+Versioning is a process allowing for the management and updating of Workflow code or Workflow Definitions without causing non-deterministic behavior in existing Workflows or Tasks.
+Workflow History contains the Execution History of a Workflow, including its Events, Tasks, Signals, and other information.
+When a Workflow Definition changes in an incompatible way, the new Workflow code can't be used to process the existing Workflow History, as the existing Workflow History and the new Workflow code have taken different code paths.
+Versioning allows Workflow to continue running on the old Workflow code, while new Workflows can start new code. This prevents Workflow Histories incompatibility issues.
+Using Versioning, you can deploy changes to your Workflows, as they evolve, without disrupting existing executions.
```
lorensr, line 16:
```suggestion
Versioning is a collection of ways to deploy incompatible changes to Workflow code.
A code change is incompatible if a [Replay](/workflows#replays) of an existing Workflow Execution results in a [non-determinism error](/workflows#deterministic-constraints)—that is, if the code, given a past sequence of [Events](/workflows#event), takes a different code path and generates different [Commands](/workflows#command).
For more information, see [Code changes can cause non-deterministic behavior](/workflows#non-deterministic-change).

Versioning allows existing Workflows to continue running on the old Workflow code, while new Workflows can use the new code. 
This prevents non-determinism errors from occurring.
```

think this might be clearer

docs-src/concepts/what-is-versioning.md
```diff
@@ -0,0 +1,191 @@
...
+Patching, is a set of steps or a technique used to manage changes or upgrades to your Workflow’s logic. Because Workflows can run for weeks, months, or years, you may want to change your Workflow code over time with changes that aren't backwards compatible.
+
+Take the example of needing to modify some business logic that oversees the shipment of an item. Terminating all Workflows midway through delivery would not be a desirable approach. Instead, you would use patching to update your Workflow code.
+
+:::note
+
+In the [Go](https://pkg.go.dev/go.temporal.io/sdk/workflow#GetVersion) and [Java](https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/workflow/Workflow.html#getVersion(java.lang.String,int,int)) SDKs, Patching is referred to as `GetVersion`.
+
+:::
+
```
lorensr, line 142: could link to each dev guide section here

PR #2194, opened 2023-07-10
Add Python worker versioning guide

Sushisource:
Doesn't look like there's a Python versioning section/sidebar item/whatever yet.

I'll leave that for y'all, but here is the equivalent version of the examples

### DIFF
--- a/docs-src/python/how-to-use-worker-versioning-in-python.md
+++ b/docs-src/python/how-to-use-worker-versioning-in-python.md
@@ -1,0 +1,101 @@
+---
+id: how-to-use-worker-versioning-in-python
+title: How to use Worker Versioning in Python
+sidebar_label: Worker Versioning
+description: Version your Python Workers by using build ID-based versioning
+tags:
+  - python
+  - how-to
+---
+
+To use [Worker Versioning](/workers#worker-versioning) in Python, you need to do the following:
+
+1. Determine and assign a Build ID to your built Worker code, and opt in to versioning.
+2. Tell the Task Queue your Worker is listening on about that Build ID, and whether its compatible with an existing Build ID.
+
+### Assign a Build ID to your Worker
+
+Let's say you've chosen `deadbeef` as your Build ID, which might be a short git commit hash (a reasonable choice as Build ID).
+To assign it in your Worker code, assign at least the following Worker Options:
+
+```python
+# ...
+worker = Worker(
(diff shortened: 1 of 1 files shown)


docs-src/python/how-to-use-worker-versioning-in-python.md
```diff
@@ -0,0 +1,101 @@
...
+description: Version your Python Workers by using build ID-based versioning
+tags:
+  - python
+  - how-to
+---
+
+To use [Worker Versioning](/workers#worker-versioning) in Python, you need to do the following:
+
+1. Determine and assign a Build ID to your built Worker code, and opt in to versioning.
+2. Tell the Task Queue your Worker is listening on about that Build ID, and whether its compatible with an existing Build ID.
```
cretz, line 14:
```suggestion
2. Tell the Task Queue your Worker is listening on that Build ID, and whether it's compatible with an existing Build ID.
```

docs-src/python/how-to-use-worker-versioning-in-python.md
```diff
@@ -0,0 +1,101 @@
...
+id: how-to-use-worker-versioning-in-python
+title: How to use Worker Versioning in Python
+sidebar_label: Worker Versioning
+description: Version your Python Workers by using build ID-based versioning
+tags:
+  - python
+  - how-to
+---
+
+To use [Worker Versioning](/workers#worker-versioning) in Python, you need to do the following:
```
cretz, line 11: May need a bit more obvious link to the concepts section as it can almost be a prerequisite before getting into Python specifics

docs-src/python/how-to-use-worker-versioning-in-python.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Specify versions for Commands
+
+By default, Activities, Child Workflows, and Continue-as-New use the same compatible version set as
+the Workflow that invoked them if they're also using the same Task Queue.
+
+If you want to override this behavior, you can specify your intent via the `versioning_intent`
+argument available on the methods you use to invoke these Commands.
+
+<!-- For more information refer to the [conceptual documentation](/concepts/what-is-worker-versioning). -->
```
cretz, line 88: Uncomment this, I think it's important context

cretz (approved):

PR #2208, opened 2023-07-18
Update definition on idempotency for Activities

rachfop:
## What does this PR do?

Clarifies and updates the definition of idempotency in the context of Activities.


flossypurse (commented):
🎈 Nice.

docs-src/concepts/what-is-an-activity-definition.md
```diff
@@ -27,14 +27,16 @@ Activity Definitions are named and referenced in code by their [Activity Type](/
 
 Temporal recommends that Activities be idempotent.
 
+Idempotent means that performing an operation multiple times will have the same result as performing it once. In the context of Temporal, Activities should be designed in a way that they can be safely executed multiple times without causing unexpected or undesired side effects.
```
djmagee, line 30:
```suggestion
Idempotent means that performing an operation multiple times has the same result as performing it once.
In the context of Temporal, Activities should be designed to be safely executed multiple times without causing unexpected or undesired side effects.
```

docs-src/concepts/what-is-an-activity-definition.md
```diff
@@ -27,14 +27,17 @@ Activity Definitions are named and referenced in code by their [Activity Type](/
 
 Temporal recommends that Activities be idempotent.
 
+Idempotent means that performing an operation multiple times has the same result as performing it once.
+In the context of Temporal, Activities should be designed to be safely executed multiple times without causing unexpected or undesired side effects.
```
lorensr, line 31:
```suggestion
In the context of Temporal, Activities should usually be designed to be safely executed multiple times without causing unexpected or undesired side effects.
```

don't need to if activity is non-local an max attempts = 1

lorensr (approved):

docs-src/concepts/what-is-an-activity-definition.md
Sushisource: Non local activities with max attempts = 1 may still be executed more than once. 

EX:
* Worker gets activity task, starts execution, and then loses connection with server. The activity starts executing
* Server times out the task and retries
* Starts again on another worker

Exactly-once semantics are impossible. For a lot of people a lot of the time, that's fine, but highly sensitive applications like finance _must_ ensure they protect against multiple invocations.

lorensr: > Worker gets activity task

This atomically counts as an attempt, right? So Server wouldn't retry after timeout?

I know we can't do exactly once, but thought I remembered from roey or max that we can do at most once and this is how.

Sushisource: Ah, yup fair point. That would indeed avoid it in this case - of course most people are probably going to want retries, but yeah I think your original point is valid. Ignore me :)

lorensr: when I read/hear it, I hear always, so "should usually" to me means there are times when it should not be / doesn't have to be

djmagee (approved):

PR #2216, opened 2023-07-21
EDU-759: update inline links

DJSanti:
## What does this PR do?
This PR fixes some bad links present in the docs.


DJSanti:
Caveat: the issue that I sourced this from had links that were already resolved by others.

djmagee (approved):

PR #2219, opened 2023-07-24
Update inline links on docs.temporal.io

rachfop:
## What does this PR do?

In combination with edu-927, clean up links.


docs-src/concepts/what-is-the-temporal-web-ui.md
```diff
@@ -9,7 +9,7 @@ tags:
 ---
 
 The Temporal Web UI provides users with Workflow Execution state and metadata for debugging purposes.
-It ships with every [Temporal CLI](/concepts/what-is-the-temporal-cli) release and [Docker Compose](/kb/all-the-ways-to-run-a-cluster#docker--docker-compose) update and is available with [Temporal Cloud](/cloud).
```
flossypurse, line 12: `/concepts/what-is-the-temporal-cli` is actually the correct link -
Link magic converts it.

docs-src/references/sdk-metrics.md
```diff
@@ -11,7 +11,7 @@ tags:
 
 The information on this page is relevant to [Temporal SDKs](/concepts/what-is-a-temporal-sdk).
 
-See [Cloud metrics](/cloud/how-to-monitor-temporal-cloud-metrics) for metrics emitted by [Temporal Cloud](/concepts/what-is-temporal-cloud).
```
flossypurse, line 14: Same here, this is actually the correct link - we don't want to change this.

flossypurse (changes requested):
It may be that we have tackled all of the inline link issues?
Because both of these are actually the right ones -

flossypurse (approved):

PR #2220, opened 2023-07-24
Adds metadata tags to TS content

rachfop:
## What does this PR do?

fixes the following error:

```
The file located at /documentation/docs-src/typescript/determinism.md is missing an id... using filename...
The file located at /documentation/docs-src/typescript/introduction.md is missing an id... using filename...
```


flossypurse (approved):
Doesn't look like either of these nodes is in use within a guide config though?

Still - good house keeping.

PR #2231, opened 2023-07-26
Offer both Algolia and Temporal IQ

flossypurse:
This change makes 2 inputs available for discovering information:
Aloglia keyword search
Temporal IQ ai semantic search


djmagee (approved):

docusaurus.config.js
```diff
@@ -197,6 +202,14 @@ module.exports = {
         },
       ],
     },
+    algolia: {
+      apiKey: "4a2fa646f476d7756a7cdc599b625bec",
+      indexName: "temporal",
+      externalUrlRegex: "temporal\\.io",
+      // contextualSearch: true, // Optional, If you have different version of docs etc (v1 and v2) doesn't display dup results
```
djmagee, line 209:
```suggestion
      // contextualSearch: true, // Optional; if you have different version of docs etc (v1 and v2), doesn't display dup results
```
What does "etc" mean in this context?

PR #2233, opened 2023-07-26
EDU-961/962: History Shard corrections

DJSanti:
## What does this PR do?
This PR makes some corrections to History Shards and History Service.

## Notes to reviewers
I've worded this close to what Rob was thinking. Could use some editing for wordage.


DJSanti:
Closing this out because I somehow made a messy merge. Going to reopen once I hear back from Rob.

PR #2235, opened 2023-07-27
Adds Workflow interuptions

rachfop:
## What does this PR do?

- edu-980
- edu-981
- edu-982
- edu-178


docs-src/python/how-to-terminate-a-workflow-execution-in-python.md
```diff
@@ -0,0 +1,12 @@
+---
+id: how-to-terminate-a-workflow-execution-in-python
+title: How to terminate a Workflow Execution in Python
+sidebar_label: Terminate a Workflow Execution
+description: Terminate a Workflow Execution.
```
djmagee, line 5:
```suggestion
description: To terminate a Workflow Execution in Python, use terminate() on the Workflow handle.
```

docs-src/python/interrupt-a-workflow-execution.md
```diff
@@ -0,0 +1,49 @@
+---
+id: interrupt-a-workflow-execution
+title: How to interrupt a Workflow Execution
+description:
```
djmagee, line 4:
```suggestion
description: You can interrupt a Workflow Execution by canceling it or terminating it.
```

docs-src/python/interrupt-a-workflow-execution.md
```diff
@@ -0,0 +1,49 @@
...
+The following are the main differences between canceling and terminating a Workflow in Temporal:
+
+##### Cancel
+
+Canceling a Workflow provides a graceful way to stop Workflow Execution.
+This action resembles sending a `SIGTERM` to a process.
+
+- The system records a `WorkflowExecutionCancelRequested` event in the Workflow History.
+- The Workflow code can handle the cancelation and execute any cleanup logic.
+- A Workflow Task gets scheduled to process the cancelation.
```
lorensr, line 24:
```suggestion
- A Workflow Task gets scheduled to process the cancelation.
- The Workflow code can handle the cancelation and execute any cleanup logic.
```

docs-src/python/interrupt-a-workflow-execution.md
```diff
@@ -0,0 +1,49 @@
...
+
+For more information, see [How to cancel a Workflow Execution](/python/how-to-cancel-a-workflow-in-python).
+
+##### Terminate
+
+Terminating a Workflow forcefully stops Workflow Execution.
+This action resembles killing a process.
+
+- The system records a `WorkflowExecutionTerminated` event in the Workflow History.
+- The termination forcefully and immediately stops the Workflow Execution.
```
lorensr, line 35: not sure what this means technically. what happens in server-worker comms? don't know if this is correct:
- In-progress Activity Executions are canceled.
- No events are recorded after `WorkflowExecutionTerminated`: if there's an in-progress Workflow Task that returns Commands, they will be ignored.

lorensr (approved):

docs-src/python/interrupt-a-workflow-execution.md
```diff
@@ -0,0 +1,49 @@
+---
+id: interrupt-a-workflow-execution
+title: How to interrupt a Workflow Execution
+description:
+sidebar_label: Interrupt a Workflow Execution
+tags:
+  - guide-context
+---
+
+You can interrupt a Workflow Execution in one of the following ways:
```
Sushisource, line 10: AFAIK we have not used this "interrupt" terminology before. I'm not necessarily against introducing it, but, it might be worth discussing more broadly.

docs-src/python/interrupt-a-workflow-execution.md
Sushisource: Worth mentioning that the workflow may choose *not* to be cancelled.

docs-src/python/interrupt-a-workflow-execution.md
Sushisource: That is correct. The workflow worker never learns about termination, except possibly indirectly via WFT not found when replying to an in-progress WFT.

docs-src/python/interrupt-a-workflow-execution.md
```diff
@@ -0,0 +1,49 @@
...
+- A Workflow Task doesn't get scheduled.
+
+For more information, see [How to terminate a Workflow Execution](/python/how-to-terminate-a-workflow-execution-in-python).
+
+##### Summary
+
+In summary:
+
+- Canceling provides a graceful way to stop the Workflow and allows it to handle cancelation logic.
+- Termination forcefully stops the Workflow and prevents any further code execution.
```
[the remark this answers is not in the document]
Sushisource: Correct

flossypurse (commented):
Why not just showcase both examples inline with the "how to interrupt a wfe content" ?

docs-src/python/interrupt-a-workflow-execution.md
cretz: I think just cancel and terminate should be documented alongside other things you can do with a workflow, I don't think we need to group them together with some term that describes both.

Also, are you planning on putting this mostly-non-python-specific docs in all languages or is this non-language-specific description of cancel and terminate in just this language-specific guide?

PR #2239, opened 2023-07-27
EDU-961/962: history shard corrections

DJSanti:
## What does this PR do?
This PR corrects some information regarding History Shards and the History Service.

## Notes to reviewers
Please only look at what-is-a-history-shard and what-is-a-history-service. I had to run yarn to update dependencies and between that and updating to keep up with main, I can't seem to shrink this back down to what I actually worked on.


DJSanti:
Once again, incorporated other changes somehow. Closing and trying again

PR #2246, opened 2023-07-31
Add back IQ

lorensr:
- Add second search for IQ
- Add Cmd/ctrl/windowskey-I shortcut
- Make responsive

Had to override `SearchBar.js`, but also had to import it, so had to add it to the exports of `@docusaurus/theme-search-algolia`, which involves a postinstall step


lorensr:
@rachfop looks like, unrelated to this PR, we need to update some dprint thing?

[Format error](https://github.com/temporalio/documentation/actions/runs/5710500630/job/15470752829?pr=2246)

image

PR #2255, opened 2023-08-01
EDU-131: What is a Logger?

DJSanti:
## What does this PR do?
This PR adds a conceptual piece about Temporal loggers. This piece would probably go into the 

## Notes to reviewers
As generalized as I can make it. I want to provide links to examples in our code, but I wasn't sure where to put them, if at all. I could also use some guidance on which guide config to stick this in. I'm torn between temporal.json or one of the ones for workflows or activities.

### DIFF
--- a/docs-src/concepts/what-is-a-logger.md
+++ b/docs-src/concepts/what-is-a-logger.md
@@ -1,0 +1,32 @@
+---
+id: what-is-a-logger
+title: What is a Logger?
+sidebar_label: Logger
+description: A Logger provides an API containing the logic for recording runtime behavior.
+tags:
+  - term
+  - explanation
+  - logger
+  - metrics
+---
+
+A logger provides an API containing the logic needed to record details about the application behavior at runtime.
+The output posts timestamps alongside the logged values, making it easier to evaluate the state and flow of [Workflows]() and [Activities]().
+
+[Temporal SDKs]() provide both a logging API and a built-in logger implementation.
+The logging implementation provided by the SDK emits messages to the standard output of the [Worker]() executing the code.
+The Temporal logging API automatically injects helpful information (such as [Workflow ID]()) about the execution into the output.
+In addition, the logging API is replay-aware, preventing the duplication of log messages if History Replay is used to restore the app after a crash.
+
+Although the output is timestamped and given a log level (similar to other logging APIs out there), the SDK does not support output customization.
+If you require the ability to customize logger output, replace the logging implementation with one of your own design, or with an adapter that maps our API calls to a third-party library's implementation.
+
+:::caution
+It's permissible to use an API provided by a third-party logging library in your Workflow and Activity Definitions.
+However, third-party APIs won't automatically inject information, and may duplicate logs upon a crash.
+:::
+
+For most cases, Temporal's API and default logging implementations will be sufficient for monitoring execution state.
+If you require more control over the format, threshold, and destination of output, consider using a third-party implementation alongside our logging API, or integrate a third-party library.
+
+
--- a/docs-src/concepts/what-is-a-logger.md
+++ b/docs-src/concepts/what-is-a-logger.md
@@ -31,5 +31,3 @@ For most cases, Temporal's API and default logging implementations will be enoug
 If you require more control over the format, threshold, and destination of output, consider using a third-party implementation alongside Temporal's logging API, or integrate a third-party library.
 
 <!-- TODO: should I add example links down here or within the text? -->
-
-
--- a/docs-src/concepts/what-is-a-logger.md
+++ b/docs-src/concepts/what-is-a-logger.md
@@ -7,26 +7,29 @@ tags:
   - term
   - explanation
   - logger
-  - metrics
+  - api
+  - sdks
 ---
 
 A logger provides an API containing the logic needed to record details about the application behavior at runtime.
-The output posts timestamps alongside the logged values, making it easier to evaluate the state and flow of [Workflows]() and [Activities]().
+The output posts timestamps alongside the logged values, making it easier to evaluate the state and flow of [Workflows](/concepts/what-is-a-workflow) and [Activities](/concepts/what-is-an-activity).
 
-[Temporal SDKs]() provide both a logging API and a built-in logger implementation.
-The logging implementation provided by the SDK emits messages to the standard output of the [Worker]() executing the code.
-The Temporal logging API automatically injects helpful information (such as [Workflow ID]()) about the execution into the output.
+[Temporal SDKs](/concepts/what-is-a-temporal-sdk) provide both a logging API and a built-in logger implementation.
+The logging implementation provided by the SDK emits messages to the standard output of the [Worker](/concepts/what-is-a-worker) executing the code.
+The Temporal logging API automatically injects helpful information (such as [Workflow ID](/concepts/what-is-a-workflow-id)) about the execution into the output.
 In addition, the logging API is replay-aware, preventing the duplication of log messages if History Replay is used to restore the app after a crash.
 
 Although the output is timestamped and given a log level (similar to other logging APIs out there), the SDK does not support output customization.
-If you require the ability to customize logger output, replace the logging implementation with one of your own design, or with an adapter that maps our API calls to a third-party library's implementation.
+If you need to customize logger output, replace the logging implementation with an adapter that maps Temporal's API calls to a third-party library's implementation.
 
 :::caution
 It's permissible to use an API provided by a third-party logging library in your Workflow and Activity Definitions.
 However, third-party APIs won't automatically inject information, and may duplicate logs upon a crash.
 :::
 
-For most cases, Temporal's API and default logging implementations will be sufficient for monitoring execution state.
-If you require more control over the format, threshold, and destination of output, consider using a third-party implementation alongside our logging API, or integrate a third-party library.
+For most cases, Temporal's API and default logging implementations will be enough for monitoring execution state.
+If you require more control over the format, threshold, and destination of output, consider using a third-party implementation alongside Temporal's logging API, or integrate a third-party library.
+
+<!-- TODO: should I add example links down here or within the text? -->
 
 


docs-src/concepts/what-is-a-logger.md
```diff
@@ -0,0 +1,33 @@
...
+  - logger
+  - api
+  - sdks
+---
+
+A logger provides an API containing the logic needed to record details about the application behavior at runtime.
+The output posts timestamps alongside the logged values, making it easier to evaluate the state and flow of [Workflows](/concepts/what-is-a-workflow) and [Activities](/concepts/what-is-an-activity).
+
+[Temporal SDKs](/concepts/what-is-a-temporal-sdk) provide both a logging API and a built-in logger implementation.
+The logging implementation provided by the SDK emits messages to the standard output of the [Worker](/concepts/what-is-a-worker) executing the code.
```
cretz, line 18: This is not true of all SDKs. Many just use existing APIs and default to no logs.

docs-src/concepts/what-is-a-logger.md
cretz, line 1: Arguably the concept of a "logger" is a general programming one and not specific to Temporal nor does Temporal have enough non-SDK-specific things to say about a logger to deserve its own page IMO.

docs-src/concepts/what-is-a-logger.md
```diff
@@ -0,0 +1,33 @@
...
+The Temporal logging API automatically injects helpful information (such as [Workflow ID](/concepts/what-is-a-workflow-id)) about the execution into the output.
+In addition, the logging API is replay-aware, preventing the duplication of log messages if History Replay is used to restore the app after a crash.
```
cretz, line 20:
```suggestion
In addition, the workflow logging API is replay-aware, preventing the duplication of log messages during replay.
```

Doesn't apply to activities and replay is not always about crashes.

docs-src/concepts/what-is-a-logger.md
```diff
@@ -0,0 +1,33 @@
...
+
+Although the output is timestamped and given a log level (similar to other logging APIs out there), the SDK does not support output customization.
```
cretz, line 22: > timestamped

This is SDK specific

> the SDK does not support output customization

This is not true. Most logging interfaces let you implement whatever approach you want. And ones using built-in logging frameworks (Java slf4j, Python logging module, .NET logging extension, etc) all support _lots_ of output customization. SDK just uses these like any other library would, there's nothing Temporal specific here in many cases (except for the extra properties we add and the skip-during-replay we offer).

docs-src/concepts/what-is-a-logger.md
```diff
@@ -0,0 +1,33 @@
...
+If you need to customize logger output, replace the logging implementation with an adapter that maps Temporal's API calls to a third-party library's implementation.
```
cretz, line 23: This is not true across SDKs. If you need to customize logger output in Python for example, you configure it, you don't replace anything.

docs-src/concepts/what-is-a-logger.md
```diff
@@ -0,0 +1,33 @@
...
+
+:::caution
+It's permissible to use an API provided by a third-party logging library in your Workflow and Activity Definitions.
+However, third-party APIs won't automatically inject information, and may duplicate logs upon a crash.
```
cretz, line 27: This is not necessarily true. We often wrap the third party libraries, not even invoking them on replay. And remember, replay situations are for more than just crashes.

docs-src/concepts/what-is-a-logger.md
```diff
@@ -0,0 +1,33 @@
...
+:::
+
+For most cases, Temporal's API and default logging implementations will be enough for monitoring execution state.
```
cretz, line 30: I don't think this is true in all cases. For Go and Python and .NET I think the defaults are usually _not enough_.

docs-src/concepts/what-is-a-logger.md
```diff
@@ -0,0 +1,33 @@
...
+If you require more control over the format, threshold, and destination of output, consider using a third-party implementation alongside Temporal's logging API, or integrate a third-party library.
```
cretz, line 31: This is not true. You don't always need third party. Many times the full advanced logging implementation is built in to the standard library (e.g. Python and the upcoming Go slog).

flossypurse (changes requested):
Using "Request changes" status to get a better idea of where things are at from the main list.

docs-src/concepts/what-is-a-logger.md
mjameswh: Would go for something more middle ground, like `defaults may be enough for some, while others may require customizations...`. In most cases, those who needs more than the default probably already know that they need more.

docs-src/concepts/what-is-a-logger.md
tomwheeler: @cretz Is there a place where the details are documented, at least internally? 

I'm most familiar with the Go SDK, which works as described. The Java SDK provides a [replay-aware logger](https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/internal/logging/ReplayAwareLogger.html), which the API docs suggest is accessible via `workflow.getLogger`, although there is no Activity logger. My understanding is that the TypeScript SDK recently added a logger. I don't know about the others. 

Perhaps since there is so much variation in the logging support offered by each SDK, documenting how it works might be better done as part of each SDK-specific dev guide.

cretz: > Is there a place where the details are documented, at least internally?
Each SDK is different here. I am not sure we document them outside of SDK reference (even if we do there). I'd totally support documenting the SDK-specific details. That's unrelated to a general page.

> I'm most familiar with the Go SDK, which works as described. The Java SDK provides a [replay-aware logger](https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/internal/logging/ReplayAwareLogger.html), which the API docs suggest is accessible via workflow.getLogger, although there is no Activity logger. My understanding is that the TypeScript SDK recently added a logger. I don't know about the others.

This comment is very specifically on the section stating that the implementation provided by the SDK emits to standard output. Not only is the standard output part not necessarily true. Python and .NET and Java all use ecosystem native logging that may not log anything by default (e.g. slf4j I think does nothing unless you include a JAR that has an impl).

> Perhaps since there is so much variation in the logging support offered by each SDK, documenting what is available and how to use it might be better done as part of each SDK-specific dev guide.

Exactly! I made this comment just below at https://github.com/temporalio/documentation/pull/2255#discussion_r1280933635

DJSanti:
Really appreciate the feedback from y'all. I'm seeing that this should be more SDK-specific, as in spinning off into how-to articles for setting up logging rather than this conceptual piece.

docs-src/concepts/what-is-a-logger.md
cretz: > I filed the original issue because a user was confused about where the output from the Workflow and Activity loggers in the Go SDK went and how to change its format and/or destination.

But this doc we're reviewing is not for Go SDK

> AFAIK, there is no way for a developer to customize the format or output destination unless they write their own implementation (or an adapter class for an existing third-party package like zap), yet I was unable to find that documented anywhere.

Yes, it depends on the implementation they have. We provide an interface, how that interface is implemented is up to users. Formatting and output destination are but a couple of the many options one may choose to think about when implementing. We should move zap into contrib (https://github.com/temporalio/sdk-go/issues/704) as well as the upcoming `slog` (https://github.com/temporalio/sdk-go/issues/1158). But this PR doesn't help you solve your original problem of how to customize format or output destination.

> For an SDK like Java, which appears to use SLF4J, that might just be a couple of sentences that say

Right, see https://docs.temporal.io/dev-guide/java/observability#logging. That can be expanded/improved.

> Documentation for other SDKs could go into more detail, where necessary. To be clear, I don't think we need to cover how to configure a third-party logging package, but IMHO we should at least identify which one is being used by the SDK's logger (where applicable) so that people can find the docs provided by that project.

:+1:

docs-src/concepts/what-is-a-logger.md
cretz: > defaults may be enough for some, while others may require customizations

That's so obvious of a statement is it worth saying?

docs-src/concepts/what-is-a-logger.md
tomwheeler: > Right, see https://docs.temporal.io/dev-guide/java/observability#logging. That can be expanded/improved.

I hadn't seen this before, but it's pretty much what I had in mind when I filed the issue (although I didn't realize how different each SDK was).

tomwheeler (commented):
When I filed [EDU-131](https://temporalio.atlassian.net/browse/EDU-131) in January, I did so on behalf of a Go SDK user who couldn't find documentation on locating or customizing the output of the Workflow logger. I understood his plight, because I had faced it myself some months earlier. I noted in the issue that what I described was in the context of Go, "the need to describe how logging works in Temporal applications transcends SDKs."

However, it's now clear that each SDK has a different approach to logging, and since all of that logging takes places on the client side, it's not really possible to provide actionable information about logging in a generic way. We can reasonably expect developers to understand what loggers are, so we just need to explain how to use them, and perhaps provide some guidance on the advantages of using the SDK-provided API for logging vs. directly using a third-party logging package such as Log4J. 

With that in mind, I think having a section on logging in the Dev Guide specific to each SDK would best serve that need. The Go, Java, TypeScript, and Python Dev Guides now provide this under the "Observability" section (though I think most of those might have still been in the "Legacy" docs when I filed the issue). IMHO, these sections should address each of these points:

- Whether the SDK provides a logging API, and if so, a brief example of how to use it
- Where the logger's output goes, by default
- How to customize the logger's output, where applicable. 
   - A few sentences should be fine in most cases. This shouldn't attempt to 
     duplicate docs provided by a third-party implementation (e.g., Java's SLF4J), 
     but should at least mention which third-party implementation is being used
     and ideally provide a link to the config documentation provided by that project.
- Explain that it's acceptable to use a third-party logging package directly, from either Workflow or Activity code
   - I think some may be unclear about this because they may infer that logging is I/O and that's a side effect which could violate our determinism constraint.
- Point out any benefits (e.g., replay awareness that avoids duplicate log messages) to using the Temporal API.

DJSanti:
Agreed with Tom. Going to close this one. Thank you all for your time.

PR #2274, opened 2023-08-09
EDU-1035 improve link previews

flossypurse:
Preview: https://temporal-documentation-git-improvepreviews.preview.thundergun.io/

This PR aims to improve the functionality and experience of the link preview feature.
This is an option. For example, we could remove it altogether: https://github.com/temporalio/documentation/pull/2273

Note that the ability to preview links is only available on desktop sized screens.


flossypurse:
went with https://github.com/temporalio/documentation/pull/2273

PR #2279, opened 2023-08-10
EDU-1024: tag Go versioning source nodes

DJSanti:
## What does this PR do?
This PR adds tags to all source nodes for the Go Versioning section of the Dev Guide.


docs-src/go/versioning.md
```diff
@@ -136,16 +133,12 @@ Upgrading a Workflow is straightforward if you don't need to preserve your curre
...
-new `WorkflowType`, and change your start path (calls to `StartWorkflow()`) to start the new Workflow
-type.
+However, if you want your currently running Workflows to proceed based on the current Workflow logic, but you want to ensure new Workflows are running on new logic, you can define your Workflow as a new `WorkflowType`, and change your start path (calls to `StartWorkflow()`) to start the new Workflow type.
 
 ## Sanity checking
 
 The Temporal client SDK performs a sanity check to help prevent obvious incompatible changes.
-The sanity check verifies whether a Command made in replay matches the event recorded in history,
-in the same order. The Command is generated by calling any of the following methods:
+The sanity check verifies whether a Command made in replay matches the event recorded in history, in the same order. The Command is generated by calling any of the following methods:
```
djmagee, line 141:
```suggestion
The sanity check verifies whether a Command made in replay matches the event recorded in history, in the same order.
The Command is generated by calling any of the following methods:
```

docs-src/go/versioning.md
```diff
@@ -155,12 +148,9 @@ in the same order. The Command is generated by calling any of the following meth
 - workflow.SignalExternalWorkflow()
 - workflow.Sleep()
 
-Adding, removing, or reordering any of the above methods triggers the sanity check and results in
-a non-deterministic error.
+Adding, removing, or reordering any of the above methods triggers the sanity check and results in a non-deterministic error.
```
djmagee, line 151:
```suggestion
Adding, removing, or reordering any of the preceding methods triggers the sanity check and results in a non-deterministic error.
```

docs-src/go/versioning.md
```diff
@@ -155,12 +148,9 @@ in the same order. The Command is generated by calling any of the following meth
...
-The sanity check does not perform a thorough check. For example, it does not check on the Activity's
-input arguments or the timer duration. If the check is enforced on every property, then it becomes
-too restricted and harder to maintain the Workflow code. For example, if you move your Activity code
-from one package to another package, that changes the `ActivityType`, which technically becomes a different
-Activity. But, we don't want to fail on that change, so we only check the function name part of the
-`ActivityType`.
+The sanity check does not perform a thorough check. For example, it does not check on the Activity's input arguments or the timer duration.
+If the check is enforced on every property, then it becomes too restricted and harder to maintain the Workflow code.
+For example, if you move your Activity code from one package to another package, that changes the `ActivityType`, which technically becomes a different Activity.
+But, we don't want to fail on that change, so we only check the function name part of the `ActivityType`.
```
djmagee, line 156:
```suggestion
The sanity check does not perform a thorough check.
For example, it does not check on the Activity's input arguments or the Timer duration.
If the check is enforced on every property, it becomes too restrictive and harder to maintain the Workflow code.
For example, if you move your Activity code from one package to another package, that move changes the `ActivityType`, which technically becomes a different Activity.
But we don't want to fail on that change, so we check only the function name part of the `ActivityType`.
```

flossypurse (approved):

PR #2296, opened 2023-08-18
EDU-466: clarify connection differences between client connection and Worker in TS

DJSanti:
## What does this PR do?
This PR adds a note to the TypeScript security page about which Connection class applies to the client connection and the Worker's.


DJSanti:
Made change to typescript/security, which appears to not be a node. Going to check with reporter to see where the user saw this code example.

antmendoza (approved):

docs-src/typescript/security.md
```diff
@@ -42,15 +42,14 @@ const client = new WorkflowClient({
 
 ## Encryption in transit with mTLS
 
-There are two classes in the SDK that connect to the Temporal server, the [Worker](https://typescript.temporal.io/api/classes/worker.Worker) and the client [Connection](https://typescript.temporal.io/api/classes/client.Connection/).
+There are two classes in the SDK that connect to the Temporal Server: the [Worker](https://typescript.temporal.io/api/classes/worker.Worker) and the client [Connection](https://typescript.temporal.io/api/classes/client.Connection/).
```
djmagee, line 45:
```suggestion
Two classes in the SDK connect to the Temporal Server: the [Worker](https://typescript.temporal.io/api/classes/worker.Worker) and the client [Connection](https://typescript.temporal.io/api/classes/client.Connection/).
```

docs-src/typescript/security.md
```diff
@@ -42,15 +42,14 @@ const client = new WorkflowClient({
...
 When instantiating either of them, you may choose whether to connect securely or not.
```
djmagee, line 46:
```suggestion
When instantiating either of them, you can choose whether to connect securely.
```

docs-src/typescript/security.md
```diff
@@ -42,15 +42,14 @@ const client = new WorkflowClient({
...
 
 - In order to connect to the server using TLS, set a _truthy_ value (`true` or [TLSConfig](https://typescript.temporal.io/api/interfaces/client.TLSConfig) for custom options) in the `tls` configuration option.
```
djmagee, line 48:
```suggestion
- To connect to the server using TLS, set a _truthy_ value (`true` or [TLSConfig](https://typescript.temporal.io/api/interfaces/client.TLSConfig) for custom options) in the `tls` configuration option.
```

docs-src/typescript/security.md
```diff
@@ -106,6 +105,8 @@ run().catch((err) => {
 });
 ```
 
+Make sure that the Worker is using `NativeConnection`, and the client Connection uses `Connection`.
```
djmagee, line 108:
```suggestion
Make sure that the Worker uses `NativeConnection` and the client Connection uses `Connection`.
```

djmagee (commented):
`docs-src/typescript/security.md` is indeed a vestige of the "legacy" dev guide and isn't rendered on our site.

djmagee (approved):

PR #2307, opened 2023-08-25
Clarify Workflow limits

lorensr:
Clarify what the limits are and what they do / don't apply to, to help with user questions like [this one](https://temporalio.slack.com/archives/C04S80QKB2Q/p1692938980554009).

### DIFF
--- a/docs-src/concepts/what-is-a-workflow-execution.md
+++ b/docs-src/concepts/what-is-a-workflow-execution.md
@@ -145,15 +145,16 @@ For example, it may be reasonable to use Continue-As-New once per day for a long
 
 ### Limits
 
-Each pending Activity generates a metadata entry in the Workflow's mutable state.
-Too many entries create a large mutable state, which causes unstable persistence.
+There is no limit to the number of concurrent Workflow Executions.
 
-To protect the system, Temporal enforces a maximum number (2,000 by default) of pending Activities, Child Workflows, Signals, or Cancellation requests per Workflow.
+Each in-progress Activity generates a metadata entry in the Workflow's mutable state.
+Too many entries create a large mutable state, which causes unstable persistence.
+To protect the system, Temporal enforces a maximum number (2,000 by default) of incomplete Activities, Child Workflows, Signals, or Cancellation requests per Workflow Execution.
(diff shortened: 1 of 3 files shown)


docs-src/concepts/what-is-a-workflow-execution.md
```diff
@@ -145,15 +145,16 @@ For example, it may be reasonable to use Continue-As-New once per day for a long
 
 ### Limits
 
-Each pending Activity generates a metadata entry in the Workflow's mutable state.
+There is no limit to the number of concurrent Workflow Executions.
+
+Each in-progress Activity generates a metadata entry in the Workflow's mutable state.
 Too many entries create a large mutable state, which causes unstable persistence.
+To protect the system, Temporal enforces a maximum number (2,000 by default) of incomplete Activities, Child Workflows, Signals, or Cancellation requests per Workflow Execution.
```
mfateev, line 152: It is unclear if this limit is aggregate across all these or per type. Is it OK to have 2k activities and 2k child workflows at the same time.

docs-src/concepts/what-is-a-workflow-execution.md
```diff
@@ -145,15 +145,16 @@ For example, it may be reasonable to use Continue-As-New once per day for a long
 
 ### Limits
 
-Each pending Activity generates a metadata entry in the Workflow's mutable state.
-Too many entries create a large mutable state, which causes unstable persistence.
+There is no limit to the number of concurrent Workflow Executions.
+
+Each in-progress Activity generates a metadata entry in the Workflow Execution's mutable state.
```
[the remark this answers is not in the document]
lorensr:
```suggestion
There is no limit to the number of concurrent Workflow Executions.

However, there is a limit to the length and size of a Workflow Execution's Event History (by default, [51,200 Events](https://github.com/temporalio/temporal/blob/e3496b1c51bfaaae8142b78e4032cc791de8a76f/service/history/configs/config.go#L382) and [50 MB](https://github.com/temporalio/temporal/blob/e3496b1c51bfaaae8142b78e4032cc791de8a76f/service/history/configs/config.go#L380)).

There is also a limit to the number of certain types of incomplete operations.

Each in-progress Activity generates a metadata entry in the Workflow Execution's mutable state.state.
```

makes sense! edited to use the "length" and "size" terms the SDKs use

would say activities/sec is throughput, but not exactly total pending, and not history size

flossypurse (approved):

docs-src/concepts/what-is-a-workflow-execution.md
flossypurse: Clarified by @lorensr to "2000 for each type of operation"

PR #2309, opened 2023-08-25
EDU-1026: add Cloud latency info

DJSanti:
## What does this PR do?
This PR adds additional information to the end of `operating-envelope-latency` concerning latency.

### DIFF
--- a/docs-src/cloud/operating-envelope-latency.md
+++ b/docs-src/cloud/operating-envelope-latency.md
@@ -18,5 +18,5 @@ In June 2023, latency measurements over a week period for starting and signaling
 
 As we continue working on improving latencies, these numbers will decrease over time.
 
-The latency observed from the Temporal Client might be impacted by other system components, such as the Codec Server, egress proxy, and the network itself.
+The latency observed from the Temporal Client might be affected by other system components, such as the Codec Server, egress proxy, and the network itself.
 Concurrent operations on the same Workflow Execution could lead to increased latency.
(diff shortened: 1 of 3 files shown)


docs-src/cloud/operating-envelope-latency.md
```diff
@@ -18,4 +18,5 @@ In June 2023, latency measurements over a week period for starting and signaling
 
 As we continue working on improving latencies, these numbers will decrease over time.
 
+Be aware that the latency observed on the Temporal Client might be impacted by other system components, such as the Codec Server, egress proxy, and the network itself.
```
lorensr, line 21:
```suggestion
The latency observed from the Temporal Client might be impacted by other system components, such as the Codec Server, egress proxy, and the network itself.
```

I like avoiding commands, and to some extent every sentence in the docs is "be aware". If important enough, would add a warning block, but don't think important enough here

lorensr (approved):

docs-src/cloud/operating-envelope-latency.md
djmagee, line 21:
```suggestion
Be aware that the latency observed on the Temporal Client might be affected by other system components, such as the Codec Server, egress proxy, and the network itself.
```
Per https://developers.google.com/style/word-list#impact.

docs-src/cloud/operating-envelope-latency.md
```diff
@@ -18,4 +18,5 @@ In June 2023, latency measurements over a week period for starting and signaling
 
 As we continue working on improving latencies, these numbers will decrease over time.
 
+The latency observed from the Temporal Client might be affected by other system components, such as the Codec Server, egress proxy, and the network itself.
```
flossypurse, line 21:
```suggestion
Other system components, such as the Codec Server, egress proxy, and the network itself, could impact the latency observed from the Temporal Client.
```

passive -> active

flossypurse (changes requested):

PR #2314, opened 2023-08-28
Update TS docker build information

bergundy:


mjameswh (approved):
@bergundy While we're at this, maybe also update version on all Docker image references to `:20`.

bergundy:
@mjameswh PTAL, I also added advice to refer to the production sample for bundling the workflow code and saving it to a file.

docs-src/typescript/how-to-run-a-worker-on-docker.md
```diff
@@ -31,7 +37,7 @@ RUN npm install --only=production \
 CMD ["build/worker.js"]
 ```
 
-For smaller images and/or more secure deployments, it is also possible to use `-slim` Docker image variants (like `node:18-bullseye-slim`) or `distroless/nodejs` Docker images (like `gcr.io/distroless/nodejs:18`) with the following caveats.
+For smaller images and/or more secure deployments, it is also possible to use `-slim` Docker image variants (like `node:20-bullseye-slim`) or `distroless/nodejs` Docker images (like `gcr.io/distroless/nodejs:20`) with the following caveats.
```
mjameswh, line 40: Looks like the distroless project changed their URL structures starting with Node 20.

```suggestion
For smaller images and/or more secure deployments, it is also possible to use `-slim` Docker image variants (like `node:20-bullseye-slim`) or `distroless/nodejs` Docker images (like `gcr.io/distroless/nodejs20-debian11`) with the following caveats.
```

docs-src/typescript/how-to-run-a-worker-on-docker.md
```diff
@@ -82,7 +88,7 @@ RUN npm install --only=production \
 
 # -- RESULTING IMAGE --
 
-FROM gcr.io/distroless/nodejs:18
+FROM gcr.io/distroless/nodejs:20
```
mjameswh, line 91:
```suggestion
FROM gcr.io/distroless/nodejs20-debian11
```

mjameswh (approved):
Good, just need to fix URL for distroless node 20

bergundy:
@mjameswh I made a few changes since your review, PTAL.

flossypurse (approved):

docs-src/typescript/how-to-run-a-worker-on-docker.md
```diff
@@ -9,29 +9,38 @@ tags:
   - typescript
 ---
 
+:::note
+
+Before proceeding, we recommend referring to our [production sample](https://github.com/temporalio/samples-typescript/tree/main/production),
```
mjameswh, line 14: I'd reverse that sentence. Our recommendation is really that they prepare the workflow bundle ahead-of-time, not that they look at the production sample.

```suggestion
To improve worker startup time, we recommend that workflow bundles be prepared ahead-of-time. See our [production sample](https://github.com/temporalio/samples-typescript/tree/main/production) for details.
```

docs-src/typescript/how-to-run-a-worker-on-docker.md
```diff
@@ -9,29 +9,38 @@ tags:
...
 ```dockerfile
-FROM node:18-bullseye
+FROM node:20-bullseye
 
+# For better cache utilization, copy package.json and lock file first and install the dependencies before copying the
+# rest of the application and building.
 COPY . /app
 WORKDIR /app
 
+# Alternatively, run npm ci, which installs only dependencies specified in the lock file and is generally faster.
```
mjameswh, line 36: I think at this point, it should be reasonable to recommend this instead:

```
npm ci --omit dev
```
`--omit` was introduced in NPM 8, which itself got introduced in 16.11.0.

mjameswh (approved):
A few more suggestions, but still good.

docs-src/typescript/how-to-run-a-worker-on-docker.md
bergundy: `--omit dev` would require a separate build step, I'd avoid that for now.

docs-src/typescript/how-to-run-a-worker-on-docker.md
bergundy: Changed to more or less what you're suggesting.

PR #2317, opened 2023-08-29
EDU-443: cluster env vars reference

DJSanti:
## What does this PR do?
This code adds a reference to the constants defined here: https://github.com/temporalio/temporal/blob/master/environment/env.go.

## Notes to reviewers
Although the title says "Cluster env var reference", I don't mind changing the file name and metadata to reflect that the constants pertain more to databases the Cluster can hook up to.


DJSanti:
Reached out across teams on Slack to get someone to review. According to Alex, Yimin, and [this PR](https://github.com/temporalio/temporal/issues/2668), the current variables defined in env.go are inconsistent and the current intent is to remove them. Therefore, I'm going to close out this draft and let them sort this out.

PR #2328, opened 2023-09-08
EDU-1141: Update messaging about tctl deprecation

djmagee:
## Preview

**https://temporal-documentation-git-edu-1141-tctl-dep.preview.thundergun.io/tctl-v1/**

## What does this PR do?

Modifies the introduction of the tctl v1.17 command reference to properly set expectations about the deprecation of tctl.


tlalfano (changes requested):
Looks good, I would remove v1.17 from the message and instead go with:
we will deprecate tctl

tlalfano (approved):
Looks great! Thank you.

flossypurse (approved):

PR #2341, opened 2023-09-28
Adds constraints for CSA

rachfop:
## What does this PR do?

I've added details regarding the validation criteria for custom Search Attribute names in Temporal Cloud:

- Names should be a maximum of 64 characters in length.
- They should adhere to the following character set: `[a-zA-Z0-9.,:-_()\/\\ ]`.
This update provides clarity for users on the constraints they need to follow when creating custom Search Attributes. Please review the changes and let me know if any adjustments are required.


docs-src/cloud/temporal-cloud-limits-sheet.md
```diff
@@ -83,6 +83,15 @@ The default maximum number of [Custom Search Attributes](/visibility#custom-sear
 
 For more information on limits on the number and size of custom Search Attributes you can create, see [Custom Search Attributes limits](/visibility#custom-search-attributes).
 
+### What are constraints for Custom Search Attribute in Temporal Cloud?
+
+When creating custom Search Attributes in Temporal Cloud, the attribute names must adhere to the following constraints:
+
+- Maximum characters: 64
+- Allowed characters: `[a-zA-Z0-9.,:-_\/\s]+`
```
shakeelrao, line 91: could we simplify to something like: `a-z`, `A-Z`, `0-9`, `.`, `,`, `:`, `-`, `_`, ` `(space), `(`, `)`, `/`, `\`.

technically, `\s` isn't accurate, since we disallow tabs and newlines (only space character).

here's the regex used in the server side code for reference: 
```go
regexp.MustCompile(`^[a-zA-Z0-9.,:\-_ ()/\\]+$`)
```

rachfop: Awesome. Yeah, good point. Reverted that and used your example.

shakeelrao (approved):
looks great to me! i'll also cc @mastermanu here as well.

mastermanu (approved):

docs-src/concepts/what-is-a-search-attribute.md
```diff
@@ -168,6 +168,8 @@ Default total maximum number of Search Attribute **keys** per Temporal Cluster i
 This is configurable with [`SearchAttributesNumberOfKeysLimit`, `SearchAttributesTotalSizeLimit` and `SearchAttributesSizeOfValueLimit`](https://github.com/temporalio/temporal/blob/v1.7.0/service/history/configs/config.go#L440-L442), if you know what you are doing.
 -->
 
+For Temporal Cloud specific configurations, see the [Default limits](/cloud/operating-envelope#what-are-the-default-maximum-numbers-of-custom-search-attributes) sheet.
```
shakeelrao, line 171: nit: capital L?

flossypurse (approved):
👍

PR #2344, opened 2023-09-29
Clarify default behavior of Activities and Child Workflows

lorensr:


rachfop (approved):
Nice. I had a comment to go and write this. Glad to see you already had.

PR #2351, opened 2023-10-03
Adds legacy Java workflow inheritance content to dev guide

flossypurse:
The content in this PR existed in the legacy Java docs and was missed during the migration.
This PR adds it to the Foundations section of the Java dev guide with minimal updates.


assembly/guide-configs/app-dev/java/foundations.json
```diff
@@ -47,6 +47,10 @@
       "type": "h2",
       "id": "java/how-to-develop-a-workflow-definition-in-java"
     },
+    {
+      "type": "h3",
+      "id": "java/interface-heirarchies"
```
tomwheeler, line 52: Probably a legacy issue, but I noticed that "hierarchies" is misspelled here (and in the name of the file itself as well as other references to it).

docs-src/java/interface-heirarchies.md
```diff
@@ -0,0 +1,90 @@
...
+}
+
+@WorkflowInterface
+public interface Workflow1 extends BaseWorkflow {}
+
+@WorkflowInterface
+public interface Workflow2 extends BaseWorkflow {}
+```
+
+If we attempt to register implementations of Workflow1 and Workflow2 with a Worker will fail.
```
tomwheeler, line 78: There's a grammatical error here that makes the sentence hard to parse. I can interpret the intent in two different ways and it's not clear which is correct, so I am not sure which to suggest. Maybe rephrase as "Any attempt to register implementations for both interfaces with a Worker will fail."

tomwheeler (changes requested):
I would recommend fixing the grammatical issue mentioned in `docs-src/java/interface-heirarchies.md` before publishing this. 

The filename misspelling is less important since the references are at least consistent with one another. It would be great if that could be fixed too, but certainly not essential.

rachfop (approved):

assembly/guide-configs/app-dev/java/foundations.json
flossypurse: No, that's me :) Not sure what I was thinking there.

docs-src/java/interface-heirarchies.md
flossypurse: Good catch.

tomwheeler (approved):
Ship it!

PR #2359, opened 2023-10-04
Adds context and anchor links to status for Workflows

rachfop:
## What does this PR do?

For the term “Workflow Execution”, under the “Status” section, each status should be an anchor and each should be explained


flossypurse (approved):
👍

PR #2364, opened 2023-10-05
clarify that updates are synchronous replacements for queries and signals

axfelix:
Adds some introductory text around updates to make the same connections as at Replay 2023 — they are a synchronous implementation of query/signal

### DIFF
--- a/docs-src/concepts/what-is-an-update.md
+++ b/docs-src/concepts/what-is-an-update.md
@@ -17,6 +17,12 @@ An Update is a request to and a response from a Temporal Client to a [Workflow E
 
 - [How to develop, send, and handle Updates in Go](/go/updates)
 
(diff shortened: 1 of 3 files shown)


angelazhou32 (approved):

docs-src/concepts/what-is-an-update.md
```diff
@@ -17,6 +17,12 @@ An Update is a request to and a response from a Temporal Client to a [Workflow E
 
 - [How to develop, send, and handle Updates in Go](/go/updates)
 
+You can think of an Update as a synchronous, blocking call that could replace both a Signal and a Query. An update is:
+
+- A Signal with a response
+- A Query that can change state
+- The logical model of a Signal with the overhead and latency of a Query
```
lorensr, line 24: More write load than a query, and perhaps lower latency? @alexshtin would know

PR #2367, opened 2023-10-05
Mentioned the requirement to have a local Temporal Cluster running

tomwheeler:
While preparing a correction to the docs by following the steps in this README.md, it occurred to me that it does not mention the requirement that you have a Temporal Cluster running. This could be frustrating for a new user who doesn't realize why the `./worker.js` command fails, since the resulting error message doesn't say anything about what connection failed or what it was trying to connect to.

## What does this PR do?

Improves the instructions for making changes to our documentation by making an implicit (and perhaps overlooked) requirement more obvious, with a specific example of the command to run and a description of what the user would observe if they didn't do it.


flossypurse (approved):
Nice addition.
TY @tomwheeler

PR #2378, opened 2023-10-10
Update activityResult type

hu-dabao:
## What does this PR do?
The documentation has an (typo) issue that `activityResult` should be a `YourActivityResultObject` type instead of the pointer. Otherwise, following this documentation will cause issues like
```
failed executing YourActivityDefinition activity: payload item 0: unable to decode: json: Unmarshal(nil *foo.activityResult)",
```

The right way in another example here https://learn.temporal.io/getting_started/go/first_program_in_go/ is also **not** declaring the var a pointer.

```
var withdrawOutput string

withdrawErr := workflow.ExecuteActivity(ctx, Withdraw, input).Get(ctx, &withdrawOutput)
```


hu-dabao:
@flossypurse if you can help to review

rachfop:
I've edited the original source:
https://github.com/temporalio/documentation-samples-go/pull/40
Thanks for pointing this out. Hope to get it approved soon.

rachfop (approved):

PR #2386, opened 2023-10-12
correct milliseconds -> seconds in SDK metrics reference

axfelix:
Addresses https://temporalio.atlassian.net/browse/EDU-1242


flossypurse (approved):
yes, appears to be correct.

PR #2392, opened 2023-10-13
Delete docs changelog and redirect users to temporal.io/change-log

angelazhou32:
This PR deletes the changelog config and redirects them instead to temporal.io/change-log, so that if a user goes to "https://docs.temporal.io/changelog", they will now be directed to "temporal.io/change-log". See screen record of new changes:

https://github.com/temporalio/documentation/assets/141692465/49d7e154-e3e5-4ded-bec8-260f06e9cdf5

### DIFF
--- a/src/pages/changelog.js
+++ b/src/pages/changelog.js
@@ -5,7 +5,7 @@ function ChangelogRedirect() {
         window.location.href = "https://temporal.io/change-log";
     }, []);
 
-    return null;  // Return null since we don't need any actual rendering here.
+    return null;
 }
 
 export default ChangelogRedirect;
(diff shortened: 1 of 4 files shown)


src/pages/changelog.js
```diff
@@ -0,0 +1,11 @@
+import React, { useEffect } from 'react';
```
flossypurse, line 1: the vercel redirect should work for all prod cases.

do we need a local/development redirect too? 

Meaning - this page only really matters when not hitting slugs at docs.temporal.io - right?

flossypurse (approved):
Not sure we need the duplicate redirects.

But it definitely does the job!

PR #2402, opened 2023-10-17
Update documentation to clarify the requirement of setting a Workflow Id for Python SDK.

rachfop:
## Summary:

Update documentation to clarify the requirement of setting a Workflow Id for Python SDK.

## Changes:

Removed the phrase implying that setting a Workflow Id is not mandatory.
Clearly stated the requirement to set a Workflow Id for Python SDK.

## Background:

There was a confusion due to the wording in the documentation suggesting that providing a Workflow Id is not required. This was brought up by a user on Temporalio's Slack and further discussed. 

The context and essence of the phrase meant that mapping a Workflow Id to a business process isn't required. However, this led to misinterpretation, and hence we are making it explicit in the docs that setting a Workflow Id is a requirement.

This change addresses the reported bug: EDU-1261


docs-src/python/set-workflow-id.md
```diff
@@ -7,4 +7,4 @@ tags:
   - guide-context
 ---
 
-Although it is not required, we recommend providing your own [Workflow Id](/concepts/what-is-a-workflow-id) that maps to a business process or business entity identifier, such as an order identifier or customer identifier.
+You must set a [Workflow Id](/concepts/what-is-a-workflow-id). When setting a Workflow Id, we recommended mapping it to a business process or business entity identifier, such as an order identifier or customer identifier.
```
flossypurse, line 10:
```suggestion
You must set a [Workflow Id](/concepts/what-is-a-workflow-id).
When setting a Workflow Id, we recommended mapping it to a business process or business entity identifier, such as an order identifier or customer identifier.
```

flossypurse (approved):

docs-src/python/set-workflow-id.md
rachfop: Thank you!

PR #2405, opened 2023-10-18
change typescript link title for consistency

fennewald:
## What does this PR do?

Fixes a tiny nit with https://docs.temporal.io/temporal

When linking to the various SDKs, the typescript link has a different format than the other languages.


fennewald:
Force push because I accidentally only changed the generated file 😅

djmagee (approved):

PR #2417, opened 2023-10-19
migrate failures and troubleshooting articles out of kb

axfelix:
Based on our conversation around the KB going away and the desire to have a dedicated troubleshooting section... here's a first pass at a few of the most obvious articles to be moved.


rachfop:
Do we want to create a perma redirect link for the older KB articles?

rachfop (approved):
Look into the redirects before merging; otherwise I don't see why we can't get this in.

PR #2422, opened 2023-10-20
create new guide directories if they don't exist

axfelix:
creates new guide directories as part of the assembly process


flossypurse:
😃

flossypurse (approved):
Nice!

PR #2432, opened 2023-10-24
add clarification around activity replay

axfelix:


docs-src/concepts/what-is-an-activity-definition.md
```diff
@@ -30,6 +30,12 @@ Temporal recommends that Activities be idempotent.
 Idempotent means that performing an operation multiple times has the same result as performing it once.
 In the context of Temporal, Activities should be designed to be safely executed multiple times without causing unexpected or undesired side effects.
 
+:::info
+
+By design, completed Activities will not re-execute as part of a [Workflow Replay](/workflows#replays). However, Activites won’t record to the [Event History](/retry-policies#event-history) until they return or produce an error. If an Activity fails to report to the server at all, it will be retried. Designing for idempotence, especially if you have a [Global Namespace](/namespaces#global-namespace), will improve reusability and reliability.
```
rachfop, line 35:
```suggestion
By design, completed Activities will not re-execute as part of a [Workflow Replay](/workflows#replays). However, Activities won’t record to the [Event History](/retry-policies#event-history) until they return or produce an error. If an Activity fails to report to the server at all, it will be retried. Designing for idempotence, especially if you have a [Global Namespace](/namespaces#global-namespace), will improve reusability and reliability.
```

rachfop (approved):

PR #2438, opened 2023-10-25
Elaborate on the zero-downtime CA cert update

rachfop:
## What does this PR do?

After reviewing ticket edu-1033, it became clear that the documentation was clear; however....

This is used to expand and tweak the docs just a bit.


flossypurse (approved):
nice,

PR #2453, opened 2023-11-01
Adds note on Child Workflows and Continue-As-New

rachfop:
## What does this PR do?

Child Workflows do not carry over when the Parent uses [Continue-As-New](/concepts/what-is-continue-as-new).
This implies that if a Parent Workflow Execution utilizes Continue-As-New, any ongoing Child Workflow Executions will not be retained in the new continued instance of the Parent.


Sushisource (approved):
Thanks!

docs-src/concepts/what-is-a-child-workflow-execution.md
```diff
@@ -25,6 +25,13 @@ A Parent Workflow Execution must await on the Child Workflow Execution to spawn.
 The Parent can optionally await on the result of the Child Workflow Execution.
 Consider the Child's [Parent Close Policy](/concepts/what-is-a-parent-close-policy) if the Parent does not await on the result of the Child, which includes any use of Continue-As-New by the Parent.
 
+:::note
+
+Child Workflows do not carry over when the Parent uses [Continue-As-New](/concepts/what-is-continue-as-new).
+This implies that if a Parent Workflow Execution utilizes Continue-As-New, any ongoing Child Workflow Executions will not be retained in the new continued instance of the Parent.
```
axfelix, line 31: I don't know if "implies" is the best word here -- it's kind of ambivalent when we're writing about our own features. I would lose "This implies that" from the second sentence, just start with "If a Parent..."

flossypurse, line 31:
```suggestion
This means that if a Parent Workflow Execution utilizes Continue-As-New, any ongoing Child Workflow Executions will not be retained in the new continued instance of the Parent.
```

flossypurse (approved):

PR #2457, opened 2023-11-02
Move KB article and create How it works docs

rachfop:
## What does this PR do?

Removes a KB article
Creates a "How does it work" doc 

adds a redirect for kb article

Add async vs sync docs


flossypurse (changes requested):
I am totally ok with doing this per SDK.

However, is it possible to make this 1 page about how it works? (1 guide config how-it-works.json for example) 

We may not need to completely rewrite these, but I do think at minimum we need to align this as an extension of the narrative started by the `/sdks` page -

Right now, these 2 articles provide focused information from specific Temporal centric perspectives. I am hoping to also improve the connection to where the developer is.

PR #2462, opened 2023-11-07
fix(python-dev-guide): make client use the relevant namespace + adjust queue naming

galkleinman:
## What does this PR do?
1. Make the client use `backgroundcheck_namespace` which is used along the guide, otherwise `default` one will be used and workflows won't be executed
2. Remove a dash in the queue name (background**-**check-boilerplate-task-queue) to match the queue name which is later used in the CLI execute workflows command. (matching the queue names in the other SDK examples as well) 

## Notes to reviewers

Spent couple of hours to find the namespace issue (my first time trying Temporal), hope no one will ever experience it again :)


rachfop:
Hey, thanks.
I edited the source files here:
https://github.com/temporalio/documentation/pull/2464

PR #2479, opened 2023-11-16
Clarify Update disabled note

lorensr:
Clarify where they're disabled, that it's not a permanent thing, and add article


rachfop:
Thanks for the fix!

rachfop (approved):

PR #2487, opened 2023-11-20
Adds timers to concept page

rachfop:
## What does this PR do?

EDU-1411

### DIFF
--- a/docs-src/go/landing-page/go-dev-guide-structure.md
+++ b/docs-src/go/landing-page/go-dev-guide-structure.md
@@ -33,7 +33,6 @@ The Durable Execution section of the Temporal Developer's guide covers advanced 
 
 The Foundations section of the Temporal Go SDK Developer's guide covers the minimum set of concepts and implementation details needed to build and run a Temporal Application in Go – that is, all the relevant steps to start a Workflow Execution that executes an Activity.
 
-
 - [How to install Temporal CLI and run a development server](/self-hosted/how-to-install-temporal-cli): How to install Temporal CLI and run a development Cluster.
 - [How to install a Temporal SDK](/go/add-sdk): A Temporal SDK provides a framework for Temporal Application development.
 - [How to connect a Temporal Client to a Temporal Cluster](/go/connect-to-a-dev-cluster): When connecting a Temporal Client to a Temporal Cluster, you must provide the address and port number of the Temporal Cluster.


docs-src/concepts/what-is-a-timer.md
```diff
@@ -0,0 +1,24 @@
...
+sidebar_label: Timer
+description: A Workflow can set a durable timer for a fixed time.
+tags:
+  - term
+  - explanation
+---
+
+A Workflow can set a durable timer for a fixed time.
+Timers in Temporal are persisted, meaning that even if your Worker or Temporal Cluster is down when the time period completes, as soon as your Worker and Cluster are back up, the `sleep()` or `timer()` call will resolve and your code will continue executing.
+This makes it a reliable and resource-light operation that does not tie up the process, and you can run millions of Timers off a single Worker.
```
tomwheeler, line 13: I wonder if "tie up" and "run ... off" would be clear to non-native English speakers. Consider this alternate phrasing, which may be more clear:

> Timers are reliable and efficient. Workers consume no additional resources while waiting for a Timer to fire, so a single Worker can await millions of Timers concurrently.

tomwheeler (changes requested):
I suggested some alternate wordings. I think that, even without those suggested changes, this PR will improve the documentation on Timers.

However, I can't help but think that it misses the original request from @cretz to some degree. I think he was looking for the docs to explain what happens during a Worker crash. You summarize the end result (i.e., everything's OK), but don't elaborate on it. Chad would be in a better position than me to decide whether that's sufficient.

docs-src/concepts/what-is-a-timer.md
```diff
@@ -0,0 +1,25 @@
...
+id: what-is-a-timer
+title: What is a Timer?
+sidebar_label: Timer
+description: A Workflow can set a durable timer for a fixed time.
+tags:
+  - term
+  - explanation
+---
+
+You can add a delay to your Workflow Execution by setting a Timer.
```
flossypurse, line 11: Currently, I think this sounds a lot like "Start Delay".

I think we either we want to be generically accurate and say something like "SDKs offer Timer APIs so that Workflow Executions are deterministic in their handling of time values".

Or.. say something highly contextualized like
"You can sleep inside a Workflow Execution for seconds, minutes, or years."
And then break it down from there.

docs-src/concepts/what-is-a-timer.md
```diff
@@ -0,0 +1,25 @@
+---
+id: what-is-a-timer
+title: What is a Timer?
+sidebar_label: Timer
+description: A Workflow can set a durable timer for a fixed time.
```
flossypurse, line 5: I think this could either be slightly more accurate (see my comment below) or more descriptive of the expectations someone might have with the content such as "Learn more about Temporal's Durable Timers"

flossypurse (approved):
Definitely would like to see my comments addressed, before merging but glad to have the content overall -

PR #2488, opened 2023-11-22
go SDK - Instead of nil activity, use method expression

karelbilek:
Method expressions are nicer and get rid of the nil object.

They are not that widely known feature of go - they basically de-curry a method and make it a function where the first parameter is the receiver - https://go.dev/ref/spec#Method_expressions.

As Temporal SDK doesn't really care about the implementation and just looks at the method name through runtime and reflection, this still works, and one gets rid of the "nil object" pattern.

### DIFF
--- a/docs-src/go/generated/how-to-spawn-an-activity-execution-in-go.md
+++ b/docs-src/go/generated/how-to-spawn-an-activity-execution-in-go.md
@@ -34,11 +34,10 @@ func YourWorkflowDefinition(ctx workflow.Context, param YourWorkflowParam) (*You
 		ActivityParamX: param.WorkflowParamX,
 		ActivityParamY: param.WorkflowParamY,
 	}
-	// Use a nil struct pointer to call Activities that are part of a struct.
-	var a *YourActivityObject
 	// Execute the Activity and wait for the result.
 	var activityResult YourActivityResultObject
-	err := workflow.ExecuteActivity(ctx, a.YourActivityDefinition, activityParam).Get(ctx, &activityResult)
+	// Use method expression to call Activities that are part of a struct.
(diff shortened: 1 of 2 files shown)


axfelix:
Hey @karelbilek ! Sorry this sat unmerged for a while. Can you take a look at the outstanding merge conflict at some point so we can try to merge it this or next week?

btw, I appreciate you running our assembly workflow! You can also commit the `ASSEMBLY_REPORT.md` file that gets produced in the project root — we use that to make sure that all the compile time changes are captured.

karelbilek:
I'm on holiday without computer for next few weeks. Sorry!

Just want to make sure, is this something that you actually want to merge?
I "just" fired this PR without much discussion.. just want to make sure
this will now be recommended way to call activity objecy.

Karel

axfelix:
I like it — pinging @tomwheeler for thoughts as we want to update this in a few more places across our docs / lessons for consistency.

flossypurse:
The actual change for this needs to be made in the source sample which I believe is this file:
https://github.com/temporalio/documentation/blob/main/sample-apps/go/yourapp/your_workflow_definition_dacx.go

But just a reminder to update the line counts in the appropriate DACX configs IF the number of lines in the file change.

flossypurse:
> The actual change for this needs to be made in the source sample which I believe is this file: https://github.com/temporalio/documentation/blob/main/sample-apps/go/yourapp/your_workflow_definition_dacx.go
> 
> But just a reminder to update the line counts in the appropriate DACX configs IF the number of lines in the file change.

On side note -
If this seems complex - it is a little bit :)

We have plans to retire DACX and revert to using the Snipsync tooling for all of this "source snippet" generation in the near future.
So, changes like this will be much easier when that happens.

PR #2513, opened 2023-12-12
Adds a Hash function for id's in Link Source HTML

rachfop:
## What does this PR do?

Creates a deterministic ID based on the link provided.

The `hashCode` function will always return the same hash value for the same input string. This is because it's a deterministic function: given the same input, it will always produce the same output. So if you pass the same link string to the `hashCode` function, it will always return the same hash value, which you're using as an ID.

```js
  function hashCode(s) {
    let h = 0;
    for (let i = 0; i < s.length; i++) {
      h = (Math.imul(31, h) + s.charCodeAt(i)) | 0;
    }
    return h;
  }
```


flossypurse (approved):
ok sweet. lets try it!

PR #2503, opened 2023-12-06
edu-1036 | TypeScript SDK | Project setup chapter

napcs:
## What does this PR do?

Adds the project setup chapter to the TypeScript dev guide.

PR #2478, opened 2023-11-16
Adds TypeScript's durable execution

rachfop:
[Preview](https://temporal-documentation-git-typescript-durable-execution.preview.thundergun.io/dev-guide/typescript/durable-execution)

PR #2476, opened 2023-11-15
Adds Python section for Replay

rachfop:

PR #2518, opened 2023-12-18
EDU-1036 | Dev guide content release | Python + TypeScript chapters

flossypurse:
**Preview**
The following pages are new:
- https://temporal-documentation-git-chapters-release.preview.thundergun.io/dev-guide/python/durable-execution
- https://temporal-documentation-git-chapters-release.preview.thundergun.io/dev-guide/typescript/project-setup
- https://temporal-documentation-git-chapters-release.preview.thundergun.io/dev-guide/typescript/durable-execution

Includes commits and contributions from the following:
- https://github.com/temporalio/documentation/pull/2476
- https://github.com/temporalio/documentation/pull/2503
- https://github.com/temporalio/documentation/pull/2478


docs-src/typescript/generated/backgroundcheck-boilerplate-workflow-details.md
```diff
@@ -10,7 +10,7 @@ tags:
 ---
 
 <!-- DO NOT EDIT THIS FILE DIRECTLY.
-THIS FILE IS GENERATED from https://github.com/temporalio/documentation-samples-typescript/blob/main/chapter_durable_execution/backgroundcheck/src/workflows.ts. -->
+THIS FILE IS GENERATED from https://github.com/temporalio/documentation-samples-typescript/blob/fixes2/chapter_project_setup/backgroundcheck/src/workflows.ts. -->
 
 Temporal Workflows may have any number of custom parameters. However, we
 strongly recommend that you use objects as parameters, so that the object's
```
axfelix, line 16: This might just be me, but for all the different places we've tried to call this out (and move this guidance around), I feel like its inclusion here is a little too brief and unclear. "Use objects as parameters" doesn't even quite get at the rationale here; I'm not sure someone reading this for the first time would understand it as "try to supply all the input to the function as a single parameter (e.g. an object with multiple values), because changing the function's definition can cause determinism errors."

docs-src/typescript/chapter-project-setup/workflow-sandbox.md
```diff
@@ -0,0 +1,30 @@
...
+If you **must** use a library that references a Node.js or DOM API and you are certain that those APIs are not used at runtime, add that module to the [ignoreModules](https://typescript.temporal.io/api/interfaces/worker.BundleOptions#ignoremodules) list.
+
+:::
+
+In the TypeScript SDK, you can safely use non-deterministic methods and rely on the sandbox to replace non-deterministic code with deterministic versions. However side effects and access to external state must be done through Activities because Activity outputs are recorded in the Event History and can read deterministically by the Workflow.
+
+This limitation also means that Workflow code cannot directly import the [Activity Definition](/concepts/what-is-an-activity-definition).
+[Activity Types](/concepts/what-is-an-activity-definition#activity-type) can be imported, so they can be invoked in a type-safe manner.
+
+To make the Workflow runtime deterministic, functions like `Math.random()`, `Date`, and `setTimeout()` are replaced by deterministic versions.
```
axfelix, line 30: Similar comment here to my other one... I think we've tried to cram a bit too much into the project setup chapter here and we're only spending like one sentence on a few important SDK concepts that could use at least a couple. The SDK replacing `Math.random()` with its own calls should be noted a little more prominently if at all.

axfelix:
General comment -- I think it's too early in the dev guide rewrite process to start renaming the side nav pages to "Ch.". I don't know if it'll help a reader to understand that some of these are "chapters" right now and others aren't.

docs-src/typescript/chapter-project-setup/choose-dev-cluster.md
```diff
@@ -0,0 +1,133 @@
...
+
+:::
+
+### Self-hosted Temporal Cluster
+
+We recommend using a self-hosted environment if you are starting something new and need to scale with production-level features, but don’t yet need or want to pay for Temporal Cloud.
+
+For example, running a self-hosted Cluster lets you try different databases, view Cluster metrics, use custom [Search Attributes](/concepts/what-is-a-search-attribute), and even play with the [Archival](/concepts/what-is-archival) feature.
+
+It's completely possible to run a production level self-hosted Cluster.
```
axfelix, line 101: This is in a couple different dev guides right now and the use of "completely" here feels a bit off -- it's one of those adverbs that accidentally sounds more caveated by its inclusion.

axfelix: (Also in Python)

docs-src/typescript/generated/intrinsic-non-deterministic-logic-example.md
```diff
@@ -1,18 +1,18 @@
...
 description: BackgroundCheckNonDeterministic Workflow is a non-deterministic Workflow as it uses a random number generator.
 ---
 
 <!-- DO NOT EDIT THIS FILE DIRECTLY.
-THIS FILE IS GENERATED from https://github.com/temporalio/documentation-samples-typescript/blob/main/chapter_durable_execution/backgroundcheck_nondeterministic/src/workflow_dacx.ts. -->
+THIS FILE IS GENERATED from https://github.com/temporalio/documentation-samples-typescript/blob/fixes2/chapter_durable_execution/backgroundcheck_nondeterministic/src/workflow_dacx.ts. -->
 
 One way to produce a non-deterministic error is to use a random number to determine whether to sleep inside the Workflow.
 
 If you run the `BackgroundCheckNonDeterministic` Workflow enough times, eventually you will see a Workflow Task failure.
```
axfelix, line 13: In Go this sentence comes after the subsequent code snippet, and I think it makes more sense to do it that way... I would also rephrase it slightly to "eventually, the way that sleep has been implemented here will produce a non-determinism error."

docs-src/typescript/generated/backgroundcheck-boilerplate-workflow-details.md
flossypurse: 1. This gets hit hard over the head with a mallet in multiple places and it is primarily a concern when versioning. So thats when we will spend more time on it.
2. Its not as important as we make it out to be to get started. It just isn't - I can build a huge app and learn a lot about Temporal without fussing about Object based params for versioning considerations - which in the end serves to understand Replay + Production level versioning.

We simply can not go into depth on everything all the time.

docs-src/typescript/chapter-project-setup/workflow-sandbox.md
flossypurse: This is all about the design of the SDK.
Temporal introduces the need for deterministic constraints - then we went and made SDKs that try to "hide" a bunch of that - allowing someone to get super deep into app development without understanding this until they suddenly need to...

There is a "base" idea of determinism due to Replay. 
This feels incomplete because it is.
I pushed for better "How the TypeScript SDK works" to supplement this narrative - but we have yet to get that.

But also, this content is meant to be more practical than conceptual - always has been. So, it really isn't the place for deep dive into concepts.

flossypurse:
> General comment -- I think it's too early in the dev guide rewrite process to start renaming the side nav pages to "Ch.". I don't know if it'll help a reader to understand that some of these are "chapters" right now and others aren't.

Then we aren't releasing it properly.
Again, this project suffers from crazy scope change due to changes in priorities before we can finish anything.

docs-src/typescript/chapter-project-setup/choose-dev-cluster.md
flossypurse: Ya, we can nix the completely.

docs-src/typescript/generated/intrinsic-non-deterministic-logic-example.md
flossypurse: due to DACX, moving this to after the code snippet is too many changes - and given the new direction... i am not sure worth the effort.

But I did update the sentence to your suggestion

PR #2573, opened 2024-01-17
Adding maximum task timeout

jsundai:
## What does this PR do?
Currently, it notes the default value of a Workflow Task Timeout. This adds the maximum task timeout. 

## Notes to reviewers
This is linked to JIRA ticket [EDU-1424](https://temporalio.atlassian.net/browse/EDU-1424).

[EDU-1424]: https://temporalio.atlassian.net/browse/EDU-1424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


flossypurse (approved):

docs-src/concepts/what-is-a-workflow-task-timeout.md
```diff
@@ -21,4 +21,5 @@ A Workflow Task Timeout is the maximum amount of time allowed for a [Worker](/co
 
 **The default value is 10 seconds.**
 This timeout is primarily available to recognize whether a Worker has gone down so that the Workflow Execution can be recovered on a different Worker.
-The main reason for increasing the default value would be to accommodate a Workflow Execution that has a very long Workflow Execution History that could take longer than 10 seconds for the Worker to load.
+The main reason for increasing the default value is to accommodate a Workflow Execution that has an extensive Workflow Execution History, requiring more than 10 seconds for the Worker to load.
+It's worth mentioning that although you can extend the timeout up to the maximum value of 120 seconds, it's not recommended to move beyond the default value.
```
flossypurse, line 25: This is good for now.

Though, I can infer from reading the first sentence that raising the timeout increases the time it would take to identify a Worker that has gone down.

But I then begin to wonder - How many Events in a History fit into the 10 second range, vs the 120 second range?

What is an "extensive" Workflow History, and how is that different from a "long" Workflow History -
Perhaps calling out an explicit number range would be far more effective here.

This is definitely good theory - but it does lack specifics the more that I explore it.
Curious if it would be worth asking someone from the SDK + Server engineering teams about this - and if we can come up with an actual range?

flossypurse (approved):
this is good enough for now.

PR #2574, opened 2024-01-18
Remove `node` in `CMD` because this image does not need it

vlechemin:


mjameswh:
It's true that specifying `node` is not required on this image. Yet, explicitly specifying it does no harm either.

I'm concerned that users may use some copy parts of that Dockerfile, but use in conjunction with different base images, in which case the `node` executable may no longer be implicit.

@vlechemin Is there any specific reason why you think we should _not_ specify `node` explicitly?

flossypurse:
@vlechemin appreciate the contribution, though we are closing this as @mjameswh said, there is no harm and being explicit can only help.

PR #2590, opened 2024-01-24
Adding .NET API link to References page

jsundai:
## What does this PR do?
Adds the [.NET API Reference](https://dotnet.temporal.io/api/) link to list in [References](https://docs.temporal.io/references/). 

This can be referenced back to for .NET PRs for the Dev Guide with the assembly/config.json file change for new .NET pages. 

## Notes to reviewers
Linked to JIRA ticket 1453.


assembly/config.json
```diff
@@ -3,6 +3,7 @@
         "docs-src/cloud",
         "docs-src/concepts",
         "docs-src/dev-guide",
+        "docs-src/dotnet",
```
cretz, line 6: Question - is there any harm in referencing a dir that doesn't exist? Would it make more sense to wait until the PR that adds this dir to add this entry?

flossypurse (approved):

PR #2613, opened 2024-02-01
Note Schedule stage

lorensr:
As a user, and I find there's a new feature, I'd like to know what stage it's in, to help me decide whether to use it. Also linking Server version so I know when it was first released. ('22, which gives me confidence)


docs-src/concepts/what-is-a-schedule.md
```diff
@@ -7,7 +7,8 @@ tags:
   - term
   - explanation
 ssdi:
-  - Introduced in Temporal Server version 1.17.0
+  - Generally Available as of Nov '23
```
flossypurse, line 10:
```suggestion
  - Generally Available as of Nov 2023
```

flossypurse (approved):

PR #2614, opened 2024-02-01
Edu 1799

fairlydurable:
## What does this PR do?

Removes reference and obscure Mac-only comment about `tctl`. Adds multi-platform note about newer CLI tool with link. Wasn't sure whether or not to remove it entirely.

## Note to reviewers

Asked Patrick yesterday about the weird key management inclusion and he wasn't worried. Just be aware that it is there, thanks.


fairlydurable:
Closing, because other docs still mention tctl

PR #2616, opened 2024-02-02
Sasha/oss 1858

ast2023:
## What does this PR do?
Update documentation to caution about unexpected schedule behavior


flossypurse (approved):

docs-src/concepts/what-is-a-schedule.md
```diff
@@ -28,6 +28,12 @@ Schedules provide a more flexible and user-friendly approach than [Temporal Cron
 A Schedule has an identity and is independent of a Workflow Execution.
 This differs from a Temporal Cron Job, which relies on a cron schedule as a property of the Workflow Execution.
 
+:::caution
+
+If the result returned by the scheduled workflow exceeds max block size minus temporal extra information (2MB - 1KB) this result will not be available for the next run.
```
flossypurse, line 33:
```suggestion
If the result returned by the scheduled Workflow Execution exceeds max block size minus temporal extra information (2MB - 1KB) this result will not be available for the next run.
```

What is "temporal extra information"?

ast2023: @flossypurse Thank you for the comment. Temporal wraps results produced by the scheduled workflow in an envelope with some extra information which allows it to find the result and provide it to the next run of the workflow.  Do you think we should explain it here?

flossypurse: Well, either remove the mention of the "extra information" completely - because the developer doesn't actually need to worry about that, or yes specify what we mean by that - perhaps something like:

> If the result returned by the scheduled Workflow Execution exceeds max block size (2MB - 1KB) this result will not be available for the next run. The max block size is not affected by the extra meta data that Temporal needs to identify the result and provide it to the next Workflow Execution.

docs-src/concepts/what-is-a-schedule.md
```diff
@@ -28,6 +28,12 @@ Schedules provide a more flexible and user-friendly approach than [Temporal Cron
 A Schedule has an identity and is independent of a Workflow Execution.
 This differs from a Temporal Cron Job, which relies on a cron schedule as a property of the Workflow Execution.
 
+:::caution
+
+The Temporal server enriches responses from scheduled Workflow Executions with additional contextual data, enabling their identification and handoff to subsequent executions. However, if the returned result exceeds the designated block size by temporal overhead (e.g., 2MB - 1KB), it becomes unavailable for use in the subsequent run.
```
flossypurse, line 33:
```suggestion
The Temporal Service enriches responses from scheduled Workflow Executions with additional contextual data, enabling their identification and handoff to subsequent executions.
However, if the returned result exceeds the designated block size by Temporal overhead (e.g., 2MB - 1KB), it becomes unavailable for use in the subsequent run.
```

- We have updated guidance from Product Marketing to use Temporal Service generically as often as possible unless it requires a specific callout to the server.
- Capitalize core terms
- I don't know what "Temporal overhead" means here. 
- Each sentence on a newline.

If my understanding is correct, that my recommendation to provide a simple and approachable message would be more like this:
> If the result returned by the scheduled Workflow Execution exceeds max block size (2MB - 1KB) this result will not be available for the next run.
The max block size is not affected by the extra meta data that Temporal needs to identify the result and provide it to the next Workflow Execution.

flossypurse (changes requested):

docs-src/concepts/what-is-a-schedule.md
dnr: First, this is an unlikely edge case and does not deserve to be in this high position on the page. It should be down around line 203 (pre-change), at the end of the `### Last completion result` section.

As for the text, I don't really like either of those :slightly_smiling_face:   I'd write

```suggestion
A scheduled workflow may complete with a result up to the maximum blob size (2 MiB by default). However, due to internal limitations, results that are within 1 KiB of this limit cannot be passed to the next execution. So, for example, a workflow that returns a result of size 2,097,000 bytes will be allowed to compete successfully, but that value will not be available as a last completion result. This limitation may be lifted in the future.
```

(capitalize as appropriate)

dnr: Last, the thing about last completion and AllowAll is going to be wrong soon and should be removed, that's lines 176-190 pre-change.

docs-src/concepts/what-is-a-schedule.md
```diff
@@ -201,6 +201,15 @@ It is important to note that the [status](/concepts/what-is-a-workflow-execution
 
 :::
 
+:::caution
+
+A scheduled workflow may complete with a result up to the maximum blob size (2 MiB by default).
+However, due to internal limitations, results that are within 1 KiB of this limit cannot be passed to the next execution.
+So, for example, a workflow that returns a result of size above 2,096,127 (2MiB - 1KiB) bytes will be allowed to compete successfully, but that value will not be available as a last completion result.
```
dnr, line 208: Reading strictly, this says that a workflow that returns any size above 2,096,127 will be allowed to complete successfully. That's not true, it has to be <= 2097152. That's why I wrote it with a specific number as an example. Written this way, it's no longer an example, it's just a statement.

Either make it use an example (I think this is cleanest and prefer it), or else reword it to be strictly correct with the min and max and remove mention of "example"

flossypurse: Also, please capitalize core terms like "Workflow".
And specify "Workflow Execution".

docs-src/concepts/what-is-a-schedule.md
```diff
@@ -203,9 +203,9 @@ It is important to note that the [status](/concepts/what-is-a-workflow-execution
 
 :::caution
 
-A scheduled workflow may complete with a result up to the maximum blob size (2 MiB by default).
+A scheduled Workflow Execution may complete with a result up to the maximum blob size (2 MiB by default).
 However, due to internal limitations, results that are within 1 KiB of this limit cannot be passed to the next execution.
-So, for example, a workflow that returns a result of size above 2,096,127 (2MiB - 1KiB) bytes will be allowed to compete successfully, but that value will not be available as a last completion result.
+So, for example, a Workflow Execution that returns a result of size 2.1MiB (which is above 2MiB - 1KiB limit) will be allowed to compete successfully, but that value will not be available as a last completion result.
```
dnr, line 208: No, a result of size 2.1MiB would not be allowed to complete since it's above the blob size limit. The whole point is we're warning about the tiny window in between the blob size limit and the "last completion" limit.

dnr (approved):

dnr:
Since you're already editing this file, could you also do this: https://github.com/temporalio/documentation/pull/2616#discussion_r1487146132

flossypurse (approved):

PR #2617, opened 2024-02-02
Make assemble and worker to read server address from environment. 

ast2023:
## What does this PR do?
Enhance worker.js and assemble.js to allow them to run on a remote machine. Temporal server address can be specified through TEMPORAL_SERVER_ADDRESS environment variable.

## Notes to reviewers
I dockerized node and yarn. I also created a script which performs both the `yarn install`. This will allow to setup environment with a single docker compose command. To run the environment inside docker, I need to be able to specify the address for the worker to connect. This is the reason for this pull request.

### DIFF
--- a/assemble.js
+++ b/assemble.js
@@ -77,7 +77,9 @@ async function useCloud(params) {
 }
 
 async function useLocal(params) {
-  const connection = await Connection.connect({});
+  const connection = await Connection.connect(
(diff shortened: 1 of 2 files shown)


flossypurse:
This edits the default which relies on the CLI `temporal server start-dev` command, and so I will be rejecting this suggestion.

There is already an option to run off of Temporal Cloud.

So, If you want to add a NEW option to run using Docker - then we can accept that. However, if i remember correctly the default network that the docker-compose repo uses enables the use of it with the local option as well...
So, I am not sure how much we really want to consider this.

PR #2629, opened 2024-02-13
Add note about Side Effect support

lorensr:
answer questions folks have about why it's not implemented in other sdks and what to do if it's not part of their sdk


docs-src/concepts/what-is-a-side-effect.md
```diff
@@ -8,6 +8,13 @@ tags:
   - explanation
 ---
 
+:::note
+
+Side Effects are included in the Go, Java, and PHP SDKs.
+They are not included in other SDKs, as [Local Activities](/concepts/what-is-a-local-activity) fit the same use case and are slighly less resource intensive.
```
flossypurse, line 14:
```suggestion
They are not included in other SDKs.
[Local Activities](/concepts/what-is-a-local-activity) fit the same use case and are slightly less resource intensive.
```

flossypurse (approved):

PR #2632, opened 2024-02-13
Removes pre-built workflow bundles section

fairlydurable:
Removes section about pre-built workflow bundles by removing content node from foundations.json. The source file is left in-place as an unconnected node.

Updated as discussed with @flossypurse


fairlydurable:
Thank you @rachfop. You are amazing.

flossypurse (approved):

PR #2635, opened 2024-02-13
Adds Schedule Metrics to docs

rachfop:


dnr (approved):
This isn't supposed to be 100% comprehensive on metrics, right? There are a few more like `schedule_cancel_workflow_errors` (or `terminate`), `schedule_action_errors`, and there's the new `schedule_action_delay` (not released yet)

I think it's fine to document the more important ones only.

PR #2636, opened 2024-02-14
Adds Visiblity Limit to docs

rachfop:

### DIFF
--- a/ASSEMBLY_REPORT.md
+++ b/ASSEMBLY_REPORT.md
@@ -1,7 +1,5 @@
 # Docs Assembly Workflow report
 
-Last assembled: Wednesday February 14 2024 09:16:37 AM -0800
-
 Assembly Workflow Id: docs-full-assembly
 
 139 guide configurations found.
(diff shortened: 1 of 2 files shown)


tomwheeler (approved):
LGTM

docs-src/cloud/limits.md
```diff
@@ -132,6 +132,12 @@ When creating custom Search Attributes in Temporal Cloud, the attribute names mu
 
 For more information on custom Search Attributes see [Custom Search Attributes limits](/visibility#custom-search-attributes).
 
+### Visibility API Rate Limit
+
+**What is the rate limit for requests to the Visibility APIs?**
+
+The default rate limit for requests to the Visibility APIs is 10 requests per second.
```
pdoerner, line 139: This is a per-namespace limit; does that need to be explicit here?

pdoerner (approved):

docs-src/cloud/limits.md
tomwheeler: I think that's an important clarification to include.

PR #2639, opened 2024-02-14
Adds updated action-per-second wording

fairlydurable:
Updates the following files with modified wording that omits mentions of spiking and explains about adjusting thresholds. The messaging is consistent across all three documents with additional original material retained and incorporated in the throughput expectations. Throughput expectations is now in sync with the wording of the activity pricing page.

* cloud/limits.md
* cloud/throughput-expectations.md
* cloud/namespace-setup-considerations.md


docs-src/cloud/limits.md
```diff
@@ -63,12 +63,11 @@ The following aspects apply at the Namespace level (per Namespace).
...
 **What is the limit of Actions per second?**
 
-The throughput of a Namespace is measured in Actions per second (APS).
-
-The default limit is 200 APS with the ability to spike to 400 APS.
-
-The APS limit is automatically adjusted based on the actual usage over the trailing 7 days.
-The lowest APS limit that ever gets set is 200 APS.
+Each Namespace has a rate limit, which is measured in Actions per second (APS).
+A namespace may be throttled when its throughput becomes too high.
```
flossypurse, line 67:
```suggestion
A Namespace may be throttled when its throughput becomes too high.
```

docs-src/cloud/limits.md
```diff
@@ -63,12 +63,11 @@ The following aspects apply at the Namespace level (per Namespace).
...
+Throttling means limiting the rate at which actions are performed to prevent overloading the system.
+A Namespace's default limit is set at 200 APS and automatically adjusts based on recent usage (over the prior 7 days).
+Your throughput limit will never fall below this default value.
```
flossypurse, line 70: The 200 APS value or the value that it automatically adjusts to?

For example, if it adjust upward to 300 APS, will the system ever decrease that?
Or will it never decrease below 200?

fairlydurable: Latest info is "Regarding automatic adjustment, it’s not correct to say it’s adjusting down - as we’ll never adjust to lower than default (unless customer requests it) and automatic adjustment will only bump the limit higher than default based on usage." My understanding is that it can revert slowly back down but never go below that default.

flossypurse (approved):

docs-src/cloud/limits.md
```diff
@@ -63,12 +63,11 @@ The following aspects apply at the Namespace level (per Namespace).
...
 
-The throughput of a Namespace is measured in Actions per second (APS).
-
-The default limit is 200 APS with the ability to spike to 400 APS.
-
-The APS limit is automatically adjusted based on the actual usage over the trailing 7 days.
-The lowest APS limit that ever gets set is 200 APS.
+Each Namespace has a rate limit, which is measured in Actions per second (APS).
+A Namespace may be throttled when its throughput becomes too high.
+Throttling means limiting the rate at which actions are performed to prevent overloading the system.
```
lorensr, line 68:
```suggestion
Throttling means limiting the rate at which Actions are performed to prevent the Namespace from exceeding its APS limit.
```

don't want to give the impression that going over 200 APS would overload the system. also, imagine it would take all namespaces in a cluster going significantly over their limits at the same time to actually impact the cluster performance

docs-src/cloud/limits.md
```diff
@@ -63,12 +63,11 @@ The following aspects apply at the Namespace level (per Namespace).
...
+A Namespace's default limit is set at 200 APS and automatically adjusts based on recent usage (over the prior 7 days).
+Your throughput limit will never fall below this default value.
```
lorensr, line 70:
```suggestion
Your APS limit will never fall below this default value.
```

"APS Limit" is the title used in the UI—suggest using it throughout instead of "throughput limit"

docs-src/cloud/limits.md
fairlydurable: Feedback from earlier iteration that did not make it in yet: https://github.com/temporalio/documentation/pull/2639

PR #2651, opened 2024-02-23
Transitioning "stack trace" to "call stack" where applicable

jsundai:
## What does this PR do?
Updating to refer to the capability as "Call Stack" instead of "Stack Trace" where relevant, while the implementation will still be denoted as "stack_trace." 

## Notes to reviewers
Linked to Jira [1505](https://temporalio.atlassian.net/browse/EDU-1505)


fairlydurable:
The Jira says "We are looking to improve the usability of the Stack Traces tab on the Workflow Details page to address user feedback, who don't intuit what "stack trace" means and assume it's an error.
We are looking to rename "Stack Trace" to "Call Stack" as well as adding additional context to the tab "This is a call stack showing each location where Workflow code is waiting." 
This would require renaming "Stack Trace" to "Call Stack" where applicable in the documentation, we want to refer to the capability in as “Call Stack” where the implementation will continue to be referred to as “stack_trace”.

Example where we want to make this change:
[Temporal Web UI | Temporal Documentation](https://docs.temporal.io/web-ui#stack-trace)  
Update this paragraph to reference the Web UI refers to this as “Call Stack”[What is a Temporal Workflow? | Temporal Documentation](https://docs.temporal.io/workflows#stack-trace-query)  
We do NOT want to replace this terminology holistically, as the CLI/API/SDK will continue to refer to this as “Stack Trace”"

After removing the CLI refs (sorry for dupes I missed), here is what I found

```
./go/error-handling.md
./references/failures.md
./cloud/what-is-an-action.md
./concepts/what-is-a-failure-converter.md
./concepts/what-is-the-temporal-web-ui.md
./concepts/what-is-a-query.md
./self-hosted/data-encryption.md
./php/queries.md
```

flossypurse (approved):

PR #2652, opened 2024-02-23
rename migration page and sidebar

axfelix:


fairlydurable (approved):
LGTM. I see no problems. Couldn't find a ticket, but I assume you know what you're doing and there are no typos I could find.

PR #2661, opened 2024-02-27
Adds coverage for Activity task execution guarantees

rachfop:


docs-src/concepts/what-is-an-activity-execution.md
```diff
@@ -23,6 +23,15 @@ You can customize [Activity Execution timeouts](/concepts/what-is-a-start-to-clo
 
 If an Activity Execution fails (because it exhausted all retries, threw a [non-retryable error](/concepts/what-is-a-retry-policy#non-retryable-errors), or was canceled), the error is returned to the [Workflow](/workflows), which decides how to handle it.
 
+:::note
+
+Temporal guarantees that an Activity Task either runs or timeouts.
+If the Activity Task times out, the Activity Execution will be retried according to the Activity Execution Retry Policy.
+
+In scenarios where the Activity Execution Retry Policy is set to `1` and a Timeout occurs, the Activity Execution will not be tried.
```
mfateev, line 31: tried or retried?

docs-src/concepts/what-is-an-activity-execution.md
```diff
@@ -23,6 +23,15 @@ You can customize [Activity Execution timeouts](/concepts/what-is-a-start-to-clo
 
 If an Activity Execution fails (because it exhausted all retries, threw a [non-retryable error](/concepts/what-is-a-retry-policy#non-retryable-errors), or was canceled), the error is returned to the [Workflow](/workflows), which decides how to handle it.
 
+:::note
```
mfateev, line 26: The above "By default, an Activity Execution has no time limit." in incorrect. The timeout is required.

docs-src/concepts/what-is-an-activity-execution.md
```diff
@@ -23,6 +23,15 @@ You can customize [Activity Execution timeouts](/concepts/what-is-a-start-to-clo
...
+
```
mfateev, line 27: I would say that:

There are multiple failure scenarios when activity task is lost. It can be lost during delivery to a worker or after the activity function was called and the worker crashed. 
Temporal doesn't detect task loss directly. It relies on StartToClose timeout instead.

fairlydurable (approved):
Looks good to me. I can't comment about technical accuracy but I see no issues with the text change.

PR #2664, opened 2024-03-05
Adds DataDog integration

rachfop:
## What does this PR do?

Provides a restructure of the Metrics section so we can adapt and add future integrations.


docs-src/cloud/datadog.md
```diff
@@ -0,0 +1,232 @@
...
+sidebar_label: DataDog
+tags:
+  - how-to
+  - observability
+  - metrics
+  - datadog
+  - temporal cloud
+---
+
+Exporting cloud metrics to DataDog provides enhanced observability, allowing you to monitor, alert, and visualize key performance indicators of your applications and infrastructure in real-time.
```
njo, line 14: It's not real-time, closer to about a 90 second delay.

docs-src/cloud/datadog.md
```diff
@@ -0,0 +1,232 @@
...
+Temporal's integration with DataDog makes it seamless to extend the monitoring capabilities of your Temporal Cloud deployment.
```
njo, line 15: Not sure I'd go so far to call it "seamless" 😆

njo (approved):
Thanks for this. Perhaps you could add a doc or mention the scrape version as well which is a prometheus compatible scrape endpoint (can also be scraped by the otel and datadog agents).
https://github.com/temporalio/samples-server/tree/main/cloud/observability/promql-to-scrape

PR #2668, opened 2024-03-07
move namespace param from Worker to Client -  project-setup.mdx

bkcsfi:
`Worker()` does not accept namespace, it must be passed to `Client.connect()` instead.

Since namespaces are introduced later in this document, a better fix might be to simply eliminate the namespace parameter from this example, rather than moving it to `Client.connect()`.


flossypurse:
@bkcsfi - Thank you!
I picked up your change here: https://github.com/temporalio/documentation/pull/2751
There was an additional spot in the source code that needed the namespace as well and we have since re-orged the docs which would have made merging or rebasing main a bit of pain - so that's why I picked it up for you.

PR #2681, opened 2024-03-14
Improve the documentation around generating certificates using Certstrap

rachfop:


flossypurse (approved):

dallastexas92 (commented):
In step 1, we should probably clarify that users can pick a certificate authority and they don't have to use "Cert". This is probably known, but worth clarifying

In step 6, can we state that the Cert.crt file is the one that will be uploaded to Temporal cloud?

Other than that, looks great!

PR #2683, opened 2024-03-14
setting up a Prometheus data source in Grafana

rachfop:

### DIFF
--- a/ASSEMBLY_REPORT.md
+++ b/ASSEMBLY_REPORT.md
@@ -1,6 +1,8 @@
 # Docs Assembly Workflow report
 
-Last assembled: Thursday March 14 2024 12:44:18 PM -0700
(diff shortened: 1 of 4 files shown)


flossypurse (approved):

dallastexas92 (commented):
I think this ticket is addressing feedback on the wrong document page, which is my fault because I referenced the wrong URL in the original JIRA ticket.

Here is the correct [docs page](https://docs.temporal.io/cloud/metrics/prometheus-grafana#data-sources-configuration-for-temporal-cloud-and-sdk-metrics-in-grafana) that has incorrect information

Additionally, this PR only addresses half of the feedback from that [ticket](https://temporalio.atlassian.net/browse/EDU-1795). The 2nd half of the feedback ties to importing the JSON file into Grafana to get the default cloud metrics

PR #2684, opened 2024-03-14
Adds statement regarding which supported env

rachfop:
## What does this PR do?
Its best practice to document what is, not what isn't.

### DIFF
--- a/ASSEMBLY_REPORT.md
+++ b/ASSEMBLY_REPORT.md
@@ -1,9 +1,5 @@
 # Docs Assembly Workflow report
 
-Last assembled: Monday March 18 2024 08:56:59 AM -0700
-
-Assembly Workflow Id: docs-full-assembly-rachfop-123
-
 147 guide configurations found.
 
 1778 information nodes found.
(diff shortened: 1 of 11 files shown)


flossypurse (approved):

docs-src/typescript/introduction-to-typescript-sdk.md
```diff
@@ -17,6 +17,8 @@ Welcome to the Temporal TypeScript SDK developer's guide!
 
 The Temporal TypeScript SDK released on July 26, 2022. The TypeScript SDK provides access to the Temporal programming model using idiomatic JavaScript and TypeScript programming paradigms.
 
+The Temporal TypeScript SDK supports the Node.js runtime environment.
```
mjameswh, line 20: Can we please add "only" here, or whatever else that would clearly mean the same thing?

We regularly get questions regarding Deno, Bun, browsers… I understand the idea that we want to document what is rather than what isn't, but the goal here is to make it clear that no other runtime environment is supported at this time.

```suggestion
The Temporal TypeScript SDK only supports the Node.js runtime environment.
```

PR #2685, opened 2024-03-15
Investigate confusing title in Cloud Docs

rachfop:
## What does this PR do?

The differentiates the two different workflows and approaches to managing namespace-level permissions in Temporal Cloud through the web UI.

### DIFF
--- a/ASSEMBLY_REPORT.md
+++ b/ASSEMBLY_REPORT.md
@@ -1,8 +1,5 @@
 # Docs Assembly Workflow report
 
-Last assembled: Thursday March 21 2024 09:06:43 AM -0700
-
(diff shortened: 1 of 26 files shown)


flossypurse (approved):

tomwheeler (commented):
"How to update a user's Namespace-level permissions across multiple Namespaces using the Web UI" improves the clarity, but it's a very long title. 

I think "How to use the Web UI to update a user's permissions across multiple Namespaces" would be shorter and perhaps just as clear, since "across multiple Namespaces" suggests that it's referring to Namespace-level permissions. However, Dallas was the one who reported it, so perhaps he can judge which he prefers. I am fine with either.

PR #2697, opened 2024-03-20
Point out that ActivityTaskStarted will not appear until Activity completed

fairlydurable:
The old branch seemed corrupted, so I'm starting another. 

Clarifies that against common-sense, the ActivityTaskStarted state does not appear in history until the Activity has completed.

This version is pulled back and uses minimal incursions, redirecting to the events page instead.

JIRA: EDU-1781 https://temporalio.atlassian.net/browse/EDU-1781

### DIFF
--- a/docs-src/references/events.md
+++ b/docs-src/references/events.md
@@ -237,7 +237,7 @@ This Event type contains Activity inputs, as well as Activity Timeout configurat
 
 This [Event](/concepts/what-is-an-event) type indicates that an [Activity Task Execution](/concepts/what-is-an-activity-task-execution) was started.
 The SDK Worker picked up the Activity Task and started processing the [Activity](/concepts/what-is-an-activity) invocation.
-The ActivityTaskStarted event is written into Workflow history when your activity completes or fails after all of its retries.
+The ActivityTaskStarted event is written into Workflow Event History when the Activity completes or fails after all of its retries.
 It may be counter-intuitive that this happens after the final Activity Event (like [ActivityTaskCompleted](/references/events#activitytaskcompleted) or [ActivityTaskFailed](/references/events#activitytaskfailed)).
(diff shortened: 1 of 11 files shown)


docs-src/references/events.md
```diff
@@ -237,7 +237,9 @@ This Event type contains Activity inputs, as well as Activity Timeout configurat
 
 This [Event](/concepts/what-is-an-event) type indicates that an [Activity Task Execution](/concepts/what-is-an-activity-task-execution) was started.
 The SDK Worker picked up the Activity Task and started processing the [Activity](/concepts/what-is-an-activity) invocation.
-Note, however, that this Event is not written to History until the terminal Event (like [ActivityTaskCompleted](/references/events#activitytaskcompleted) or [ActivityTaskFailed](/references/events#activitytaskfailed)) occurs.
+The ActivityTaskStarted event is written into Workflow history when your activity completes or fails after all of its retries.
```
flossypurse, line 240:
```suggestion
The ActivityTaskStarted event is written into Workflow Event History when the Activity completes or fails after all of its retries.
```

docs-src/references/events.md
```diff
@@ -237,7 +237,9 @@ This Event type contains Activity inputs, as well as Activity Timeout configurat
...
+It may be counter-intuitive that this happens after the final Activity Event (like [ActivityTaskCompleted](/references/events#activitytaskcompleted) or [ActivityTaskFailed](/references/events#activitytaskfailed)).
+Don't be misled into thinking that the activity is failing.
```
flossypurse, line 242:
```suggestion
Don't be misled into thinking that the Activity did not start if you don't see this Event in the History.
```

This feels like a strange warning. Updated to give it more context.

fairlydurable:
Temporarily closing.

PR #2699, opened 2024-03-20
Update pricing table for credit actions

napcs:


fairlydurable (approved):
Could not find any problems. You should be GTG

PR #2700, opened 2024-03-21
Update Cloud limits page

lorensr:
Only uses "default" when it's configurable. Also corrects some inaccuracies.

Fixes EDU-2115


docs-src/cloud/limits.md
```diff
@@ -28,7 +28,8 @@ The following aspects apply at the Temporal Cloud Account level (per account).
 
 **How many users can I add?**
 
-300 users across all Namespaces.
+By default, 300 users across all Namespaces.
+You can request this limit be rasied by opening a [support ticket](/cloud/support-create-ticket).
```
lorensr, line 32:
```suggestion
```

removing per sergey

PR #2712, opened 2024-03-26
Add detail to typescript docs

lauhon:
When manually skipping time, the test-worker still has to be activated via .run() for example. This is not represented in the current docs.


flossypurse:
Hi @lauhon, suggestion applied here: https://github.com/temporalio/documentation/pull/2741

We recently deprecated the pre-build system and the src files where you originally made the change - so we just picked it up applied it in the new spot.

Ty!

PR #2714, opened 2024-03-26
Fix Python Feature typos

axfelix:
Fixes some minor typos from the Python Features page


tomwheeler (approved):
I think it's spelled "Cancelable" in US English, but both spellings are widely used.

axfelix:
Fortunately in this case I'm just matching an existing function name; the call was incorrect

PR #2716, opened 2024-03-27
Adds guidance on BlobSizeLimitError

rachfop:

### DIFF
--- a/docs/troubleshooting/blob-size-limit-error.mdx
+++ b/docs/troubleshooting/blob-size-limit-error.mdx
@@ -34,7 +34,7 @@ There are multiple strategies you can use to avoid this error:
 
 2. Break larger batches of commands into smaller batch sizes:
    - Workflow-level batching:
-     1. Modify the Workflow to process Activities in smaller batches.
+     1. Modify the Workflow to process Activities or Child Workflows into smaller batches.
      2. Iterate through each batch, waiting for completion before moving to the next.
    - Workflow Task-level batching:
      1. Execute Activities in smaller batches within a single Workflow Task.


docs-src/cloud/limits.md
```diff
@@ -169,7 +169,42 @@ These operations create Events that are then logged in the Event History.
 The transaction size limit restricts the total size of Events that can be accommodated within a single transaction.
 
 The size limit for any given [Event History](/concepts/what-is-an-event-history) transaction is 4 MB.
-This isn't configurable.
+This limit is non-configurable for Temporal Cloud.
+
+**What is the blob size limit for a transaction?**
+
+- **Blob size limit** for Payloads, including Workflow context and each Workflow and Activity argument and return value:
```
mjameswh, line 176: I don’t think that documenting warning thresholds is pertinent for Cloud limit docs, as there is no way ATM for customers to see these warnings anyway.

docs-src/cloud/limits.md
```diff
@@ -169,7 +169,42 @@ These operations create Events that are then logged in the Event History.
...
+
+This limit is non-configurable for Temporal Cloud.
+
+The following provides solutions for handling large payloads.
+
+**Solutions for handling large payloads**
+
+If you're at risk of reaching the blob size limit, consider the following solutions:
+
+1. Use compression with a [custom Data Converter](/concepts/what-is-a-custom-data-converter) for large payloads.
```
mjameswh, line 188: I would hint more directly at a custom payload _codecs_, as that's how compression should be implemented.

docs-src/cloud/limits.md
```diff
@@ -169,7 +169,42 @@ These operations create Events that are then logged in the Event History.
...
+2. Break larger batches of commands into smaller batch sizes.
```
mjameswh, line 189: Might want to give explanation on how to "break batches of commands". That might not be clear for users. This may be done either at the workflow-level (i.e., have the workflow itself handle smaller batches), or at the workflow _task_ level (i.e., wait on a 1ms timer after each batch of X scheduled activities).

docs-src/cloud/limits.md
```diff
@@ -169,7 +169,42 @@ These operations create Events that are then logged in the Event History.
...
+  - Temporal warns at 256 KB: `Blob size exceeds limit`.
+  - Temporal errors at 2 MB: `ErrBlobSizeExceedsLimit: Blob data size exceeds limit.`
+
+This limit is non-configurable for Temporal Cloud.
+
+The following provides solutions for handling large payloads.
+
+**Solutions for handling large payloads**
+
+If you're at risk of reaching the blob size limit, consider the following solutions:
```
mjameswh, line 186: We should also mention the possibility of offloading payloads to an object store. Many users actually choose to go that way.

docs-src/cloud/limits.md
```diff
@@ -169,7 +169,42 @@ These operations create Events that are then logged in the Event History.
...
+
+- **Blob size limit** for Payloads, including Workflow context and each Workflow and Activity argument and return value:
+  - Temporal warns at 256 KB: `Blob size exceeds limit`.
+  - Temporal errors at 2 MB: `ErrBlobSizeExceedsLimit: Blob data size exceeds limit.`
+
+This limit is non-configurable for Temporal Cloud.
+
+The following provides solutions for handling large payloads.
+
+**Solutions for handling large payloads**
```
mjameswh, line 184: The solutions proposed here are not only applicable to cases where one would face blob-size errors, but also may be pertinent to cases of exceeding per-workflow limits.

For example:
- payload compression (and payload offloading, as suggested later) might help in cases of workflow size being too large; 
- smaller batches may help in cases of exceeding the number of concurrent ScheduleActivityTask/SignalExternalWorkflowExecution/StartChildWorkflowExecution/RequestCancelExternalWorkflowExecution.

These solutions are also pertinent in self-hosted scenarios.

mjameswh: For those reasons, you may want to consider wither these propositions would better fit as a distinct page, linked from here.

docs-src/cloud/limits.md
```diff
@@ -169,7 +169,42 @@ These operations create Events that are then logged in the Event History.
...
+
+**Monitoring and Debugging**
+
+To proactively identify and address issues related to blob size limits, consider the following:
+
+1. Monitor the `temporal_request_failure` SDK metric for operations such as:
+   - `RespondWorkflowTaskCompleted`
+   - `ActivityTaskCompleted`
+2. Set up alerts on the `temporal_request_failure` SDK metric for these operations to receive notifications when failures occur.
+3. Implement a gRPC proxy between the Client Worker and the service to obtain gRPC message size information. This allows you to check the message size and identify potential issues before they cause failures.
```
mjameswh, line 199: I think we should remove this last suggestion.

docs-src/cloud/limits.md
```diff
@@ -169,7 +169,42 @@ These operations create Events that are then logged in the Event History.
...
+
+**Versioning Workflows**
```
mjameswh, line 201: This subsection feels out of place on the present page. It's true that one might need to modify their workflow code to fix the above issue, and might need to use versioning to preserve compatibility, but there are tons of other cases where they might need to do so too, and we're not discussing Workflow Versioning every single time.

Issue #1838, opened 2022-12-09
[Documentation request] Recommend that workers `identity` should be set to something meaningful


mjameswh:
# Brief description

Workers have an "identity". That identity is visible to users in workflow history and in the list of pollers on a task queue, and may help users debug some operational issues... However, its usefulness is generally limited unless the user can link back that identity to the corresponding machine/process/execution context/log stream/...

For example, in most SDKs, a worker's identity defaults to `${process.pid}@${os.hostname()}`. This seems like a sensible default, but in practice, turns out to mean nothing useful in most cloud environments. For example, on ECS, the process id is always 1 (because of Docker), and the hostname is a completely random string. In other case, hostname will be an ephemeral IP address.

We should encourage users to explicitly set worker identity to something that can actually help them. For example, on ECS, a user might want to set its worker's identity to the ECS task id.

#  Your recommended content

I can provide content, but it would help to first know where this would go.

Some possibilities:
- https://docs.temporal.io/workers
- https://docs.temporal.io/application-development/debugging
- https://docs.temporal.io/application-development/foundations?lang=go#run-worker-processes
- https://docs.temporal.io/application-development/observability#logging
- A new page including 
- A knowledge base article (seems more appropriate if we want to include example code of how to obtain meaningful id for various cloud providers)

PR #2722, opened 2024-03-27
Adds Worker Id

rachfop:
## What does this PR do?

https://github.com/temporalio/documentation/issues/1838


docs-src/concepts/what-is-a-worker-identity.md
```diff
@@ -0,0 +1,53 @@
...
+id: what-is-a-worker-identity
+title: What is a Worker Identity?
+sidebar_label: Worker Identity
+description: A Worker Identity is an identifier for a Worker Client.
+tags:
+  - term
+  - explanation
+---
+
+Workers have an associated identifier that helps identify the specific Worker instance.
```
flossypurse, line 11: When we say "Worker instance", Is this at the Worker Process level or the Worker Entity level?

From below it seems like it is at the Process level - 
If so, it means one can have multiple Entities in the same process, but they would be identified as the same.

Can we clarify that?

flossypurse (commented):
I don't think we want to introduce a new term here.

Let's try add this to one of the existing Worker terms.

atm it sounds like it should go in the Worker Process term/concept page as it is reflective of that level. But just clarify that first.

docs-src/concepts/what-is-a-worker-identity.md
mjameswh: In Core-based SDKs, `identity` can be specified per Client or per Worker (I mean instances of the SDK's Worker class), while in Go and Java, it can only be specified on Client. To keep documentation simple, it's probably better not include per-worker suffix in snippets.

docs-src/concepts/what-is-a-worker-identity.md
```diff
@@ -0,0 +1,53 @@
...
+In most SDKs, the Worker Identity is referenced as `${process.pid}@${os.hostname()}`, which combines the Worker's process ID (`process.pid`) and the hostname of the machine running the Worker (`os.hostname()`).
```
mjameswh, line 12:
```suggestion
By default, Temporal SDKs set a Worker Identity to `${process.pid}@${os.hostname()}`, which combines the Worker's process ID (`process.pid`) and the hostname of the machine running the Worker (`os.hostname()`).
```

mjameswh: AFAIK, that default is applicable in all SDKs.

PR #2723, opened 2024-03-28
Describes how to find Namespace endpoints

fairlydurable:
The original ask was to find the port. This expands the brief to provide the endpoint, which includes the port. This patch:

* Adds a new concepts node (What is a Cloud gRPC endpoint)
* Links it into the Manage Namespaces guide config
* Adds a new image
* Updates the migration page
* Updates all five of the "connect to temporal cloud" foundation coverage points, being sure to point out that the endpoint contains the port to reduce that particular pain point.

It does not change code. If that is needed, they will need to be separate tickets.

https://temporalio.atlassian.net/browse/EDU-2148


axfelix:
LGTM!

axfelix (approved):

PR #2729, opened 2024-04-03
Update TS SDK doc to 1.9.x

mjameswh:
## What does this PR do?

- Replace various deprecated/non-preferred APIs and idioms by their 1.9.x alternatives, including:
  - `WorkflowClient` => `Client.workflow`;
  - Activity's `Context.current().info` => `activityInfo()` (same for `.sleep()`, `.heartbeat()` and others);
  - `WorkflowClientCallsInterceptor` => `WorkflowClientInterceptor`, and no more factory for these;
  - Document and recommend use of Workflow Context Logger and Activity Context Logger, instead of complicated alternatives;
  - Remove references to the `sourceMapPath` property when passing prebuilt Workflow bundle to a Worker.
  - Fixed all known invalid links to generated TS SDK API docs;
  - Updated Node.js version on TS samples to Node 20;
  - Similarly updated TypeScript to 5.x, and many more dependencies.

## Notes to reviewers

1. I tried as much as possible to keep this PR at a "mechanical" level, mostly find-and-replace-kind of things, with rewrites only in places where that was needed. Splitting this into smaller change sets would not make sense IMHO. More "editorial" changes will be covered in subsequent PRs.

2. I also updated dependencies of the Doc repo itself, because that seems the proper thing to do, and because having different versions of Node and TS for the repo vs the samples would make testing samples awkward.

3. This PR is touching several content nodes that are actually not used anymore. I only figured that out after spending time fixing them. Not sure what the policy is regarding unreferenced files.


docs-src/typescript/activities.md
```diff
@@ -385,8 +393,8 @@ Not suitable for Heartbeating:
 // activity implementation
 export async function example(sleepIntervalMs = 1000): Promise<void> {
   for (let progress = 1; progress <= 1000; ++progress) {
-    await Context.current().sleep(sleepIntervalMs);
-    Context.current().heartbeat();
+    await sleep(sleepIntervalMs);
+    heartbeat();
```
bergundy, line 397: I'd add `import * as activity from '@temporalio/activity'` and change the sample to this:

```suggestion
    await activity.sleep(sleepIntervalMs);
    activity.heartbeat();
```

It'll help users understand where these function are coming from.

docs-src/typescript/activities.md
```diff
@@ -424,28 +432,33 @@ Activity Cancellation is an optional capability that lets you do graceful cleanu
...
 
 ```ts
-import { CancelledFailure, Context } from '@temporalio/activity';
+import {
+  activityInfo,
+  CancelledFailure,
+  heartbeat,
+  log,
+  sleep,
+} from '@temporalio/activity';
```
bergundy, line 456: I would rather use `import * as activity` and not import the individual methods. it's mostly a stylistic preference but I think it ends up more readable if methods are prefixed with `activity`.

docs-src/typescript/how-to-list-search-attributes-in-typescript.md
```diff
@@ -66,7 +66,7 @@ const decodedWorkflows = executions.map(
       memo: memo && Object.keys(memo!).length === 0 ? null : memo, // convert empty object to null
       searchAttributes,
       stateTransitionCount: stateTransitionCount!.toInt(),
-    } as unknown as Workflow;
+    } as unknown as WorkflowInfo;
```
bergundy, line 69: This is not something users should do.

And this whole sample showing listing workflows with the `Connection` is outdated.

bergundy (approved):
I skimmed this. Thanks for updating these docs. It's been long overdue.

flossypurse:
@mjameswh - We just deprecated Assembly. I merged main in - but then realized that you may want to address Roey's comments in code source and use Assembly to propagate first before merging in main, so I reverted to you last commit.

LMK when you have done that, and I can help you merge the Assembly deprecation changes in.

PR #2730, opened 2024-04-03
Remove unneeded nodes in the docs

rachfop:
## Notes to reviewers

Seeing if this build passes on vercel.


flossypurse:
We removed Assembly including all `docs-src` nodes.
Let's keep this PR alive for the moment to easily see what files we needed to audit.

PR #2768, opened 2024-04-22
add admonition for db compatibility

axfelix:


axfelix:
Addresses https://temporalio.atlassian.net/browse/EDU-2224

docs/cli/index.mdx
```diff
@@ -277,6 +277,13 @@ This is helpful if you plan on stopping and re-starting the dev server.
 temporal server start-dev --db-filename temporal.db
 ```
 
+:::note
+
+The SQLite database is not guaranteed to remain compatible across Temporal CLI updates.
+The `temporal server` command is only intended for development environments.
```
flossypurse, line 283: I recommend being a bit more specific here about the database file and maybe not even mention SQLite.

> If you start the development service with a database file, it may not be compatible with newer versions of the Temporal CLI. 

Something that says it has everything to do with the schema of the DB file and the version of the CLI.

Right now it sounds way too much like general use of SQLite databases with the Server and the CLI.

If the message is focused on the compatibility of the file - then you can drop this sentence altogether.
> The `temporal server` command is only intended for development environments.

axfelix: Went with:

Local databases created with `--db-filename` may not be compatible with newer versions of the Temporal CLI.
The `temporal server` command is only intended for development environments.

flossypurse (approved):

PR #2782, opened 2024-04-26
Update the main CLI page with changes for 0.12

josh-berry:
I manually tested all the steps for installing auto-completion for each shell (bash, fish, zsh).

I have removed the "Install via cURL" options because it's generally bad form, security-wise, to do anything that looks like `curl | sh`.  (Yes, many people do it, however we don't want to encourage it.)

Note: I have not updated all the places with syntax changes in the documentation generally; we'll need to do this in the future, but for now, it's okay because the older syntax will continue to work.


josh-berry:
This PR is ready for review, but is not ready for merge until 0.12 is released.

flossypurse:
@josh-berry 

> I have removed the "Install via cURL" options because it's generally bad form, security-wise, to do anything that looks like `curl | sh`. (Yes, many people do it, however we don't want to encourage it.)

Is it OK for Windows? We have this directly on the main website about 2/3rds down the page: https://temporal.io/product

Screenshot 2024-04-26 at 8 18 40 AM

josh-berry:
> @josh-berry
> 
> > I have removed the "Install via cURL" options because it's generally bad form, security-wise, to do anything that looks like `curl | sh`. (Yes, many people do it, however we don't want to encourage it.)
> 
> Is it OK for Windows? We have this directly on the main website about 2/3rds down the page: https://temporal.io/product

I wouldn't recommend it anyhere, tbh. But I'm noticing that both the Linux and Windows commands seem to be wrong anyway; we're not publishing packages to Ubuntu/Debian today, and `curl |sh` is a UNIX-ism. So we should probably sort that out separately.

cretz (approved):

docs/cli/index.mdx
```diff
@@ -60,7 +60,7 @@ temporal server start-dev --namespace foo --namespace bar
 Register Namespaces with `namespace create`:
 
 ```shell
-temporal operator namespace create foo
+temporal operator namespace create -n foo
```
fairlydurable, line 63:
```suggestion
temporal operator namespace create --namespace foo

temporal operator namespace create -n foo
```

As a tech writing rule, prefer the full "readable" version to the shortcut. Clarity trumps concision. 

Several ways to approach this, use the full version, use both with the preferred first, or stick with organizational standard (-n) and ignore suggestion. This is a due diligence suggestion. 

The extra characters take up more space, especially in help messages, but I was scratching my head a lot when I was first exposed to `-n` in messages and had no context to explain it. As a new adopter being shown --namespace first and then knowing as a Unix user there might be a shorter flag, would have made some work (coverage of creating a namespace using the Temporal CLI) easier for me. I expect other new adopters, a big target I've been told for approachable tooling and documentation, would also benefit.

docs/cli/index.mdx
```diff
@@ -95,7 +95,6 @@ The following table lists and describes the environment variables you can set fo
 | `TEMPORAL_ADDRESS`                       | Host and port (formatted as host:port) for the Temporal Frontend Service. | --address                       |
 | `TEMPORAL_CODEC_AUTH`                    | Sets the authorization header for requests to Codec Server.               | --codec-auth                    |
```
fairlydurable, line 96: All the others start with the important noun, not a a description of use. To be consistent:
```suggestion
| `TEMPORAL_CODEC_AUTH`                    | Authorization header for requests to Codec Server.               | --codec-auth                    |
```

docs/cli/index.mdx
```diff
@@ -138,39 +137,49 @@ Enable auto-completion using the following commands.
 
 ### zsh auto-completion
 
-<!-- TODO: add more information about zsh to make comparable to bash section -->
+1. Add the following line to your `~/.zshrc` file:
```
fairlydurable, line 140: The pedant in me:

```suggestion
1. Add the following line to your `~/.zshrc` configuration file:
```

docs/cli/index.mdx
```diff
@@ -138,39 +137,49 @@ Enable auto-completion using the following commands.
...
-alias t='temporal'
-alias tw='temporal workflow'
-alias ts='temporal server start-dev'
-alias tsdb='temporal server start-dev --db-filename ~/temporal.db'
+### Fish auto-completion
 
-# send process to background so you can continue using the terminal
-alias tsbg='temporal server start-dev &> /dev/null & disown'
-```
+1. Make sure there is a place to store custom completions for Fish:
```
fairlydurable, line 170: A little more clarity that is action-oriented rather than "checking"

```suggestion
1. Create the Fish custom completions if one does not already exist:
```

docs/cli/index.mdx
```diff
@@ -138,39 +137,49 @@ Enable auto-completion using the following commands.
...
+
+   ```fish
+   mkdir -p ~/.config/fish/completions
+   ```
+
+2. Arrange for the completions to be loaded when needed (note: the file name must be `temporal.fish` or the completions will not be found):
```
fairlydurable, line 176:
```suggestion
2. Configure the completions to load when needed. Note: the file name must be `temporal.fish` or the completions will not be found:
```

docs/cli/index.mdx
fairlydurable: Also, leaving this here because I don't know how to add it to the points where it applies (starting on line 25). The intro can be simplified for easier reading.

Starting at Line 25 -> 


The Temporal CLI, formerly 'tctl', is a command-line interface that offers direct access to a Temporal Cluster (Service).
It supports management, monitoring, and debugging of Temporal Applications, and administrative functions like Namespace management. 
With it, you can initiate, signal, or cancel Workflows and query Workflow states from your terminal.
The Temporal CLI is an indispensable component of your Temporal developer toolkit.

When upgrading from `tctl` to Temporal CLI, update your existing environment variables and be aware that some commands use new syntax.

docs/cli/index.mdx
```diff
@@ -138,39 +137,49 @@ Enable auto-completion using the following commands.
...
 
-After setting the variable, run the following command:
+1. Install [bash-completion](https://github.com/scop/bash-completion#installation) and add the software to your `~/.bashrc`.
 
-`source ~/.zshrc`.
+2. Add the following line to your `~/.bashrc` file:
 
-### Bash auto-completion
+   ```sh
+   eval "$(temporal completion bash)"
```
fairlydurable, line 159: Comment: tested this with the default Apple Bash install, and found it errored during completion. I leave my system as "vanilla" as possible, and as close to the default installs that people will encounter when first using their computer (in this case macOS Sonoma, Bash).

```
bash-3.2$ temporal serbash: _get_comp_words_by_ref: command not found
ver
```

docs/cli/index.mdx
josh-berry: This is may be a little pedantic in turn, but it's actually a script, not a configuration file—it can run arbitrary code when `zsh` is launched. The `zsh` manpage calls it a "shell startup file", the `bash` manpage calls it an "initalization file`; I think I'll go with "startup script" if that works for you?

docs/cli/index.mdx
josh-berry: Do you have `bash-completion` installed? This error usually means the completion library isn't loaded in your shell. I can add a note on that in the docs.

docs/cli/index.mdx
fairlydurable: If memory serves, rc means "run commands" so have at it. Generally, I have described it as a configuration file that customizes the launch of a shell. Adding some kind of modifier explains what you're changing to any people who don't know _why_ they're editing the file. I like shell initialization slightly more than just initialization.

docs/cli/index.mdx
josh-berry: 👍 I tested it too! And apparently had bash-completion installed even though I use zsh, heh. I blame Homebrew. :)

tlalfano:
Looks good, for this sentence:

tlalfano:
Looks good, for this sentence: 

>  embedded Temporal Cluster suitable for use in development and CI/CD
>

docs/cli/index.mdx
```diff
@@ -20,13 +20,12 @@ tags:
...
-The Temporal CLI, evolving from its precursor `tctl`, provides developers with direct CLI access to a Temporal Service.
+The Temporal CLI provides direct access to a Temporal Service via the terminal.
 It's a powerful tool for managing, monitoring, and debugging Temporal Applications.
-With the Temporal CLI, developers can start Workflows, Signal or cancel them, and Query Workflow state, directly from their terminal.
-This interface also facilitates administrative tasks such as Namespace management.
+You can use it to start, stop, inspect and operate on Workflows and Activities, and perform administrative tasks such as Namespace, Schedule, and Task Queue management.
 
-The Temporal CLI encapsulates complex operations into simpler command-line inputs, making it an indispensable component of a developer’s toolkit when working with the Temporal ecosystem.
-Its design philosophy emphasizes ease of use and efficiency, catering to both quick inspections and more complex batch operations, thereby streamlining the development and operational processes for durable applications.
+The Temporal CLI also includes an embedded Temporal Cluster suitable for use in development and CI/CD.
```
flossypurse, line 27:
```suggestion
The Temporal CLI also includes an embedded Temporal Service suitable for use in development and CI/CD.
```

We are in the process of updating "Temporal Cluster" to "Temporal Service".

flossypurse (approved):

PR #2809, opened 2024-05-07
DO NOT MERGE: EDU-2385: Add Evaluate > Production Deployment Features > Cloud > Multi-region Namespaces

fairlydurable:
Builds up Evaluate > Cloud > Multi-region Namespaces

Status: minimal skeleton


fairlydurable:
Leaving for archival reference but will delete after I capture the tagging and other frontmatter items

PR #2819, opened 2024-05-09
Typescript Client Page 

jsundai:


docs/develop/tscript/temporal-client.mdx
flossypurse, line 1: Let's remove "typescript" from here https://github.com/temporalio/documentation/blob/main/docusaurus.config.js#L215 and use "typescript" directory name.

PR #2820, opened 2024-05-09
EDU-2427: `tcld` coverage of MRN

fairlydurable:
Adds the specific command-line details for using `tcld` to provision multi-region namespaces, plus warnings about de-provisioning.

### DIFF
--- a/docs/production-deployment/cloud/tcld/namespace.mdx
+++ b/docs/production-deployment/cloud/tcld/namespace.mdx
@@ -42,7 +42,7 @@ Alias: _none_
 
 :::caution
 
-To end multi-region service and end charges, contact [Temporal Support](https://support.temporal.io) directly.
+To end multi-region service and discontinue replication charges, contact [Temporal Support](https://support.temporal.io) directly.
 You can't use `tcld` to remove a region from a namespace.
 
 :::
--- a/docs/production-deployment/cloud/tcld/namespace.mdx
+++ b/docs/production-deployment/cloud/tcld/namespace.mdx
@@ -17,6 +17,7 @@ The `tcld namespace` commands enable [Namespace](/namespaces) operations in Temp
 
 Alias: `n`
 
+- [tcld namespace add-region](#add-region)
 - [tcld namespace create](#create)
 - [tcld namespace delete](#delete)
 - [tcld namespace get](#get)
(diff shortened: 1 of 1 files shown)


docs/production-deployment/cloud/tcld/namespace.mdx
```diff
@@ -28,6 +30,57 @@ Alias: `n`
 - [tcld namespace retention](#retention)
 - [tcld namespace update-codec-server](#update-codec-server)
 
+## add-region
+
+Use `tcld namespace add-region` to add a standby region to an existing Temporal Cloud [Namespace](/namespaces).
```
yux0, line 35: Should we link this to a Multi Region Namespaces section?

docs/production-deployment/cloud/tcld/namespace.mdx
```diff
@@ -28,6 +30,57 @@ Alias: `n`
...
+Enrolling a second region upgrades the Namespace to multi-region. <!--[multi-region](/cloud/multi-region)-->
+Once provisioned, the multi-region Namespace enables Temporal Cloud to start replicating Workflow Execution data from the active to the standby region and trigger failover during adverse conditions.
+
+`tcld namespace add-region`
+
+Alias: _none_
+
+:::caution
+
+To end multi-region service and end charges, contact [Temporal Support](https://support.temporal.io) directly.
```
yux0, line 45: I think we should mention the charges in add-region, if we specify the end charges in remove region.

docs/production-deployment/cloud/tcld/namespace.mdx
```diff
@@ -28,6 +30,57 @@ Alias: `n`
...
+_Required modifier_
+
+#### --region
+
+The region to add to the existing Namespace.
+
+Valid options: `ap-northeast-1` | `ap-northeast-2` | `ap-south-1` | `ap-southeast-1` | `ap-southeast-2` | `eu-central-1` | `eu-west-1` | `eu-west-2` | `ca-central-1` | `us-east-1` | `us-east-2` | `us-west-2`
+
+See [Service Availability](/cloud/service-availability). 
+The `sa-east-1` region is not supported at this time.
```
yux0, line 78: @atihkin Is this still true?

yux0 (commented):
Should we make a specific that the MRN is not supporting API Keys at this point?

docs/production-deployment/cloud/tcld/namespace.mdx
atihkin: Yeah because there's only one region available in South America. So can't pair it with any other region in the same continent.

docs/production-deployment/cloud/tcld/namespace.mdx
fairlydurable: We can't do that until MRN is pushed first because Vercel will reject the draft

docs/production-deployment/cloud/tcld/namespace.mdx
fairlydurable: Will update to something like. Charges for data replication will apply as soon as the new multi-region service is provisioned.

docs/production-deployment/cloud/tcld/namespace.mdx
fairlydurable: Thank you both

fairlydurable:
> Should we make it specific that the MRN is not supporting API Keys at this point?

I'd like to limit the scope here. Please file a ticket now at https://t.mp/edu-request and request that for when it becomes relevant. Thanks

docs/production-deployment/cloud/tcld/namespace.mdx
fairlydurable: Looking at this again, this is located in the add-region section of the coverage. I don't believe any change is needed.

PR #2954, opened 2024-07-09
EDU-2746: Adds post-failover state information

fairlydurable:
Informs customers that the regions stay switched and explains how to mitigate.


axfelix (approved):

axfelix:
LGTM

PR #2959, opened 2024-07-14
Update overview docs

drewhoskins-temporal:
## What does this PR do?
Revamp message passing overview docs for the Public Preview launch of Workflow Update with the following goals.
* Teach users about workflow concurrency and how to make safe hanadlers.
* Correct misstatements about Update from the pre-release.
* Help users choose whether to use signals or updates.

## Test plan
http://localhost:3000/encyclopedia/workflow-message-passing
Click all the links.  

## Notes to reviewers

* This was reviewed by Engineers and @axfelix in https://docs.google.com/document/d/1XsCgi7ibqxanDEfh2nS19b_4KBdZajiqtpmyy_cpErc/edit
* A few of the links are waiting for the language-specific docs to create the #subsections, to be checked in in separate PRs before Update launches in Public preview.  They currently nav to the top of the page, which I hope is acceptable until we make those commits.
* According to @napcs, the image needs to be redrawn in the house style.


docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
 
-Signal-With-Start is a Client method that takes the following arguments:
+### Sending Signals
 
-- A Workflow Id
-- Workflow input
-- A [Signal](#signals) name
-- Signal input
+You can send Signals from any Temporal Client, the Temporal CLI, or you can Signal one Workflow to another.
+You can also "Signal-With-Start" and send a Signal from
```
angelazhou32, line 82: Incomplete sentence

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
 
-A Signal must include a destination (Namespace and Workflow Id) and name.
-It can include a list of arguments.
+- Queries are efficient–they never add entries to the [Workflow Event History](/workflows#event-history).
+- Queries can operate on completed Workflows.
 
-Signal handlers are Workflow functions that listen for Signals by the Signal name.
-Signals are delivered in the order they are received by the Temporal Service and written to History.
-If multiple deliveries of a Signal would be a problem for your Workflow, add idempotency logic to your Signal handler that checks for duplicates.
+However, because Queries cannot block, sometimes Updates are best.\
```
angelazhou32, line 69:
```suggestion
However, because Queries cannot block, sometimes Updates are best.
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
-    <RelatedReadItem path="/develop/dotnet/message-passing#define-a-signal" text="Send and handle Signals with the .NET SDK" archetype="feature-guide" />
-</RelatedReadContainer>
+From the above descriptions, it may sound like Updates can do all the things that Queries and Signals can do. But, unlike Signals, Updates must be synchronous, that is, they must wait for the worker running the Workflow to acknowledge the request.
+
+Use Signals instead of Updates when:
+
+- You want to send a fully asynchronous message.
+- Your client does not want to rely on the worker being available.
+- Your Workflow’s client cares more about the latency of sending the message than it does of the end-to-end operation.
+- When you need to [Signal-With-Start](signal-with-start) atomically start the Workflow and then Signal as the first operation. (since there is currently no Update-with-start operation.)
```
angelazhou32, line 53: Doesn't seem to be a link with signal-with-start

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
+### For read requests
 
-A Signal delivers data to a running Workflow Execution.
-It cannot return data to the caller; to do so, use a [Query](#queries) instead.
-The Workflow code that handles a Signal can mutate Workflow state.
-A Signal can be sent from a Temporal Client or a Workflow.
-When a Signal is sent, it is received by the Temporal Service and recorded as an Event to the Workflow Execution [Event History](/workflows#event-history).
-A successful response from the Temporal Service means that the Signal has been persisted and will be delivered at least once to the Workflow Execution.[^1]
-The next scheduled Workflow Task will contain the Signal Event.
+You normally want to do a Query:
```
angelazhou32, line 64:
```suggestions
Use Queries when:
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
-A Signal must include a destination (Namespace and Workflow Id) and name.
-It can include a list of arguments.
+- Queries are efficient–they never add entries to the [Workflow Event History](/workflows#event-history).
+- Queries can operate on completed Workflows.
 
-Signal handlers are Workflow functions that listen for Signals by the Signal name.
-Signals are delivered in the order they are received by the Temporal Service and written to History.
-If multiple deliveries of a Signal would be a problem for your Workflow, add idempotency logic to your Signal handler that checks for duplicates.
+However, because Queries cannot block, sometimes Updates are best.\
+When your goal is to do a read once the Workflow achieves a certain desired state, you have two options.
```
angelazhou32, line 70:
```suggestion
When your goal is to do a read once the Workflow achieves a certain desired state, you have two options:
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
 
-[^1]: The Temporal Service usually deduplicates Signals, but does not guarantee deduplication: During shard migration, two Signal Events (and therefore two deliveries to the Workflow Execution) can be recorded for a single Signal because the deduping info is stored only in memory.
+- You could poll periodically with Queries until the workflow is ready.
```
angelazhou32, line 72: Noticed a few instances workflow/worker are not capitalized - can we capitalize them please when talking about a Temporal Workflow/Worker? Thanks!

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
 
-Temporal applications can receive and provide data through messages.
+Workflows can be thought of as stateful web services that can receive messages. The Workflow can have powerful message handlers akin to endpoints that can react to the incoming messages in combination with the current state of the Workflow.
 Temporal supports three types of messages: Signals, Queries, and Updates.
+To define these, consider these messages from the perspective of the client making the request:
 
-## Signals {#signals}
+- Queries are read requests. They can read the current state of the Workflow but cannot block in doing so.
+- Signals are write requests. They cause changes in the running Workflow, but you cannot await any response or error.
+- Updates are interactive write requests, wherein the sender of the Update tracks completion and waits for a response or an error.
```
dandavison, line 40: I don't think "interactive" is quite the right word here; how about

> Updates are trackable write requests

which I think you had in an earlier version.


"Interactive" suggests a bit too much, i.e. that the client can modify the execution of the request after first issuing it. An "interactive request" is not currently a standard concept in software engineering and in our case, we don't even support cancellation of update requests yet.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
-    <RelatedReadItem path="/develop/go/message-passing#signals" text="Send and handle Signals with the Go SDK" archetype="feature-guide" />
-    <RelatedReadItem path="/develop/java/message-passing" text="Send and handle Signals with the Java SDK" archetype="feature-guide" />
-    <RelatedReadItem path="/develop/php/message-passing#signals" text="Send and handle Signals with the PHP SDK" archetype="feature-guide" />
-    <RelatedReadItem path="/develop/python/message-passing#signals" text="Send and handle Signals with the Python SDK" archetype="feature-guide" />
-    <RelatedReadItem path="/develop/typescript/message-passing#signals" text="Send and handle Signals with the TypeScript SDK" archetype="feature-guide" />
-    <RelatedReadItem path="/develop/dotnet/message-passing#define-a-signal" text="Send and handle Signals with the .NET SDK" archetype="feature-guide" />
-</RelatedReadContainer>
+From the above descriptions, it may sound like Updates can do all the things that Queries and Signals can do. But, unlike Signals, Updates must be synchronous, that is, they must wait for the worker running the Workflow to acknowledge the request.
```
dandavison, line 46:
```suggestion
From the above descriptions, it may sound like Updates can do all the things that Queries and Signals can do. But, unlike Signals, Updates must wait for the worker running the Workflow to acknowledge the request.
```

I'm not sure we should define "synchronous" here, since that word (and "asynchronous") have multiple similar meanings that are relevant to Temporal. After all, Update can also be viewed as "asynchronous" in the sense that an acknowledgement response can be received before the result is available. Indeed, there is some precedent for referring to that latter style as "async update" (although I propose that we also do not allow that terminology to get into our public-facing docs).

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
+
+Use Signals instead of Updates when:
+
+- You want to send a fully asynchronous message.
```
dandavison, line 50:
```suggestion
- You want to send a message to a workflow and wait only for the Server to receive it; not the Worker.
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
+- Your client does not want to rely on the worker being available.
```
dandavison, line 51: I think these two bullet points might be making just one point, or at least the distinction is very subtle. Also I don't think "fully asynchronous" is clear. Perhaps:


```suggestion
- You want to send a message and wait only for the Server to receive it, without relying on a Worker being available.
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
+- Your Workflow’s client cares more about the latency of sending the message than it does of the end-to-end operation.
```
dandavison, line 52: There's a typo (`s/does of/does about/`) but, more importantly, I'm finding that I have to re-read this one multiple times. At least part of my brain is saying: "But a client who chooses signal over update _does_ care about the end-to-end latency: it thinks that it is too slow to be able to wait for it". I think perhaps something more like the following is clearer:

```suggestion
- Your Workflow’s client wants the overall latency to be low, and is prepared to give up the ability to track the operation in return for low latency.
```

docs/encyclopedia/application-message-passing.mdx
dandavison: I think maybe it's missing "to" as in "When you need to Signal-With-Start atomically to start the Workflow...", but how about we just leave it at

```suggestion
- When you need to [Signal-With-Start](signal-with-start) atomically. (since there is currently no Update-with-start operation.)
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
+- When you need to [Signal-With-Start](signal-with-start) atomically start the Workflow and then Signal as the first operation. (since there is currently no Update-with-start operation.)
+
+Use Updates instead of Signals when:
+
+- You want to track the completion of the message from the client.
+- You need a result or an exception from your message without having to query subsequently.
+- You’d like to “validate” the Update before accepting it into the Workflow and its history.
+- Your client cares more about the latency of the end-to-end operation than you care about the latency of sending the message.
```
dandavison, line 60: As mentioned above, I'm not managing to understand these "cares mores about" bulletpoints. I've read several times and it's not becoming perfectly clear, so I think we need to drop them. (Also, who's "you"?)

```suggestion
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
+Use Signals instead of Updates when:
+
+- You want to send a fully asynchronous message.
+- Your client does not want to rely on the worker being available.
+- Your Workflow’s client cares more about the latency of sending the message than it does of the end-to-end operation.
+- When you need to [Signal-With-Start](signal-with-start) atomically start the Workflow and then Signal as the first operation. (since there is currently no Update-with-start operation.)
+
+Use Updates instead of Signals when:
+
+- You want to track the completion of the message from the client.
```
dandavison, line 57: We're using "you" and "the client" and "your client" and "the Workflow's client" -- let's tighten them up to make sure that we're using the same terminology for the same thing and not referring to the same thing multiple times. I think in this case "you" and "the client" are the same person, right?

Also, how about we add a bullet like

```markdown
- It's acceptable for the Workflow's client to have to wait for a Worker to respond before they get a response
```

docs/encyclopedia/application-message-passing.mdx
dandavison: I think how @drewhoskins-temporal has it is good here. The language in the following bullet points doesn't work if the lead-in is "Use Queries when".

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
-The Workflow code that handles a Signal can mutate Workflow state.
-A Signal can be sent from a Temporal Client or a Workflow.
-When a Signal is sent, it is received by the Temporal Service and recorded as an Event to the Workflow Execution [Event History](/workflows#event-history).
-A successful response from the Temporal Service means that the Signal has been persisted and will be delivered at least once to the Workflow Execution.[^1]
-The next scheduled Workflow Task will contain the Signal Event.
+You normally want to do a Query:
 
-A Signal must include a destination (Namespace and Workflow Id) and name.
-It can include a list of arguments.
+- Queries are efficient–they never add entries to the [Workflow Event History](/workflows#event-history).
```
dandavison, line 66: We could also make this even more explicit / straightforward to understand, e.g.

```
- An update (if accepted) would add an entry to [Workflow Event History](/workflows#event-history), which would be pointless if it's a pure read operation.
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
 Temporal supports three types of messages: Signals, Queries, and Updates.
+To define these, consider these messages from the perspective of the client making the request:
 
-## Signals {#signals}
+- Queries are read requests. They can read the current state of the Workflow but cannot block in doing so.
+- Signals are write requests. They cause changes in the running Workflow, but you cannot await any response or error.
+- Updates are interactive write requests, wherein the sender of the Update tracks completion and waits for a response or an error.
 
-**What is a Signal in the context of Temporal?**
+## How to choose between Signals, Updates, and Queries? {#choosing-messages}
```
dandavison, line 42: I now see you have a "Sending Messages" section below. WDYT about changing this heading to:

```suggestion
## How to choose between Signals, Updates, and Queries as a Workflow author? {#choosing-messages}
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
+    <RelatedReadItem path="/develop/java/message-passing#send-signal-from-workflow" text="Send Signals from Workflows with the Java SDK" archetype="feature-guide" />
+    <RelatedReadItem path="/develop/php/message-passing#send-signal-from-workflow" text="Send Signals from Workflows with the PHP SDK" archetype="feature-guide" />
+    <RelatedReadItem path="/develop/python/message-passing#send-signal-from-workflow" text="Send Signals from Workflows with the Python SDK" archetype="feature-guide" />
+    <RelatedReadItem path="/develop/typescript/message-passing#send-signal-from-workflow" text="Send Signals from Workflows with the TypeScript SDK" archetype="feature-guide" />
+    <RelatedReadItem path="/develop/dotnet/message-passing#send-signal-from-workflow" text="Send Signals from Workflows with the .NET SDK" archetype="feature-guide" />
+</RelatedReadContainer>
+
+#### Signal-With-Start
+
+Signal-With-Start is a great tool for lazily creating workflows. When you send this operation, if there is a running Workflow Execution with the given Workflow Id, it will be Signaled. Otherwise, a new Workflow Execution is started and immediately sent the Signal.
```
dandavison, line 109: To my mind, "lazily" isn't quite accurate here. In "lazy evaluation", there is no doubt that the evaluation will occur; it's just that it's deferred until needed. I'd prefer to say that Signal-With-Start is "get or create" semantics. So I'd change this to

```suggestion
When you send this operation, if there is a running Workflow Execution with the given Workflow Id, it will be Signaled. Otherwise, a new Workflow Execution is started and immediately sent the Signal.
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,85 @@ keywords:
...
 
-[^1]: The Temporal Service usually deduplicates Signals, but does not guarantee deduplication: During shard migration, two Signal Events (and therefore two deliveries to the Workflow Execution) can be recorded for a single Signal because the deduping info is stored only in memory.
+- You could poll periodically with Queries until the workflow is ready.
+- You could write your read operation as an Update, which will give you better efficiency and latency, though it will write an entry to the [Workflow Event History](/workflows#event-history).
 
-### Signal-With-Start {#signal-with-start}
+## Sending Messages {#sending-messages}
 
-**What is a Signal-With-Start in the context of Temporal?**
+This content will help you write clients that send messages to Workflows.
```
dandavison, line 77:
```suggestion
This section will help you write clients that send messages to Workflows.
```

docs/encyclopedia/application-message-passing.mdx
dandavison: There are more than two options: another one that we currently document is that the client-side application could run an activity worker to receive responses from signals: https://github.com/temporalio/samples-go/tree/main/reqrespactivity

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -139,76 +197,161 @@ Stack Trace Queries are available only for running Workflow Executions.
...
-   In this case, the system rejects the request, doesn't record anything in the Workflow Event History to indicate that the Update ever happened, and the Update processing doesn't proceed to later phases.
-   If the Update completes the validation stage without error, the Platform changes its state to **Accepted** and a [WorkflowExecutionUpdateAcceptedEvent](/references/events#workflowexecutionupdateacceptedevent) Event is added to Workflow Execution [Event History](/workflows#event-history).
-3. **Execution.** Accepted Update requests move to the execution phase.
-   In this phase, the Worker delivers the request to the Update handler.
-4. **Completion.** The Update handler can return a result or a language-appropriate error/exception to indicate its completion.
-   The Platform sends the Update outcome back to the original invoking entity as an Update response.
-   A [WorkflowExecutionUpdateCompletedEvent](/references/events#workflowexecutionupdatecompletedevent) Event in the Workflow Execution Event History denotes the completion of an Update.
+- Throw any exception from your validator, OR
```
dandavison, line 306:
```suggestion
- Reject the update, by throwing any exception from your validator, OR
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -139,76 +197,161 @@ Stack Trace Queries are available only for running Workflow Executions.
...
+In Go, most exceptions behave like [Application Failures](/references/failures#application-failure) in the other SDKs. Panics behave like other exceptions and fail the Signal or Update handler tasks.
 
-For example, to enable Workflow Updates with the Temporal CLI, pass the value when executing the Temporal CLI command:
+#### Idempotency
 
-```command
-temporal server start-dev --dynamic-config-value frontend.enableUpdateWorkflowExecution=true
-```
+Most developers want their message handlers to run exactly once in cases where a Signal is delivered twice or sent by two different call sites. You can pass a unique key among your message arguments and use that to check for duplicate messages in your handler.
+For Updates, clients can instead pass the ID argument into the Update to serve as an idempotency key and as logging context. This can be retrieved within your handler and used to check whether the Update has already been handled.
```
dandavison, line 318: I think that some of this idempotency advice is superfluous.
- Updates are deduplicated by the server on updateID, but this section makes that fact unclear.
- For Signals there is a "request ID" deduplication mechanism that I think covers some subset of the "delivered twice" error mode
- I think it would help to clarify what it means for a signal to be delivered redundantly by two different call sites. I suspect that this concept is confusing to people, since of course Temporal absolutely allows multiple invocations of the same signal with the same arguments around the same time to be considered non-redundant.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -139,76 +197,161 @@ Stack Trace Queries are available only for running Workflow Executions.
...
-- Available in [PHP SDK](https://php.temporal.io/classes/Temporal-Client-WorkflowStubInterface.html#method_startUpdate) since [v2.8.0](https://github.com/temporalio/sdk-php/releases/tag/v2.8.0)
+If your Workflow receives messages, you may need to consider how those messages interact with one another or with the main Workflow method.
+Behind the scenes, Temporal is running a loop that looks like this:
+
+<PrettyImage src="/img/messages-workflow-loop.png" title="Diagram that shows the execution ordering of workflows" />
+
+Every time the Workflow wakes up--generally, it wakes up when it needs to--it will process messages in the order they were received, followed by making progress in the Workflow’s main method.
+
+This execution is on a single thread–while this means you don’t have to worry about parallelism, you do need to worry about concurrency if you have written Signal and Update handlers that can block. These can run interleaved with the main Workflow and with one another, resulting in potential race conditions. These methods should be made reentrant.
```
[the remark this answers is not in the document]
dandavison: [The event-loop diagram needs some updates along the lines of the changes I've proposed above.]

dandavison (commented):
This document is awesome. I'd like to collaborate on it and have made several suggestions. I can make a separate commit with my suggestions.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: I changed to "synchronous trackable" which is what I was going for with "interactive".

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: I think we need a short term for this, and  synchronous really is a term of art for this in software engineering ("Interactive" was my half-attempt to be more nuanced, but maybe it doesn't land) especially in the sense that the remote worker must be present.  And when we add durable-on-admitted, we can call it "async updates" as shorthand.  Going to keep it in the absence of an alternative term.
My theory is that the waiting-for-just-validation is niche enough that it's okay to gloss over that slightly at this level.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: The first bullet is a latency concern and the second is an availability concern.  I'll clarify.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: Your rephrasing isn't quite what I was emphasizing, as I addressed trackability in other points.  However, I made some minor readability tweaks.

docs/encyclopedia/application-message-passing.mdx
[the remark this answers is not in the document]
drewhoskins-temporal: FWIW, I'm really not a fan of the style of "Workflow Execution" and "send a Signal" when it's already clear from context.  It feels way too dry and makes me want to stop reading the docs.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: worked that in.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: Made a clarification

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: Interesting, I'm going to pass on this suggestion as I've not heard of anybody doing this and it feels a bit off topic, but feel free to add it as part of your edits if you think I'm off here.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: Per Claude: "In programming, the "get-or-create" pattern is commonly referred to as the "lazy initialization" pattern."
I'll use the more precise term.

docs/encyclopedia/application-message-passing.mdx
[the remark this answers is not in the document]
drewhoskins-temporal: I don't prefer all of these suggested changes, and I'm not sure what you're really going for with them.  I know you're trying to save me cycles, but it would be more helpful to get your experience reading than prescriptive non-trivial edits like this.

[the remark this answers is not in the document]
drewhoskins-temporal: (In particular, I fear you may be trying to do too much/be too comprehensive in this paragraph.)

docs/encyclopedia/application-message-passing.mdx
dandavison: My suggestion (which I didn't make clear) is probably just to say "you have at least two options" or "options include" or something.

dandavison: (That pattern is interesting though.)

docs/encyclopedia/application-message-passing.mdx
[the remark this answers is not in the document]
drewhoskins-temporal: Will need more precise suggestions if you want the diagram changed.  Please follow up on slack.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -139,76 +197,161 @@ Stack Trace Queries are available only for running Workflow Executions.
...
+
+#### Running asynchronous tasks
+
+Sometimes, you need your message handler to wait for long-running operations such as executing an Activity. When this happens, the handler will yield control back to [the loop](#message-handler-concurrency). This means that your handlers can have race conditions if you’re not careful.
+You can guard your handlers with concurrency primitives like mutexes or semaphores, but you should use versions of these primitives provided for Workflows in most languages. See the links below for examples of how to use them in your SDK.
+
+#### Inject work into the main Workflow
+
+Sometimes you want to process work provided by messages in the main Workflow. Perhaps you’d like to accumulate several messages before acting on any of them. For example, message handlers might put work into a queue, which can then be picked up and processed in an event loop that you yourself write.
+This option is considered advanced but offers powerful flexibility. And if you serialize the handling of your messages inside your main Workflow, you can avoid using concurrency primitives like mutexes and semaphores. See the links above for how to do this in your SDK.
```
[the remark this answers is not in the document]
drewhoskins-temporal: For the cohort of people who don't like concurrency that you pointed out above.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -139,76 +197,161 @@ Stack Trace Queries are available only for running Workflow Executions.
...
+    <RelatedReadItem path="/develop/typescript/message-passing#handler-patterns" text="Author message handler patterns in Typescript" archetype="feature-guide" />
+    <RelatedReadItem path="/develop/dotnet/message-passing#handler-patterns" text="Author message handler patterns in .NET" archetype="feature-guide" />
+</RelatedReadContainer>
+
+### Update Validators {#update-validators}
+
+When you define an Update handler, you may optionally define an Update Validator, a read operation that's responsible for accepting or rejecting the Update. You can use Validators to verify arguments or make sure the Workflow is ready to accept your Updates.
+
+- If it accepts, the Update will become part of your Workflow’s history and the client will be notified that the operation has been Accepted. The Update handler will then run until it returns a value.
+- If it rejects or throws an exception, the client will be informed that it was Rejected, and the Workflow will have no indication that it was ever requested, similar to a Query handler.
```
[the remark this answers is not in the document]
drewhoskins-temporal: Not following this suggestion.  Was it inaccurate?

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: Our deduplication does not work across CaN and in my mind that's too much of a gotcha to be able to recommend anything other than good hygiene here.  (we could say "if your workflow doesn't CaN" but then people will copy/paste or the workflow will add CaN later.)
Let's chat offline if we want to keep thinking here.

docs/encyclopedia/application-message-passing.mdx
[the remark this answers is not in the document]
dandavison: Yes -- throwing an exception (or returning an error in Go) _is_ rejection. So "rejects or throws an exception" doesn't make sense.

docs/encyclopedia/application-message-passing.mdx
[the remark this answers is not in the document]
dandavison: The main change I'm making here relates to

> Every time the Workflow wakes up [...] it will process messages in the order they were received, followed by making progress in the Workflow’s main method.

Our main aim in this section is to help users program safely in the face of interleaving, and my feeling is that the above description should be made a bit more accurate / explicit.  Specifically, it gives the impression that message processing happens, then the main workflow. But this downplays the opportunity for interleaving:
what's really happening is that a population of coroutines are set running: handlers first, until they reach their first yield points, then the main workflow coroutine. Thereafter, all coroutines have a chance to execute concurrently, with interleaving, until eventually no further progress can be made. The changes to the diagram would reflect the preceding sentences.

Yes, this is technical, but this is a technical section: we're introducing a concept of the "event loop" of Temporal workflows. I think these changes are not a great increase in technicality over the current content, and will help readers understand the safety challenges they face.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: Re: signal, at least in python, looks like the signal's `request_id` is not exposed and is set to a uuid.  So no ability for a client to combine different callsites and I'm not sure it's worth recommending.  But maybe we should open tasks to make it consistent with updates?

drewhoskins-temporal: Okay, I decided to be more verbose about it because if people try to test their deduplication, they may run into trouble without fully understanding what's going on.

drewhoskins-temporal:
I believe I've addressed or considered all the feedback, and I've uploaded a new image built by the docs team + @dandavison 
Logistics note: I'm going on vacation tomorrow, and we would like to get this in so that the per-language docs can be tested with it.
We're going to do another update to these docs in the coming days/weeks as we ship public preview (deadline Aug 6).  I welcome as much more feedback as people have, but hope that it can be addressed in the next PR unless it is a shipstopper.

dandavison (commented):
Great! I have two high-level concerns / things to address in the future

1. The document is quite detailed, which is a good thing, but a downside is that it risks making handling messages seem complex/sophisticated. Perhaps we should emphasize that synchronous handlers are still useful and generally simple to write safely (although not if they are enqueing work...).

2. As mentioned inline, the situation regarding deduplication / idempotency seems unsatisfactory. We're recommending a bunch of additional steps and users are going to wonder why the SDK doesn't do it itself if it's necessary.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -87,46 +119,74 @@ Otherwise, a new Workflow Execution is started and immediately sent the Signal.
...
+:::
 
-**What is a Stack Trace Query in the context of Temporal?**
+Updates can be sent from a Temporal Client or the Temporal CLI to a Workflow Execution. This call is synchronous and will call into the corresponding Update handler. If you’d rather make an asynchronous request, you should use Signals.
+
+In most languages (except Go), you may call `executeUpdate` to complete an Update and get its result.
+
+Alternatively, to start an Update, you may call `startUpdate` and pass in the Workflow Update Stage as an argument. You have two choices on what to await:
+
+- Accepted - wait til the Worker is contacted, which ensures that the Update is persisted. See [Update Validators]([link]) for more information.
```
angelazhou32, line 158: Missing link?

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -139,76 +199,173 @@ Stack Trace Queries are available only for running Workflow Executions.
...
+
+<PrettyImage src="/img/messages-workflow-loop.png" title="Diagram that shows the execution ordering of Workflows" />
+
+Every time the Workflow wakes up--generally, it wakes up when it needs to--it will process messages in the order they were received, followed by making progress in the Workflow’s main method.
+
+This execution is on a single thread–while this means you don’t have to worry about parallelism, you do need to worry about concurrency if you have written Signal and Update handlers that can block. These can run interleaved with the main Workflow and with one another, resulting in potential race conditions. These methods should be made reentrant.
+
+### Message handler patterns {#message-handler-patterns}
+
+Here are several common patterns for write operations, Signal and Update handlers. They don't apply to pure read operations, i.e. Queries or [Update Validators](#update-validators)):
```
angelazhou32, line 220: You have double paren on (#update-validators)).

```suggestion
Here are several common patterns for write operations, Signal and Update handlers. They don't apply to pure read operations, i.e. Queries or [Update Validators](#update-validators):
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -87,46 +119,74 @@ Otherwise, a new Workflow Execution is started and immediately sent the Signal.
...
 
-**What is a Stack Trace Query in the context of Temporal?**
+Updates can be sent from a Temporal Client or the Temporal CLI to a Workflow Execution. This call is synchronous and will call into the corresponding Update handler. If you’d rather make an asynchronous request, you should use Signals.
+
+In most languages (except Go), you may call `executeUpdate` to complete an Update and get its result.
+
+Alternatively, to start an Update, you may call `startUpdate` and pass in the Workflow Update Stage as an argument. You have two choices on what to await:
+
+- Accepted - wait til the Worker is contacted, which ensures that the Update is persisted. See [Update Validators]([link]) for more information.
+- Completed - wait til the handler finishes and returns a result. (This is equivalent to `executeUpdate`.)
```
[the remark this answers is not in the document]
drewhoskins-temporal: I'm torn as to whether we should pollute the overview docs with that information.  I'm thinking we just call it out explicitly in the typescript docs.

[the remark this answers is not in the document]
dandavison: That's fine. I'm not even sure TS should be departing in that way.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -26,57 +28,87 @@ keywords:
...
 
-Signal handlers are Workflow functions that listen for Signals by the Signal name.
-Signals are delivered in the order they are received by the Temporal Service and written to History.
-If multiple deliveries of a Signal would be a problem for your Workflow, add idempotency logic to your Signal handler that checks for duplicates.
+- Queries are efficient–they never add entries to the [Workflow Event History](/workflows#event-history), whereas an Update would (if accepted).
+- Queries can operate on completed Workflows.
 
-[^1]: The Temporal Service usually deduplicates Signals, but does not guarantee deduplication: During shard migration, two Signal Events (and therefore two deliveries to the Workflow Execution) can be recorded for a single Signal because the deduping info is stored only in memory.
+However, because Queries cannot block, sometimes Updates are best.
+When your goal is to do a read once the Workflow achieves a certain desired state, you have two options.
```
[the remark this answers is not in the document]
drewhoskins-temporal: Well, this is in the "for read requests" sub-heading.  I'm thinking I'll add a "### For read/write requests" sub-heading

docs/encyclopedia/application-message-passing.mdx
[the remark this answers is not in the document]
drewhoskins-temporal: Dan helped me update the diagram.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -139,76 +199,173 @@ Stack Trace Queries are available only for running Workflow Executions.
...
+
+Many developers want their message handlers to run exactly once--to be idempotent--in cases where the same Signal or Update is delivered twice or sent by two different call sites. To do this, you can use an idempotency key.
+
+For Signals, you can pass a unique key among your message arguments and use that to check for duplicate messages in your handler.
+
+For Updates, you can additionally retrieve the Update ID within your handler and use it to check whether the Workflow has already handled the Update. Clients randomize this Update ID by default, which means you'll be deduplicating retries from the same call. If you'd like multiple calls to share an idempotency key, such as when your client itself retries, pass in an Update ID of your own choosing.
+
+:::note
+
+Temporal deduplicates messages for you on the server. However, this doesn't work across [Continue-As-New](/workflows#continue-as-new) boundaries, meaning you would risk processing messages twice for such Workflows if you don't do this yourself.
```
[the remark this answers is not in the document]
drewhoskins-temporal: Made it a bit less intimidating/more well-structured.

dandavison (approved):
LGTM! Important, large and in places tricky topic which I don't think we'd ever covered in depth and in this unified manner before.

angelazhou32 (approved):

axfelix:
PS -- in the future, we need to get @drewhoskins-temporal a Vercel seat, and I'd strongly encourage making branches on this repo rather than on forks -- both of these will make actually generating previews much easier.

axfelix:
And FYI, the build is failing right now (due to a bad link), which is exactly why we like to do work on this repo rather than on forks and need to make sure everyone who's going to work on docs has a Vercel seat. @drewhoskins-temporal , as it stands I would have to open a PR on your fork to fix this before merging, which is not ideal — do you want me to do that, or can you take a look at the failure and sort it out?

dandavison:
I've updated the [diagram](https://lucid.app/lucidchart/9c1e9b18-4656-4513-8154-7a23cee7af85/edit?viewport_loc=-523%2C192%2C1963%2C1139%2CytZWJ9T~VQ8V&invitationId=inv_0a10ce4e-ed82-46a6-903a-e00b1fe25485) to move the "Signals, Updates, Queries" arrow to the left rather than the right (It's better practice to have causality flowing left-to-right). 

Here's a screenshot; the original is in lucidchart.

image

PR #2962, opened 2024-07-16
EDU-2760: Use schedules instead

fairlydurable:


napcs (approved):

fairlydurable:
@napcs Before merging I did a review of the pages and found some issues and corrected them ("leave it better than you found it"). Can you do another look over for sanity?

PR #2978, opened 2024-07-29
Golang scheduler deletion example: Remove unnecessary error check for GetHandle()

ulascansenturk:
## What does this PR do?

Error handling for GetHandle():
The original code checks for an error after calling GetHandle(), but GetHandle() doesn't return an error. This check was unnecessary and has been removed.


jsundai:
> Hello @jsundai i am authorizing the deployment on vercel but it doesn't do anything, and CI is failing because of that.

Thanks for the contribution! Authorizing deployment on Vercel is for internal folks but we'll make sure to review this PR in this sprint.

napcs (approved):

PR #2980, opened 2024-07-30
Fixed PHP example code syntax

gplanchat:
## What does this PR do?

Fix a syntax error in the PHP examples

## Notes to reviewers

Arrow functions in PHP has an implicit return statement : https://www.php.net/manual/en/functions.arrow.php


roxblnfk (approved):

jsundai (approved):

jsundai:
Thanks for the contribution, @roxblnfk!

PR #2981, opened 2024-07-30
Multi-tenant production feature landing page 

jsundai:


docs/evaluate/development-production-features/multi-tenant.mdx
```diff
@@ -0,0 +1,39 @@
+---
+id: multi-tenant
+title: Multi-tenant - Temporal feature
```
axfelix, line 3: I think "Multi-tenancy" might be a better name for the actual feature, but if this has already been decided, ignore me on this one

docs/evaluate/development-production-features/multi-tenant.mdx
```diff
@@ -0,0 +1,39 @@
...
+- scalability
+- cost-effectiveness
+- noisy neighbor 
+- database performance
+- high throughput
+---
+
+import { RelatedReadContainer, RelatedReadItem } from '@site/src/components/related-read/RelatedRead';
+
+Temporal Cloud's architecture scales to handle multiple tenants efficiently.
```
axfelix, line 22: Seems a bit strange to lead with "Temporal Cloud" here and let people figure out by implication that this is a Cloud feature... I notice the page before this one in the sidebar (Low latency) is like that too. Are we just putting Cloud features side by side with platform features in the Evaluate section?

docs/evaluate/development-production-features/multi-tenant.mdx
```diff
@@ -0,0 +1,39 @@
...
+Our features enable you to handle multiple Namespaces sharing the same compute and persistence layer, ensuring performance and operational efficiency.
```
axfelix, line 23: Sorry for being nitpicky on this, but this is also stylistically strange to me! "Our features" seems marketing-y in this context... I'd rather see this written as "this feature"

docs/evaluate/development-production-features/multi-tenant.mdx
```diff
@@ -0,0 +1,39 @@
...
+A Namespace is a unit of isolation within the Temporal Platform.
```
axfelix, line 24: I think we can rewrite the rest of this to flow a bit better and emphasize what's good about what we're doing here... here's my attempt:

A Namespace is a unit of isolation within the Temporal Platform -- but even a single Namespace is still multi-tenant. Multi-tenancy ensures extra capacity is available for all customers during traffic spikes.

However, multi-tenancy can also presents the challenge of "noisy neighbors", where high-traffic tenants consume excess resources, causing slower performance for other tenants. This is a common problem for scaling database scaling. 

Temporal's write-heavy workload, where changes in execution state are constantly written to the persistence layer, demands a database that supports reliably high throughput with low latency for multiple customers, concurrently and fairly. With Temporal Cloud, customers pay for consumption instead of entire sets of hardware, providing a cost-effective solution.

axfelix (commented):
Had some nitpicks on this one! Hope they're useful :)

docs/evaluate/development-production-features/multi-tenant.mdx
jsundai: This is fantastic and definitely agree this flows better! 🚀

axfelix (approved):

docs/evaluate/development-production-features/multi-tenant.mdx
jsundai: Not neccessarily. Put we are grouping cloud features with self hosted (all of them together) in the features section. Open to any ideas about recategorization as the sidebar gets longer.

PR #2989, opened 2024-08-01
Fix typos in docs

jrobles98:
## What does this PR do?

Just fixed some little typos on the documentation.

Also here is the regex to check if any markdown link has a letter right after them:
`\[.*?\]\([^()]*\)[a-zA-Z]`
matches `[link](url)ups` therefore rendering as '[link](url)ups'


fairlydurable (approved):
Thanks for catching these typos. I reviewed all four and  they seem correct and appropriate.

fairlydurable:
Thank you for contributing to the docs site.

PR #2993, opened 2024-08-06
add note about workflow,host,worker specific task queues

axfelix:
Sort of addresses https://temporalio.atlassian.net/browse/EDU-776 -- this had sat empty for a while and I wanted to just get it taken care of without waiting longer on Worker Versioning docs. Feedback welcome.


docs/develop/go/temporal-client.mdx
```diff
@@ -244,6 +244,9 @@ workflowRun, err := c.ExecuteWorkflow(context.Background(), workflowOptions, "Yo
 
 In most SDKs, the only Workflow Option that must be set is the name of the [Task Queue](/workers#task-queue).
 
+You can configure Task Queues that are host-specific, Worker-specific or Workflow-specific to distribute your application load.
```
jsundai, line 247: Maybe move this to after the code sample on the "Set Workflow Task Queue" section. It might be a little confusing if someone is just learning Temporal and trying to figure out how to set a task queue.

jsundai: Also just noticing wierd with the header sidebar repeating for this page. Probably done after the auto-migration, so might need to be fixed. I'll have to check if I manually fixed it in the Editing Go PR.

axfelix: I'd had it there in the first place but didn't want to break the convention of the page where the code sample is always last in each section! Agree it belongs better there.

jsundai (approved):

PR #2999, opened 2024-08-07
add ap-south-2 region to documentation.

napcs:


fairlydurable (approved):
Checked all updates. Confirmed hash with Brian. Stamping.

PR #3002, opened 2024-08-09
EDU-2934: AUG 12 PUB updates coverage for automatic failbacks

fairlydurable:
MRN update. Source of truth: @atihkin 

Also fixes consistent capitalization of failover and failback.


docs/glossary.md
```diff
@@ -224,7 +228,7 @@ _Tags: [product-release-stages](/tags/product-release-stages), [term](/tags/term
 
 A Global Namespace is a Namespace that duplicates data from an active [Temporal Service](#temporal-cluster) to a standby Service using the replication to keep both Namespaces in sync.
 Global Namespaces are designed to respond to service issues like network congestion.
-When service to the primary Cluster is compromised, a [Failover](#failover) transfers control from the active to the standby cluster.
+When service to the primary Cluster is compromised, a [failover](#failover) transfers control from the active to the standby cluster.
```
angelazhou32, line 231: Cluster or service? Just checking!

angelazhou32 (approved):

angelazhou32 (approved):

docs/glossary.md
```diff
@@ -183,10 +183,14 @@ An append-only log of Events that represents the full state a Workflow Execution
 
 _Tags: [term](/tags/term), [explanation](/tags/explanation)_
 
+#### [Failback](/cloud/multi-region)
+
+After Temporal Cloud has resolved an outage or incident involving a failover, a failback process shifts Workflow Execution processing back to the original Temporal Namespace that was active before the incident.
```
atihkin, line 188: Small distinction here (my mistake I think I drafted this text). We should probably say "shifts Workflow Execution processing back to the original region that was active before the incident." because it's still the same Namespace, just the region is different.

atihkin: But I realize we have defined Failover similarly as well -- so this is more of a nit. Feel free to ignore if this increases the scope of the change too much.

fairlydurable: Fixing both.

fairlydurable:
@atihkin 

```
#### [Failover](/cloud/multi-region)

A failover shifts Workflow Execution processing from an active Temporal Namespace region to a standby Temporal Namespace region during outages or other incidents.
Standby Namespace regions use replication to duplicate data and prevent data loss during failover.
```

PR #3004, opened 2024-08-09
adds data converter stuff

angelazhou32:
This adds some info about custom payload converter into docs from the ts course.


docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,82 @@ tags:
...
-- [new Client(\{ ..., dataConverter })](https://typescript.temporal.io/api/interfaces/client.ClientOptions#dataconverter)
-- [Worker.create(\{ ..., dataConverter })](https://typescript.temporal.io/api/interfaces/worker.WorkerOptions#dataconverter)
+## Composite Data Converters
 
-A Data Converter has two parts:
+Use a [Composite Data Converter](https://typescript.temporal.io/api/classes/common.CompositePayloadConverter) to apply custom, type-specific Payload Converters in a specified order. Defining a new Composite Data Converter is not always necessary to implement custom data handling. You can override the default Converter with a custom Codec, but a Composite Data Converter may be necessary for complex Workflow logic.
 
-- [Payload Converter](#payload-converter): Sync methods that sometimes run inside the Workflow isolate (and are thus limited).
-- [Payload Codec](#payload-codec): Async methods that run outside the isolate.
+A Composite Data Converter can be comprised of custom rules that you created, and it can also
```
napcs, line 50: passive voice.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,82 @@ tags:
...
+      new UndefinedPayloadConverter(),
+      new BinaryPayloadConverter(),
+      new JsonPayloadConverter()
+    );
+  }
 }
 ```
 
-### Payload Converter
+The order in which the Payload Converters are applied is important. During
```
napcs, line 67: first sentence is passive-voice

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,82 @@ tags:
...
+export const payloadConverter = new CompositePayloadConverter(
+  new UndefinedPayloadConverter(),
+  new EjsonPayloadConverter()
+);
+```
+
+## How to use a custom payload converter in TypeScript {#custom-payload-conversion}
+
+To support custom Payload conversion, create a [custom Payload Converter](/dataconversion#composite-data-converters) and configure the Data Converter to use it in your Client options.
+Custom Payload Converters can be used to change how application objects get serialized to binary Payload. To handle custom data types that are not natively JSON-serializable (e.g., `BigInt`, `Date`, or binary data), you can create a custom Payload Converter. A Custom Payload Converter is responsible for converting your custom data types to a payload format that Temporal can manage.
```
napcs, line 83:
```suggestion
You can use Custom Payload Converters to change how application objects get serialized to binary Payload. To handle custom data types that are not natively JSON-serializable (e.g., `BigInt`, `Date`, or binary data), you can create a custom Payload Converter. A Custom Payload Converter is responsible for converting your custom data types to a payload format that Temporal can manage.
```

recasted to active voice

napcs (changes requested):

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,82 @@ tags:
...
+A Composite Data Converter can be comprised of custom rules that you created, and it can also
+leverage the default Data Converters built into Temporal. In fact, the default
+Data Converter logic is implemented internally in the Temporal source as a
+Composite Data Converter. It defines these rules in this order:
 
-```ts
-interface DataConverter {
-  payloadConverterPath?: string;
-  payloadCodecs?: PayloadCodec[];
+```typescript
```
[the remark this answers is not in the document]
angelazhou32: It comes from the course repo

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,111 @@ tags:
...
+
+In practice, these two components operate in a chain to handle data securely. Incoming data first passes through a Payload Converter through the `toPayload` method, turning application objects into Payloads. These Payloads are then processed by the Payload Codec through the `encode` method, which adjusts the Payload according to the required security or efficiency needs before it is sent to the Temporal Cluster.
+
+The process is symmetric for outgoing data. Payloads retrieved from the Temporal Cluster first pass through the Payload Codec through the `decode` method, which reverses any transformations applied during encoding. Finally, the resulting Payload is converted back into an application object by the Payload Converter through the `fromPayload` method, making it ready for use within the application.
+
+## Use custom payload conversion
 
 Temporal SDKs provide a [Payload Converter](/dataconversion#payload-converter) that can be customized to convert a custom data type to a [Payload](/dataconversion#payload) and back.
 
 Implementing custom Payload conversion is optional.
```
axfelix, line 38: I think this sentence is the same in the other SDKs but I wonder if we should rewrite it to emphasize that it's not just "optional," it's also not necessary to provide a new payload converter even if you're using a custom codec. I tried to rewrite I think at least Go and Java a bit to emphasize this, there might be examples to draw on there.

axfelix:
Looks good, though I think the Payload Codec section at the bottom of this doc should have more emphasis (after this PR it doesn't have a righthand nav H2/H3), and maybe even moved up to all the way before "use custom payload conversion." I know I argued with Chad about this at some point, but if we're going to call this page "Data encryption," we should probably present custom codecs *before* custom converters, since they're easier to understand and what 80% of people who come to this page will need.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+The process is symmetric for outgoing data. Payloads retrieved from the Temporal Cluster first pass through the Payload Codec through the `decode` method, which reverses any transformations applied during encoding. Finally, the resulting Payload is converted back into an application object by the Payload Converter through the `fromPayload` method, making it ready for use within the application.
 
-A Data Converter has two parts:
+## Payload Codec
 
-- [Payload Converter](#payload-converter): Sync methods that sometimes run inside the Workflow isolate (and are thus limited).
-- [Payload Codec](#payload-codec): Async methods that run outside the isolate.
+> API documentation: [PayloadCodec](https://typescript.temporal.io/api/interfaces/common.PayloadCodec)
+
+The default `PayloadCodec` does nothing. To create a custom one, we implement the following interface:
```
fairlydurable, line 38: Avoid using "we". "We" can imply you're speaking from a position of authority or shared experience. Prefer "you". The person taking action is the reader.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
 description: Learn how to create a custom Payload Converter in TypeScript with Temporal SDKs to handle non-JSON-serializable values, configure your Data Converter, and utilize protobufs and encryption seamlessly in your Workflows and Activities.
 ---
 
-## How to use a custom payload converter in TypeScript {#custom-payload-conversion}
+## Payload Converter and Payload Codec Summary
 
-Temporal SDKs provide a [Payload Converter](/dataconversion#payload-converter) that can be customized to convert a custom data type to a [Payload](/dataconversion#payload) and back.
+This section will summarize the difference between a Payload Converter and Payload Codec.
```
fairlydurable, line 16: Avoid modal auxilliary (will)

"This section summarizes" presents the information as immediately relevant and actionable.

fairlydurable: I think it would help to add a summary to your intro to prep your readers for the main topic. For example: 

"Payloads are structured data transmitted between components in a Temporal installation. Sending payloads as plain text exposes security risks and increases bandwidth usage. Payload codecs and converters address these issues. Codecs provide a standardized format for data transmission, while converters handle the process of converting data into and out of that format. This page describes how to create custom Payload Converters in TypeScript to manage values that cannot be serialized with JSON. You'll configure a Data Converter for serialization and integrate Protobufs and encryption into your Workflows and Activities."

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+
+### Payload Converter
 
-Implementing custom Payload conversion is optional.
-It is needed only if the [default Data Converter](/dataconversion#default-data-converter) does not support your custom values.
+Payload Converters are responsible for serializing application objects into a Payload and deserializing them back into application objects. A Payload, in this context, is a binary form suitable for network transmission that may include some metadata. This serialization process transforms an object (like those in JSON or Protobuf formats) into a binary format and vice versa. For example, an object might be serialized to JSON with UTF-8 byte encoding or to a protobuf binary using a specific set of protobuf message definitions.
```
fairlydurable, line 20: I don't think you need to use "are responsible for". I think they "serialize application objects" (assuming objects is the correct phrase for data types here). You capitalized Payload, which is not a reserved Temporal term. I think the plural "into payloads, and deserialize them back" also makes the writing smoother. 

I like the definition of the payload. Consider splitting metadata into a separate thought. The next sentence "This serialization process" feels redundant to me. You might put the focus on JSON or Protobuf instead.  JSON, with its human-readable structure, is a common data interchange solution for web APIs due to its simplicity and ease of integration. Protobufs offer efficient binary serialized formats that reduce payload sizes and support fast processing. That's ideal for high-performance, etc etc. 

What I don't see is how this connects to the next line though. Serialization/deserialization is non-deterministic, so converters can live in Workflows, but encryption and other types of novel encoding, cannot, right? So there needs to be a second piece that can handle non-deterministic logic, which lives outside the Workflow, if I'm understanding this correctly. And it would help to have words here that explain the difference.

"Because of this" is I think better than "Consequently", we prefer "can't" to "cannot", "use" to "employ". And then I think a breath is called for. Maybe new sentence "This excludes (most) encryption because encryption is almost always non-determ."

And then there's the missing piece that leads to "Well, what do you do with the part that can't live inside the Workflow", a transition that gets you from converters to codecs.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
 
-The order in which your encoding Payload Converters are applied depend on the order given to the Data Converter.
-You can set multiple encoding Payload Converters to run your conversions.
-When the Data Converter receives a value for conversion, it passes through each Payload Converter in sequence until the converter that handles the data type does the conversion.
+### Payload Codec
+
+Payload Codecs transform one Payload into another, converting binary data to a different binary format. Unlike Payload Converters, Payload Codecs do not operate within the Workflow sandbox. This allows them to execute operations that can include calls to remote services and the use of non-deterministic modules, which are critical for tasks such as encrypting Payloads, compressing data, or offloading large payloads to an object store. Payload Codecs can also be implemented as a Codec Server (which will be described later on).
```
fairlydurable, line 26: This is where I'd get more confused because serialization is also about converting binary data to a different binary format because all data is binary. Whether you're using compression/decompression or encryption, they all (potentially) fall under the non-determ umbrealla. Oh, and you mention some really good examples here of what those tasks are. 

I'm not sure what our group's position is on forward references ("which will be described") but I've had my hand slapped a lot about that.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+
+### Operational Chain
 
-To send values that are not [JSON-serializable](https://en.wikipedia.org/wiki/JSON#Data_types) like a `BigInt` or `Date`, provide a custom [Data Converter](https://typescript.temporal.io/api/interfaces/common.DataConverter/) to the Client and Worker:
+In practice, these two components operate in a chain to handle data securely. Incoming data first passes through a Payload Converter through the `toPayload` method, turning application objects into Payloads. These Payloads are then processed by the Payload Codec through the `encode` method, which adjusts the Payload according to the required security or efficiency needs before it is sent to the Temporal Cluster.
```
fairlydurable, line 30: Maybe stating or explaining why data security is important to many of our adopters before just jumping into chaining? And then instead just jump in "chaining these components allows ..." Also, avoid passive voice, as part of our in-house style, and you can say Payloads instead of "These payloads" and rework the sentence so the actor is the Payload Codec type, which uses its `encode` method to process the payload. The work applied depends on the security or efficiency needed before sending it onto the (I assume) Temporal Service.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
 
-- [new Client(\{ ..., dataConverter })](https://typescript.temporal.io/api/interfaces/client.ClientOptions#dataconverter)
-- [Worker.create(\{ ..., dataConverter })](https://typescript.temporal.io/api/interfaces/worker.WorkerOptions#dataconverter)
+The process is symmetric for outgoing data. Payloads retrieved from the Temporal Cluster first pass through the Payload Codec through the `decode` method, which reverses any transformations applied during encoding. Finally, the resulting Payload is converted back into an application object by the Payload Converter through the `fromPayload` method, making it ready for use within the application.
```
fairlydurable, line 32: A figure might help to show the parallel and reversed process. Also you're using specific calls while you're still introducing the process. Moving them out both makes the intro easier to read and also supports porting to other SDKs (I don't know where this page lives or how it will be or not be ported)

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
 ```
 
-### Payload Converter
+## Use custom payload conversion
 
-> API documentation: [PayloadConverter](https://typescript.temporal.io/api/interfaces/common.PayloadConverter)
+Temporal SDKs provide a [Payload Converter](/dataconversion#payload-converter) that can be customized to convert a custom data type to a [Payload](/dataconversion#payload) and back.
 
-```ts
+The order in which your encoding Payload Converters are applied depending on the order given to the Data Converter.
```
fairlydurable, line 59: Also, the sentence doesn't really make sense "applied depending on". Maybe flip the sentence around? The Data Converter applies the Payload Converters in a specific order, which you pass to it?" or something? It's confusing.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+You can set multiple encoding Payload Converters to run your conversions.
```
fairlydurable, line 60: Set where? Add them to the Data Converter? Create a chain?

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+When the Data Converter receives a value for conversion, the value gets passes through each Payload Converter in sequence until the converter that handles the data type does the conversion. You will explore more in detail now. You will explore more in detail now.
```
fairlydurable, line 61: Does the value enter the Data Converter? Does a value get sent to the Data converter?  After sending a value to the Data Converter, the converter passes it through each of its Payload Converters (do these live in a queue?) until the "converter that handles the data type" which one? Both Data and Payload are converters. 

You will explore more in detail now is hard to understand. Maybe "In the following sections, you'll explore this process in more detail?" And the words are duplicated, the last two sentences on the line.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+    super(
+      new UndefinedPayloadConverter(),
+      new BinaryPayloadConverter(),
+      new JsonPayloadConverter()
+    );
+  }
+}
+```
+
+The order of applying the Payload Converters is important. During
```
fairlydurable, line 84: Under what circumstances is the payload null? What does that tell you? How do you handle it?

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+```
+export const payloadConverter = new CompositePayloadConverter(
+  new UndefinedPayloadConverter(),
+  new EjsonPayloadConverter()
+);
+```
+
+You can do this in its own `payload-conterter.ts` file for example.
+
+In the code snippet above, a converter is created that first attempts to handle `null` and `undefined` values. If the value isn't `null` or `undefined`, the EJSON serialization logic written in the `EjsonPayloadConverter` is then used. The Payload Converter is then provided to the Worker and Client.
```
fairlydurable, line 99: The voice of previous editors says "previous" not "above" for some reason, although if you think about it, there's no way this is being split onto separate pages (like a book) so maybe that doesn't apply here.

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+
+Here is the Worker code:
```
fairlydurable, line 101: Maybe move the sentence below the code fencing into this sentence? This worker serializes and deserializes...?

docs/develop/typescript/data-encryption.mdx
```diff
@@ -11,41 +11,129 @@ tags:
...
+
+```typescript
+const client = new Client({
+  dataConverter: {
+    payloadConverterPath: require.resolve("./payload-converter"),
+  },
+});
+```
+
+You can now use a variety of data types in arguments.
```
fairlydurable, line 125: This is important because...?

docs/develop/typescript/data-encryption.mdx
```diff
@@ -60,6 +148,10 @@ interface PayloadConverter {
 }
 ```
 
+This custom converter should include logic for both serialization (`toPayload`) and deserialization (`fromPayload`), handling your specific data types or serialization format. The method `toPayload` returns a Payload object, which is used to manage and transport serialized data. The method `fromPayload` returns the deserialized data. This ensures that the data returned is in the same format as it was before serialization, allowing it to be used directly in the application.
+
+2. Configure the Data Converter.  To send values that are not JSON-serializable like a `BigInt` or `Date`, provide the custom Data Converter to the Client and Worker as described in the [Composite Data Converters](#composite-data-converters) section.
```
fairlydurable, line 153: Extra spaces here? 

Also might help to introduce a reason before sending those values. Some values cannot be expressed in JSON primitives like BigInt or Date. To send values like these, which are not JSON-serializable, ....

PR #2976, opened 2024-07-26
SDK-1486: Python update docs

dandavison:
[[rendered](https://temporal-documentation-aj72f1nyt.preview.thundergun.io/develop/python/message-passing)]

This is a new version of the message-passing documentation for Python. It follows on from the new top-level section on https://docs.temporal.io/encyclopedia/workflow-message-passing that @drewhoskins-temporal recently wrote. The significant changes here are:

- There is now a code sample at the top of the page that will immediately give many visitors to the page what they're looking for: it is minimal, but it shows how to define Query, Signal, and Update handlers, and an Update validator, and has one-line comments introducing the concepts.

- Following the style taken in https://docs.temporal.io/encyclopedia/workflow-message-passing, the document recognizes that many aspects of Signal and Update handling can be covered together. This avoids a lot of repetition.

- In particular, a new section named `Message handler patterns` gives Python-specific discussion and examples of the content outlined in [encyclopedia/workflow-message-passing](https://docs.temporal.io/encyclopedia/workflow-message-passing).

- Similarly, there was no need to discuss dynamic handlers separately for Signal, Query, and Update since the signatures are the same.

The structure of the document is (I'll keep this up-to-date it it changes)

```
## Writing message handlers
  ### Query handlers {#handle-query}
  ### Signal handlers {#handle-signal}
  ### Update handlers and validators {#handle-update}

## Sending messages
  ### Sending a Query {#send-query}
  ### Sending a Signal
    #### Sending a Signal from a Client {#send-signal-from-client}
    #### Sending a Signal from a Workflow {#send-signal-from-workflow}
    #### Signal-With-Start {#signal-with-start}
  ### Sending an Update {#send-update}
    #### 1. Use [`execute_update`](https://python.temporal.io/temporalio.client.WorkflowHandle.html#execute_update) to wait for the update to complete
    #### 2. Use [`start_update`](https://python.temporal.io/temporalio.client.WorkflowHandle.html#start_update) to receive a handle as soon as the update is accepted or rejected
    #### Non-type safe APIs

## Message handler patterns {#message-handler-patterns}
  ### Async handlers {#async-handlers}
    #### Waiting
    #### Use `asyncio.Lock` to prevent concurrent handler execution {#control-handler-concurrency}
    #### Finishing handlers before the Workflow completes {#wait-for-message-handlers}

## Troubleshooting
  ### Problems when sending a Signal
  ### Problems when sending an Update
  ### Problems when sending a Query
## Dynamic Handlers {#dynamic-handler}
[...]

PR #3006, opened 2024-08-12
SDK-2426: .NET message passing docs refresh

cretz:
## What does this PR do?

* Copied message-passing section from #2976 (as of when this branch was created) and altered for .NET
* In Python version, there are lots of API links to python.temporal.io since they have a single stable link even for overloads. However, for many .NET items with overloads, there would be ambiguous URLs, so we did not link those which would be confusing/ambiguous.
* In Python version, dynamic workflow/activity were included which IMO is not valid. Left off here because they are already in the proper core-application spot for .NET docs.

### DIFF
--- a/docs/develop/dotnet/message-passing.mdx
+++ b/docs/develop/dotnet/message-passing.mdx
@@ -420,30 +420,41 @@ public class MyActivities
 [Workflow]
 public class GreetingWorkflow
 {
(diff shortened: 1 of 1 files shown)


dandavison (approved):
=> docs team!

dandavison:
Superseded by #3048

PR #3014, opened 2024-08-14
duplicate workflow clarification

axfelix:


docs/encyclopedia/workflows.mdx
```diff
@@ -626,7 +626,11 @@ The Workflow Id Reuse Policy can have one of the following values:
 - **Terminate if Running:** Specifies that if a Workflow Execution with the same Workflow Id is already running, it should be terminated and a new Workflow Execution with the same Workflow Id should be started. This policy allows for only one Workflow Execution with a specific Workflow Id to be running at any given time.
 
 The first three values (Allow Duplicate, Allow Duplicate Failed Only, and Reject Duplicate) of the Workflow Id Reuse Policy apply to Closed Workflow Executions that are retained within the Namespace.
-For example, the Temporal Service can only check the Workflow Id of the spawning Workflow Execution based on the Workflow Id Reuse Policy against the Closed Workflow Executions for the last 30 days given a default Retention Period.
+For example, the Temporal Service can only check the Workflow Id of the spawning Workflow Execution based on the Workflow Id Reuse Policy against the Closed Workflow Executions for the last *30 days* given a default Retention Period. 
+
+If you need to start a Workflow for a particular implementation only if it hasn't started yet, pay close attention to your Retention Period.
```
fairlydurable, line 631: I find this line completely confusing. Pay close attention to what in the retention period? Why? What does the retention period tell you? How would you respond? When would you do this? Why? 

Would combining this line with the next one make this less confusing? Who wants to start Workflows atomically, and why are they trying to do this?

fairlydurable (approved):
This is one of those situations where there are no procedural or docs-specific issues that would block this moving forward but I don't understand either the ticket or our response. 

I'm approving it to unblock you and trusting that you know what you're doing.

docs/encyclopedia/workflows.mdx
axfelix: Yeah, good points, I'll go into more detail here.

PR #3016, opened 2024-08-14
EDU-2853: clarify local activity execution failure in references

axfelix:


fairlydurable (approved):
Looks fine to me. Noting that: 1. don't is less formal than does not.  (Google contractions) and 2. "use parentheses judiciously".

I don't think either one should block this moving forward.

Thank you for making this fix.

PR #3023, opened 2024-08-19
EDU-2837: clarify datadog subsection in cloud docs is a tutorial

axfelix:
Longer term we should probably consider moving this to the Learn site with the other tutorials and replacing this with more reference-style content in docs.

### DIFF
--- a/docs/production-deployment/cloud/metrics/datadog.mdx
+++ b/docs/production-deployment/cloud/metrics/datadog.mdx
@@ -34,8 +34,7 @@ Temporal's integration with Datadog extends the monitoring capabilities of your 
 
 :::note
 
-This tutorial provides an example application to help users capture metrics from Temporal's PromQL endpoint and export to  DataDog.
-Temporal Cloud's Observability endpoint supports PromQL only. 
+This tutorial provides an example application to help users capture metrics from Temporal's PromQL endpoint and export to DataDog.
 
 :::
 
--- a/docs/production-deployment/cloud/metrics/datadog.mdx
+++ b/docs/production-deployment/cloud/metrics/datadog.mdx
@@ -32,6 +32,13 @@ tags:
 Exporting cloud metrics to Datadog provides enhanced observability, allowing you to monitor, alert, and visualize key performance indicators of your applications and infrastructure.
 Temporal's integration with Datadog extends the monitoring capabilities of your Temporal Cloud deployment.
 
+:::note
+
+This tutorial provides an example application to help users capture metrics from Temporal's PromQL endpoint and export to  DataDog.
+Temporal Cloud's Observability endpoint supports PromQL only. 
+
+:::
+
 ### What will you learn?
 
 You will set up your environment to export metrics from Temporal Cloud to Datadog, including:


docs/production-deployment/cloud/metrics/datadog.mdx
```diff
@@ -32,6 +32,13 @@ tags:
 Exporting cloud metrics to Datadog provides enhanced observability, allowing you to monitor, alert, and visualize key performance indicators of your applications and infrastructure.
 Temporal's integration with Datadog extends the monitoring capabilities of your Temporal Cloud deployment.
 
+:::note
+
+This tutorial provides an example application to help users capture metrics from Temporal's PromQL endpoint and export to  DataDog.
```
fairlydurable, line 37: 1. Double space on line 37.
2. Consider "This tutorial's [sample app](...) helps you capture..." because it's so indirect the way it's currently worded. It's a reader, not a user, and more specifically it's you, the reader.

docs/production-deployment/cloud/metrics/datadog.mdx
```diff
@@ -32,6 +32,13 @@ tags:
...
+Temporal Cloud's Observability endpoint supports PromQL only.
```
fairlydurable, line 38: 3. The second line really confused me so I went over to Kapa to check. We all know Kapa isn't super accurate but the second line is so definitive that I thought I'd raise this.

```
What data does Temporal Observability emit and what format? Is PromQL the only way to consume it?

Temporal Observability emits metrics data that can be used to monitor performance and troubleshoot issues. This data is emitted by both the Temporal Service and SDKs. The metrics are emitted in a format that is supported by Prometheus by default, but any metrics software that supports the same format can be used. This includes tools like Prometheus, StatsD, and M3. Temporal Cloud emits metrics through a Prometheus HTTP API endpoint, which can be directly used as a Prometheus data source in Grafana or to query and export Cloud metrics to any observability platform[1](https://docs.temporal.io/cloud/metrics/datadog#user-content-fn-1%5E)[2](https://docs.temporal.io/cloud/metrics/datadog#user-content-fn-3%5E).
PromQL is a query language for Prometheus and is used to query these metrics[3](https://docs.temporal.io/cloud/metrics/datadog#user-content-fn-5%5E). However, it's not the only way to consume Temporal's metrics. You can also use Grafana, a popular observability platform, to view these metrics[1](https://docs.temporal.io/cloud/metrics/datadog#user-content-fn-1%5E)[4](https://docs.temporal.io/cloud/metrics/datadog#user-content-fn-2%5E). Additionally, you can set up alerts on these metrics using Prometheus Alertmanager[5](https://docs.temporal.io/cloud/metrics/datadog#user-content-fn-7%5E).
In the case of the Python SDK, metrics are configured globally and you should set a Prometheus endpoint before any other Temporal code[4](https://docs.temporal.io/cloud/metrics/datadog#user-content-fn-2%5E). Here's an example of how to do this:
```

fairlydurable (approved):
I left non-blocking comments, but am approving this PR.

Thank you for taking care of this, Alex.

docs/production-deployment/cloud/metrics/datadog.mdx
axfelix: I believe this is phrased the way it is to emphasize that this is just an unmaintained sample (as I mentioned in my comment, we eventually need to both take over maintenance of this and move it to learn), so this is more of a disclaimer, hence "an example application". I know it's kind of weak language but I might prefer to keep it (and thank you for catching the double space).

docs/production-deployment/cloud/metrics/datadog.mdx
axfelix: Hm. I think it's still technically accurate as stated, but would you be more comfortable with "Temporal Cloud's Observability endpoint provides metrics in PromQL format?"

Or, honestly, I think what I was given to work from here might be more redundant than intended -- I'm going to delete the second sentence, I don't think it adds anything.

docs/production-deployment/cloud/metrics/datadog.mdx
fairlydurable: Works for me.

docs/production-deployment/cloud/metrics/datadog.mdx
fairlydurable: Sounds good to me. (And I think it provides metrics in Prometheus format. PromQL is a query language.)

PR #3030, opened 2024-08-23
add workflow metrics to cluster metrics reference

axfelix:


fairlydurable (approved):
I can't find any issues after reviewing. Approving.

PR #3056, opened 2024-09-03
remove bash/curl install method

axfelix:


axfelix:
For the record -- I use homebrew on other platforms than macOS and am tempted to remove the distinction entirely, but I'm afraid this way is probably still more intuitive.

fairlydurable (approved):
Inspected changes directly. Checked all three pages on Vercel preview.

PR #3100, opened 2024-09-19
Document next retry delay for Python

dandavison:
Document the Activity "next retry delay" feature. Analogous to existing Java documentation: https://docs.temporal.io/develop/java/failure-detection#next-retry-delay

### DIFF
--- a/docs/develop/python/failure-detection.mdx
+++ b/docs/develop/python/failure-detection.mdx
@@ -179,7 +179,7 @@ from temporalio.common import RetryPolicy
         )
 ```
 
-### Override the retry interval with Next Retry Delay {#next-retry-delay}
+### Override the retry interval with `next_retry_delay` {#next-retry-delay}
 
 To override the next retry interval, pass `next_retry_delay` when raising [ApplicationError](/references/failures#application-failure) in an Activity:
 
--- a/docs/develop/python/failure-detection.mdx
+++ b/docs/develop/python/failure-detection.mdx
@@ -179,6 +179,22 @@ from temporalio.common import RetryPolicy
         )
 ```
 
+### Override the retry interval with Next Retry Delay {#next-retry-delay}
+
+To override the next retry interval, pass `next_retry_delay` when raising [ApplicationError](/references/failures#application-failure) in an Activity:
+
+```python
+from temporalio.exceptions import ApplicationError
+
+@activity.defn
+async def my_activity():
+    ...
+    raise ApplicationError(
+        "Something bad happened",
+        next_retry_delay=timedelta(seconds=3),
+    )
+```
+
 ## Heartbeat an Activity {#activity-heartbeats}
 
 **How to Heartbeat an Activity using the Temporal Python SDK.**


docs/develop/python/failure-detection.mdx
```diff
@@ -179,6 +179,22 @@ from temporalio.common import RetryPolicy
         )
 ```
 
+### Override the retry interval with Next Retry Delay {#next-retry-delay}
```
fairlydurable, line 182:
```suggestion
### Override the retry interval with `next_retry_delay` {#next-retry-delay}
```

docs/develop/python/failure-detection.mdx
```diff
@@ -179,6 +179,22 @@ from temporalio.common import RetryPolicy
...
+
+To override the next retry interval, pass `next_retry_delay` when raising [ApplicationError](/references/failures#application-failure) in an Activity:
```
fairlydurable, line 184:
```suggestion
To override the next retry interval, set the `next_retry_delay` field to a time duration when raising [ApplicationError](/references/failures#application-failure) in an Activity:
```

fairlydurable: Would benefit from some explanation of the circumstances under which you'd want to override this delay. 

Could you replace something bad happened" with anything more contextualized and meaningful so the example shows how and why the feature is useful?

fairlydurable (changes requested):
Minor questions, asking for context

docs/develop/python/failure-detection.mdx
dandavison: The code sample shows a `timedelta` being used; I don't see any need to expand the text in that regard. (Our documentation is for people who can read code.)

I considered whether to use a less generic error message but decided to follow our Java docs and keep it simple. My view is that this feature is for users who _already_ have the desire to do this for their own reasons, and the purpose of the documentation is to tell them how to do it. I think that we should hesitate before padding accompanying text beyond the minimum necessary there is already a lot of prose in the docs.

docs/develop/python/failure-detection.mdx
dandavison: Changed

docs/develop/python/failure-detection.mdx
fairlydurable: Thanks for sharing your thoughts. While I understand your desire to keep things concise, context enhances documentation value, even for experienced users. 

- Expanding line 184 provides greater accuracy about the expected value type, in alignment with documentation standards.
- Adding motivation to the introduction clarifies why someone might want to adjust the next_retry_delay for specific scenarios like managing retries based on external service availability. 
- In code, vague phrases like "something bad happened" lacks specificity. It doesn't convey the sense of an actual situation users might encounter. Adding that context respects the reader, makes the sample more concrete and relatable, and prioritizes clarity.

dandavison:
I think it would be more efficient if the docs team produces the PR (you can use this one if you like). I've added the initial suggested content from this PR to https://temporalio.atlassian.net/browse/EDU-3194.

PR #3107, opened 2024-09-23
Workflow-init support docs (Typescript)

antlai-temporal:
## What does this PR do?

This PR clarifies how to safely register update and signal handlers after needed workflow initialization (TypeScript).

This closes https://github.com/temporalio/sdk-typescript/issues/1483

### DIFF
--- a/docs/develop/typescript/message-passing.mdx
+++ b/docs/develop/typescript/message-passing.mdx
@@ -195,6 +195,7 @@ wf.setHandler(
   This includes the Update ID, which can be useful for deduplication when using Continue-As-New: see [Ensuring your messages are processed exactly once](/encyclopedia/workflow-message-passing#exactly-once-message-processing).
 - Update (and Signal) handlers can be `async`, letting them use Activities, Child Workflows, durable [`workflow.sleep`](https://typescript.temporal.io/api/namespaces/workflow#sleep) Timers, [`workflow.condition`](https://typescript.temporal.io/api/namespaces/workflow#condition) conditions, and more.
   See [Async handlers](#async-handlers) and [Workflow message passing](/encyclopedia/workflow-message-passing) for safe usage guidelines.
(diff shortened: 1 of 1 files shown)


docs/develop/typescript/message-passing.mdx
```diff
@@ -195,6 +195,7 @@ wf.setHandler(
...
 - Update (and Signal) handlers can be `async`, letting them use Activities, Child Workflows, durable [`workflow.sleep`](https://typescript.temporal.io/api/namespaces/workflow#sleep) Timers, [`workflow.condition`](https://typescript.temporal.io/api/namespaces/workflow#condition) conditions, and more.
   See [Async handlers](#async-handlers) and [Workflow message passing](/encyclopedia/workflow-message-passing) for safe usage guidelines.
+- Delay calling [`workflow.setHandler`](https://typescript.temporal.io/api/namespaces/workflow#sethandler) until the workflow initialization needed by Update (or Signal) handlers has finished. This is safe because the SDK buffers messages when there are no registered handlers for them. Note that [`workflow.setHandler`](https://typescript.temporal.io/api/namespaces/workflow#sethandler) will immediately invoke the handler of buffered Updates (or Signals) with matching types, and this could lead to out-of-order processing of messages with different types.
```
dandavison, line 198: I agree that the message-passing docs are an appropriate location for this. (The feature is only there for message handling reasons, and the Workflow Definition docs don't obviously to have an appropriate point for an "advanced usage" bullet point like this.)

However, since this feature applies to Update and Signal I wonder whether it should be placed in the "Message handler patterns" section, rather than in the Update section.

antlai-temporal: If you see in the previous bullet it already talks about "Update (and Signal) handlers can be...", and most of the discussion in that bigger paragraph on `async` belongs to both.
 I don't think it belongs in a "Message handler patterns" this is a bit more general than specific patterns. I agree that maybe we want to have a foundation section for both Update and Signal, but this PR was not trying to do that...
 Maybe restructure the whole section in a different PR?

dandavison: > If you see in the previous bullet it already talks about "Update (and Signal) handlers can be...", and most of the discussion in that bigger paragraph on async belongs to both.

Actually, that is true of both the signal and update handler sections: they mirror each other and contain some duplicated content. So IMO the options are either to duplicate this across both, or move it to the "message handler patterns", which is where Python discusses `@workflow.init`. But I think you're probably right that this should stay here; it's just that it needs to be duplicated for signal handlers IMO.

dandavison:
Superseded by https://github.com/temporalio/documentation/pull/3133

Issue #3111, opened 2024-09-26
[ Documentation request ] Python: Pytest fixtures and WorkflowEnvironment


edmondop:
# Brief description

I had encounter a problem when reusing a global WorkflowEnvironment defined as a pytest fixture resulted in time-skipping not working. I followed the common practice of define a `@pytest.fixture(scope="session")` which invoked `WorkflowEnvironment.start_time_skipping` but time-skipping wasn't happening, i.e. calls to `env.sleep(x)` were effectively behaving as blocking call.  I debugged and realized the source of the blocking was https://github.com/temporalio/sdk-python/blob/a1b5d653fa712d35194e275295c5ec9c3721ecc9/temporalio/testing/_workflow.py#L424

#  Your recommended content

I naively adopted a well-known pattern (using pytest fixtures with scope="session" for test infrastructure) that lead me in the wrong direction. I think we should make clear in the documentation that tests should be written like so:

```
def test1():
     with WorkflowEnvironment.start_time_skipping() as env:
              pass

def test2():
     with WorkflowEnvironment.start_time_skipping() as env:
              pass
```

and the idea of caching the workflow environment, although natural, can lead to a bad outcome.

I am happy to contribute this back to the docs and maybe a sample

PR #3113, opened 2024-09-28
Python: Added paragraph that warns about pytest.fixture(scope="session") with test server

edmondop:
## What does this PR do?

Addresses #3111 . When running tests in a project I noticed:
- Time skipping wasn't working even if it was enabled
- samples-python test never reuse the same workflow environment with timeskipping

This additional paragraph should warn developers familiar with pytest.fixtures not to use the scope session for caching the workflow environment

### DIFF
--- a/docs/develop/python/testing-suite.mdx
+++ b/docs/develop/python/testing-suite.mdx
@@ -43,6 +43,34 @@ Some SDKs have support or examples for popular test frameworks, runners, or libr
 
 One recommended framework for testing in Python for the Temporal SDK is [pytest](https://docs.pytest.org/), which can help with fixtures to stand up and tear down test environments, provide useful test discovery, and make it easy to write parameterized tests.
 
+### A note on pytest fixtures
+
+When using pytest fixtures to setup databases, application servers and common infrastructure it is not unusual to use the `session` scope to ensure the fixture is shared among multiple tests. However, with the test server it is important that each test is executed in isolation. Launching a test server per test introduce negligible overhead and ensures your test do not conflict with each other
+
+
+```python
+import pytest
+import temporalio.testing.WorkflowEnvironment
+
+
+# This might break because it will reuse the workflow environment 
+pytest.fixture(scope="session")
+def workflow_environment():
+  return WorkflowEnvironment.start_time_skipping()
+
+def test_1(workflow_environment):
+  pass
+
+def test_2(workflow_environment):
+  pass
+
+# Correct: if you don't specify a scope, the default scope is function,
+# so a new workflow environment will be created for each test
+@pytest.fixture 
+def workflow_environment():
+  return WorkflowEnvironment.start_time_skipping()
+```
+
 ## Testing Activities {#test-activities}
 
 An Activity can be tested with a mock Activity environment, which provides a way to mock the Activity context, listen to Heartbeats, and cancel the Activity.


fairlydurable:
Created [SDK-2941](https://temporalio.atlassian.net/browse/SDK-2941) to review

docs/develop/python/testing-suite.mdx
```diff
@@ -36,6 +36,34 @@ Some SDKs have support or examples for popular test frameworks, runners, or libr
...
+  pass
+
+def test_2(workflow_environment):
+  pass
+
+# Correct: if you don't specify a scope, the default scope is function,
+# so a new workflow environment will be created for each test
+@pytest.fixture 
+def workflow_environment():
+  return WorkflowEnvironment.start_time_skipping()
```
cretz, line 64: It is a bit strange to have the primary code be the wrong code and have the same code correctly below (which you don't usually redefine the same function in the same code block). Maybe something like:

```suggestion
# Notice how this doesn't specify a "scope" like "session" scope. If you don't
# specify a scope, the default scope is function, so a new workflow
# environment will be created for each test.
@pytest.fixture
def workflow_environment():
  return WorkflowEnvironment.start_time_skipping()

def test_1(workflow_environment):
  pass

def test_2(workflow_environment):
  pass
```

Or if you must have the incorrect way be the in the initial block, can you move the correct way into its own separate block? It's a bit confusing to see the same method defined twice in the same block.

docs/develop/python/testing-suite.mdx
```diff
@@ -36,6 +36,34 @@ Some SDKs have support or examples for popular test frameworks, runners, or libr
...
+When using pytest fixtures to setup databases, application servers and common infrastructure it is not unusual to use the `session` scope to ensure the fixture is shared among multiple tests. However, with the test server it is important that each test is executed in isolation. Launching a test server per test introduce negligible overhead and ensures your test do not conflict with each other
+
+
+```python
+import pytest
+import temporalio.testing.WorkflowEnvironment
+
+
+# This might break because it will reuse the workflow environment 
+pytest.fixture(scope="session")
```
cretz, line 50:
```suggestion
@pytest.fixture(scope="session")
```

If you do keep this, need `@`

docs/develop/python/testing-suite.mdx
```diff
@@ -36,6 +36,34 @@ Some SDKs have support or examples for popular test frameworks, runners, or libr
 
 One recommended framework for testing in Python for the Temporal SDK is [pytest](https://docs.pytest.org/), which can help with fixtures to stand up and tear down test environments, provide useful test discovery, and make it easy to write parameterized tests.
 
+### A note on pytest fixtures
+
+When using pytest fixtures to setup databases, application servers and common infrastructure it is not unusual to use the `session` scope to ensure the fixture is shared among multiple tests. However, with the test server it is important that each test is executed in isolation. Launching a test server per test introduce negligible overhead and ensures your test do not conflict with each other
```
cretz, line 41:
```suggestion
When using pytest fixtures to setup databases, application servers and common infrastructure it is not unusual to use the `session` scope to ensure the fixture is shared among multiple tests. However, with the time-skipping test server it is important that each test is executed in isolation. Launching a test server per test introduce negligible overhead and ensures your test do not conflict with each other
```

Notice, this only applies to the time skipping test server, not the local test server

fairlydurable:
Thanks, Chad. Back to you @edmondop

docs/develop/python/testing-suite.mdx
edmondop: Thanks for catching this!

docs/develop/python/testing-suite.mdx
```diff
@@ -36,6 +36,30 @@ Some SDKs have support or examples for popular test frameworks, runners, or libr
 
 One recommended framework for testing in Python for the Temporal SDK is [pytest](https://docs.pytest.org/), which can help with fixtures to stand up and tear down test environments, provide useful test discovery, and make it easy to write parameterized tests.
 
+### A note on pytest fixtures
+
+When using pytest fixtures to setup databases, application servers and common infrastructure it is not unusual to use the `session` scope to ensure the fixture is shared among multiple tests. However, with the time-skipping test server it is important that each test is executed in isolation. Launching a test server per test introduce negligible overhead and ensures your test do not conflict with each other
```
cretz, line 41:
```suggestion
When using pytest fixtures to setup databases, application servers and common infrastructure it is not unusual to use the `session` scope to ensure the fixture is shared among multiple tests. However, with the time-skipping test server it is important that each test is executed in isolation. Launching a test server per test introduce negligible overhead and ensures your test do not conflict with each other.
```

docs/develop/python/testing-suite.mdx
```diff
@@ -36,6 +36,30 @@ Some SDKs have support or examples for popular test frameworks, runners, or libr
...
+import pytest
+import temporalio.testing.WorkflowEnvironment
+
+
+# Notice how this doesn't specify a "scope" like "session" scope. If you don't
+# specify a scope, the default scope is function, so a new workflow
+# environment will be created for each test.
+@pytest.fixture
+def workflow_environment():
+  return WorkflowEnvironment.start_time_skipping()
```
cretz, line 54: I am not sure this works without `@pytest_asyncio.fixture` and `async` and `await`. Have you tested this code? And for many, they may prefer a "session" scoped `start_local` instead of time skipping even though this only demonstrates time skipping.

edmondop: That's a good point, was using the auto-mode in my tests https://pytest-asyncio.readthedocs.io/en/latest/reference/configuration.html  but I agree that being explicit works better

fairlydurable:
@edmondop Thanks for all your work on this. Please ping me and let me know when it is ready to move forward. Cheers!

cretz:
@fairlydurable - sorry I missed your request. I think https://github.com/temporalio/documentation/pull/3113/files#r1866332621 still needs to be addressed

docs/develop/python/testing-suite.mdx
edmondop: Should be addressed on @cretz

cretz: It appears the `workflow_environment()` function is now duplicated in this snippet

edmondop: It is to show the good and bad example, do you think one of the two fixtures should be named differently ?

cretz: Yeah, I think it's a bit confusing to even have a bad example unless it's labeled more clearly and in a separate snippet. Many of our users just grab the first snippet and move on.

So in this case, we do want `start_local` to be used by users and used across tests, but if they need `start_time_skipping` you are correct that it should be per test. So maybe we should have both forms, named differently, with comments above each on why start-local is session scoped and start-time-skipping is not.

PR #3119, opened 2024-10-01
EDU-3240: Change "Login" to "Log In" (verb) in API Keys docs

tomwheeler:
## What does this PR do?

Corrects all cases in which "Login" was used as a verb on this page, replacing them with "Log in."


fairlydurable (approved):
Thank you for improving this

PR #3121, opened 2024-10-02
docs onboarding - contributing guidance 

jsundai:

### DIFF
--- a/docs-onboarding/IA-guidance.md
+++ b/docs-onboarding/IA-guidance.md
@@ -1,1 +1,0 @@
-# IA Guidance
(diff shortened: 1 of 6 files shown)


tomwheeler (changes requested):
My one request would be to rename `IA-guidance.md` as `ia-guidance.md` so that it's consistent with the other file names. 

I doublechecked just now, and every other Markdown file in the `documentation` repo uses lowercase letters exclusively. Since UNIX systems tend to have case-sensitive filesystem, we should avoid arbitrarily mixing case in filenames because it can create trouble down the road.

PR #3129, opened 2024-10-07
add encode endpoint sample

axfelix:
One thing I'm not sure about here is whether showing the POST values as non-base64 encoded is potentially misleading (I've seen people get confused about this aspect of our codec server reference before) -- but if I *do* show the input as base64 encoded, it's *more* confusing since it's indistinguishable from encrypted strings. Made a small change to the "Request body" section of this page simultaneously to try to rectify that.


fairlydurable:
@axfelix Is there a ticket that tracks this PR? Thanks

axfelix:
https://temporalio.atlassian.net/browse/EDU-3231

brianmacdonald-temporal (approved):
Looks good. No editorial changes.

PR #3141, opened 2024-10-09
update Nexus data converter encryption options

prasek:
## What does this PR do?

Makes it clear that Nexus does support callers and handlers using different encryption keys today, if you flow the KMS encryption key id in the payload metadata and both caller and handler have access to those KMS keys.

## Notes to reviewers

Reference example: https://github.com/temporalio/reference-app-orders-go/blob/main/app/temporalutil/data_converter.go

### DIFF
--- a/docs/encyclopedia/temporal-nexus-security.mdx
+++ b/docs/encyclopedia/temporal-nexus-security.mdx
@@ -60,4 +60,18 @@ In Temporal Cloud multiple security provisions are in place to ensure it can act
 ### Payload Encryption and Data Converter {#payload-encryption-data-converter}
 
 The Data Converter works the same for a Nexus Operation as it does for other payloads sent between a Worker and Temporal Cloud.
-There is support for a single Data Converter to be used for inbound inputs and outbound outputs, which means the encryption key often used in a custom Data Converter must be the same for both the caller and the handler.
+The caller and handler Workers must have compatible Data Converters as operation inputs and results are passed between the two.
+
+If encryption keys are used to encrypt payloads, they must be available in both the caller and handler. This may be achieved by either:
+- Using the same encryption key in both the caller and handler, for example a shared symmetric key stored in your KMS.
+- Using different encryption keys and passing the KMS encryption key ID via payload metadata.
+  - Operation input payload:
+    - Caller encrypts the input payload and puts the caller's KMS encryption key ID in payload metadata.
+    - Handler decrypts the input payload using the KMS encryption key ID in payload metadata.
+  - Operation result payload:
+    - Handler encrypts the result payload and puts the handler's KMS encryption key ID in payload metadata.
+    - Caller decrypts the result payload using the KMS encryption key ID in payload metadata.
+  - Codec server can decrypt payloads using the KMS encryption key ID in payload metadata.
+  - Nexus Workers in different regions may require a multi-region KMS.
+
+Please let us know if you need per-Endpoint or per-Service payload encryption or better handling for asymmetric keys.
--- a/docs/encyclopedia/temporal-nexus-security.mdx
+++ b/docs/encyclopedia/temporal-nexus-security.mdx
@@ -62,16 +62,7 @@ In Temporal Cloud multiple security provisions are in place to ensure it can act
 The Data Converter works the same for a Nexus Operation as it does for other payloads sent between a Worker and Temporal Cloud.
 The caller and handler Workers must have compatible Data Converters as operation inputs and results are passed between the two.
 
-If encryption keys are used to encrypt payloads, they must be available in both the caller and handler. This may be achieved by either:
-- Using the same encryption key in both the caller and handler, for example a shared symmetric key stored in your KMS.
-- Using different encryption keys and passing the KMS encryption key ID via payload metadata.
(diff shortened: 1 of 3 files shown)


docs/encyclopedia/temporal-nexus-security.mdx
```diff
@@ -60,4 +60,18 @@ In Temporal Cloud multiple security provisions are in place to ensure it can act
 ### Payload Encryption and Data Converter {#payload-encryption-data-converter}
 
 The Data Converter works the same for a Nexus Operation as it does for other payloads sent between a Worker and Temporal Cloud.
-There is support for a single Data Converter to be used for inbound inputs and outbound outputs, which means the encryption key often used in a custom Data Converter must be the same for both the caller and the handler.
+The caller and handler Workers must have compatible Data Converters as operation inputs and results are passed between the two.
+
+If encryption keys are used to encrypt payloads, they must be available in both the caller and handler. This may be achieved by either:
+- Using the same encryption key in both the caller and handler, for example a shared symmetric key stored in your KMS.
+- Using different encryption keys and passing the KMS encryption key ID via payload metadata.
```
bergundy, line 67: Isn't this just using the same encryption keys on both sides? I would remove this bullet point until we have a better story around per service payload converters.

docs/production-deployment/cloud/nexus/operations.mdx
```diff
@@ -201,7 +201,10 @@ See [Nexus Secure Routing](/nexus/security#secure-routing) for additional detail
 ## Payload Encryption
 
 For payload encryption, the DataConverter works the same for a Nexus Operation as it does for other payloads sent between a Worker and Temporal Cloud.
-Currently there is support for a single DataConverter to be used for inbound inputs and outbound outputs, which means the encryption key often used in a custom DataConverter must be the same for both the caller and the handler.
-Please let us know if you need per-Endpoint or per-Service payload encryption.
+The caller and handler Workers must have compatible Data Converters as operation inputs and results are passed between the two.
```
bergundy, line 204: Can we avoid duplicating the same point three times?

prasek: The encyclopedia page has the full info. We do refer these other sections (evaluate, deploy to prod) to the encyclopedia entry, but keep a summary in these other section which IMO is useful to make folks aware, without getting into the details -- which is what the encyclopedia entry is for.

For new users, would expect them to read the evaluate section or deploy to prod sections first, so having a little context/summary there is goodness IMO to build awareness.

docs/encyclopedia/temporal-nexus-security.mdx
prasek: If you're adding a Nexus service/handler to an existing worker, you may have an existing encryption key for your data converter. The caller workflow may have a different encryption key in their worker. If they must encrypt with the same key, then something has to change, or just walk away since it doesn't seem like a fit right now.

But you don't have to encrypt payloads with the same encryption keys on both sides, the caller and handler can keep using whatever existing keys they have, and just flow the KMS key id and if the other side has permissions to get that key, they can decrypt the payload and it just works.

We can definitely update when we get a better story around per service payload converters, but think what we have right now will work for many folks, esp. intra org -- so want to make sure folks are not dissuaded from using Nexus due to a perceived limitation.

docs/production-deployment/cloud/nexus/operations.mdx
bergundy: I would rather not have so much duplication if possible. It gets hard to maintain and when a user browses these docs pages, they'll get tired from seeing the same content appear everywhere.

docs/encyclopedia/temporal-nexus-security.mdx
bergundy: Yes, I just think that this comment is more confusing than helpful.
I would keep it at for now:

```
Using the same encryption key in both the caller and handler, for example a shared symmetric key stored in your KMS.
```

docs/production-deployment/cloud/nexus/operations.mdx
jsundai: I do agree with the duplication since it's the exact same on the security page. Could this sentence just be rephrased?

docs/encyclopedia/temporal-nexus-security.mdx
jsundai, line 67: Small nit - we typically try to stay away from Latin words like via. I'd replace via with "through".

docs/production-deployment/cloud/nexus/operations.mdx
```diff
@@ -201,7 +201,10 @@ See [Nexus Secure Routing](/nexus/security#secure-routing) for additional detail
...
+
+If encryption keys are used to encrypt payloads, they must be available in both the caller and handler. This may be achieved by either:
+- Using the same encryption key in both the caller and handler, for example a shared symmetric key stored in your KMS.
+- Using different encryption keys and passing the KMS encryption key ID via payload metadata.
```
jsundai, line 208: Other than those small things, LGTM!

prasek:
@jsundai all feedback has been addressed, think we're good to go!

PR #3146, opened 2024-10-10
yarn format catchup commit

axfelix:
This is catching up on yarn format after not running it in a while. I ran a `git diff` on a few of the files that were automatically changed by this, and *it looks preliminarily good to me* — it did some dedenting (this is *very* good as it means we might not need to rely on snipsync for this if we continue using dprint for docs/learn instead), some fixing of unnecessary multiple spaces, and some cleanup of nested lists. If we like this, we probably want it to run Github-side on all commits from now on, so no one is caught out by it.


jsundai:
Looks good! Thanks for doing this! 🙌 I would just check the previews of the code samples before merging.

axfelix:
Yeah, I think this is good to merge! Looked over all of it and it's making all good changes.

jsundai (approved):

PR #3151, opened 2024-10-12
tcld nexus docs

prasek:
## What does this PR do?

Adds `tcld nexus` docs

## Notes to reviewers

Auto generated from https://github.com/prasek/tcld/tree/nexus-tcld-docs

Builds on previous docs gen enhancements: https://github.com/prasek/temporal-cli/tree/nexus-docs

### DIFF
--- a/docs/production-deployment/cloud/tcld/nexus.mdx
+++ b/docs/production-deployment/cloud/tcld/nexus.mdx
@@ -25,12 +25,125 @@ The `tcld nexus endpoint` commands manage Nexus Endpoints in Temporal Cloud.
 
 Alias: `ep`
 
+- [tcld nexus endpoint allowed-namespace](#allowed-namespace)
 - [tcld nexus endpoint create](#create)
 - [tcld nexus endpoint delete](#delete)
 - [tcld nexus endpoint get](#get)
 - [tcld nexus endpoint list](#list)
 - [tcld nexus endpoint update](#update)
-- [tcld nexus endpoint allowed-namespace](#allowed-namespace)
+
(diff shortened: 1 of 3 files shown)


docs/production-deployment/cloud/tcld/nexus.mdx
```diff
@@ -0,0 +1,297 @@
...
+
+Alias: `ttq`
+
+#### --unset-description
+
+Unset endpoint description.
+
+### allowed-namespace
+
+These commmands manage the allowed namespaces for a Nexus Endpoint.
```
fairlydurable, line 187:
```suggestion
These commands manage the allowed namespaces for a Nexus Endpoint.
```

fairlydurable (changes requested):
One typo

prasek:
thanks @fairlydurable, addressed comments and should be good to go!

fairlydurable (approved):
All files read. This material aligns with existing CLI doc form.

I don't want to get in the way of the ticket. If this needs to be out there, there are no issues that would prevent this going live. I am approving this ticket.

I would strongly recommend a second pass on a follow-up ticket:

- The tags are not in alignment with our new SEO optimization (cc @napcs)
- The layout of the page can be substantially improved for readability. We can sync and I'd be happy to mock up suggestions.
- If this material is auto-generated and should not be directly edited (and instead go through the tcld team), that should be noted loudly in the file, whether with comments, metadata, or other non-user-facing element.

Also cc'ing in @josh-berry and @yuandrew since this feedback will likely affect Temporal CLI doc generation.

PR #3155, opened 2024-10-14
fix: merge for typescript failure detection page

jsundai:


angelazhou32 (approved):
TY for taking care of this!

PR #3171, opened 2024-10-22
EDU-3257 - Knock out one stray "metric" that was missed

fairlydurable:
Regardless of their utility for evaluating Worker capacity, these values that are being produced are not Temporal metrics and cannot be read and processed as those would be.

Updated wording so they refer to "values" and not metrics.

Please note, one anchor point retains "metric" in its name, but needs to either not change or to redirect. This can be addressed in a future ticket.


fairlydurable:
cc @atihkin for heads up.

napcs (approved):

PR #3182, opened 2024-10-29
EDU-3373: Updates Namespace authentication from Pre-release to Public Preview

fairlydurable:
Source of Truth: Peter Sullivan


tomwheeler (approved):
Verified that the new link has the correct text and target

pvsone (approved):
LGTM

PR #3188, opened 2024-11-01
EDU-3585: Add references to Nexus Operations on Java testing page

pdoerner:
## What does this PR do?
Added references to Nexus Operations on the Java testing page.
Removed section for time skipping in activities because it is not supported.

## Notes to reviewers
There is no separate section for `Testing Nexus Operations` like there is for Activities and Workflows because we do not currently have a separate mocked environment for Nexus Operations. They are only supported in the Java SDK within the context of a workflow.

I used `Operation` because that's what we usually refer to when taking about Nexus but I know that the Java SDK APIs also reference Nexus Service implementations, so LMK if `Service` would be better here.


docs/develop/java/testing-suite.mdx
```diff
@@ -301,7 +308,7 @@ Learn to set up the time-skipping test framework in the SDK of your choice.
 
 You can skip time automatically in the SDK of your choice.
 Start a test server process that skips time as needed.
-For example, in the time-skipping mode, Timers, which include sleeps and conditional timeouts, are fast-forwarded except when Activities are running.
+For example, in the time-skipping mode, Timers, which include sleeps and conditional timeouts, are fast-forwarded except when Activities or Nexus Operations are running.
```
bergundy, line 311: I would adjust a bit and add a caveat that nexus operation handlers timeout after 10 seconds.

bergundy (approved):

docs/develop/java/testing-suite.mdx
```diff
@@ -280,6 +281,12 @@ Mock the Activity invocation when unit testing your Workflows.
 
 When integration testing Workflows with a Worker, you can mock Activities by providing mock Activity implementations to the Worker.
 
+### How to mock Nexus Operations {#mock-nexus-operations}
+
+Mock the Nexus Operation invocation when unit testing your Workflows.
+
+When integration testing Workflows with a Worker, you can mock Nexus Operations by providing mock Nexus Service implementations to the Worker.
```
Quinn-With-Two-Ns, line 288:
```suggestion
When integration testing Workflows with a Worker, you can mock Nexus Operations by providing a mock Nexus Service implementations to the Worker.
```

docs/develop/java/testing-suite.mdx
```diff
@@ -311,6 +318,10 @@ Learn to skip time manually in the SDK of your choice.
 
 Learn to skip time in Activities in the SDK of your choice.
 
+#### Skip time in Nexus Operations {#skip-time-in-nexus-operations}
+
+Learn to skip time in Nexus Operations in the SDK of your choice.
```
Quinn-With-Two-Ns, line 323: We don't support/can't support time skipping in nexus operations. I know this comment is what we say for activities, but it doesn't really make sense there either.

Quinn-With-Two-Ns (approved):

docs/develop/java/testing-suite.mdx
pdoerner: To make the sentence singular it would read `you can mock a Nexus Operation by providing a mock Nexus Service implementation to the worker` but I don't like that phrasing as much because it doesn't match what we have for activities.

docs/develop/java/testing-suite.mdx
```diff
@@ -25,12 +26,12 @@ The Testing section of the Temporal Application development guide describes the
 
 In the context of Temporal, you can create these types of automated tests:
 
-- **End-to-end:** Running a Temporal Server and Worker with all its Workflows and Activities; starting and interacting with Workflows from a Client.
+- **End-to-end:** Running a Temporal Server and Worker with all its Workflows, Activities and Nexus Operations; starting and interacting with Workflows from a Client.
```
fairlydurable, line 29:
```suggestion
- **End-to-end:** Running a Temporal Server and Worker with all its Workflows, Activities, and Nexus Operations; starting and interacting with Workflows from a Client.
```

fairlydurable (approved):
Looks good. Added a couple of Oxford commas.

@prasek signed off on technical accuracy.

Unblocking so it can move forward. Thank you for working on this ticket.

PR #3209, opened 2024-11-14
EDU-3543: [PHP] Document next retry delay

roxblnfk:
## What does this PR do?

Document Activity next Retry delay block
Update references page

Closes https://github.com/temporalio/sdk-php/issues/485


docs/develop/php/failure-detection.mdx
```diff
@@ -130,6 +130,26 @@ Activity Execution semantics rely on several parameters.
 The only required value that needs to be set is either a [Schedule-To-Close Timeout](/encyclopedia/detecting-activity-failures#start-to-close-timeout) or a [Start-To-Close Timeout](/encyclopedia/detecting-activity-failures#start-to-close-timeout).
 These values are set in the Activity Options.
 
+## Activity next Retry delay {#activity-next-retry-delay}
+
+**How to override the next Retry delay following an Activity failure using the Temporal PHP SDK**
+
+You may throw an [Application Failure](/references/failures#application-failure) with the NextRetryDelay field set.
```
fairlydurable, line 137:
```suggestion
You may throw an [Application Failure](/references/failures#application-failure) with the `NextRetryDelay` field set.
```

roxblnfk: Should I do the same for Java SDK?

fairlydurable: That would be great. 

I'm still trying to figure out if Application Failure needs to be lowercased ("application failure") or (better) become more specific to the language. It's not on my Temporal special upcase terms list. So I think you may want to use:

```
`ApplicationFailure`
```

for Java. I don't see anything listed for PHP on the application failure page (and, yes, it shouldn't be in title case because we use sentence case, but there are a _lot_ of things we're still developing standards for).  More specifically, I think using the exception name would be better, and code-based terms use backticked code face.

brianmacdonald-temporal (approved):
This looks good to me. Thanks, Aleksi!

PR #3210, opened 2024-11-15
gcp: service availability: remove Sydney, add Mumbai

XaF:
## What does this PR do?

We have been pulling Sydney for now because of the low usage of the region, and Mumbai got added because we reached the right level of commitment to open the region. As we are finalizing the region being available, I'm updating the documentation about our service regions!

## Notes to reviewers

Please review but wait for final confirmation (go-live of the region) to merge :)


MasonEgger (approved):
LGTM. Sorry for the delay, I didn't see this review request come through.

PR #3217, opened 2024-11-21
EDU-3579: Fix a mistake in message-passing docs

dandavison:
Fix two places where we incorrectly implied that a handle would be returned when the update is rejected.


MasonEgger (approved):
LGTM

PR #3218, opened 2024-11-22
small seo tweak to activities page

napcs:
This is a small fix to the Activities page that I'm also using to test an integration.


fairlydurable (approved):
Nice update. Thank you for pushing on this.

PR #3232, opened 2024-12-09
EDU-3361: Align Python and Go with .NET Workflow Init

fairlydurable:
- Brings feature explanation to the first line.
- Removes parentheticals.
- Minor adjustments.

### DIFF
--- a/docs/develop/java/message-passing.mdx
+++ b/docs/develop/java/message-passing.mdx
@@ -516,11 +516,11 @@ See [Finishing handlers before the Workflow completes](/encyclopedia/workflow-me
 
 ### Use `@WorkflowInit` to operate on Workflow input before any handler executes
 
-Normally, your Workflows constructor won't have any parameters.
-However, if you use the `@WorkflowInit` annotation on your constructor, you can give it the same [Workflow parameters](/develop/java/core-application#workflow-parameters) as your `@WorkflowMethod`.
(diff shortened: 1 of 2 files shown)


fairlydurable:
@dandavison @Quinn-With-Two-Ns Please check that no technical information was compromised with the .NET alignment. Thanks.

dandavison (changes requested):
I don't think this change is an improvement. In particular, this line

> The `@workflow.init` decorator method gives message handlers access to [Workflow input]

is more confusing than what exists currently. It gives the _constructor_ access; users already have a way to give message handlers access.

The point that we're trying to convey is

1. In Java/Python/.NET there's a way to make your _constructor_ receive the workflow input, in addition to the workflow run method.

2. This is useful because there's a non-obvious pitfall with Temporal: it's possible for a message handler to execute before the first line of your workflow has run.

fairlydurable:
@dandavison Would specifying 

> The @workflow.init decorator method gives message handler constructors/initializers access to [Workflow input]

work?

dandavison:
> The @workflow.init decorator method gives message handler constructors/initializers access to [Workflow input]

I find the current version of the Python and Java docs to make the point more clearly than variants of this PR. (Perhaps that means I'd support making .NET the same as them)

fairlydurable:
Closing the ticket. SDK does not have confidence in changes.

PR #3234, opened 2024-12-10
EDU-3690: Clarify that Activities and Child Workflows that exhaust their retries will cause Update / Workflow to fail.

dandavison:
Make it clear that an Activities and Child Workflows that exhaust their retries will cause an Update / Workflow to fail.


docs/encyclopedia/application-message-passing.mdx
```diff
@@ -342,16 +342,20 @@ The following content applies in every SDK except the Go SDK. See below.
 #### Exceptions in Signals
 
 In Signal handlers, throw [Application Failures](/references/failures#application-failure) only for unrecoverable errors, because the entire Workflow will fail.
+Similarly, allowing a failing Activity to exhaust its retries, so that it throws an [Activity Failure](https://docs.temporal.io/references/failures#activity-failure) will cause the entire Workflow to fail. This will only happen if you change the default Activity [Retry Policy](https://docs.temporal.io/encyclopedia/retry-policies).
```
drewhoskins-temporal: Presumably this occurs on a timeout as well?

The wording makes it ambiguous whether this is just an example or is intended to be comprehensive.  You might try:
"For example, Activities can cause unrecoverable errors if you change their settings from the default, such as when they time out or exhaust their retries."

drewhoskins-temporal (commented):
lgtm, one comment.

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -342,16 +342,21 @@ The following content applies in every SDK except the Go SDK. See below.
...
 
 #### Exceptions in Updates
 
-When you want to fail the Update and for the client to receive that error:
+Doing any of the following will fail the Update and cause the client to receive the error:
 
-- Reject the Update by throwing any exception from your Validator, OR
-- Throw an [Application Failure](/references/failures#application-failure) from your Update handler. Unlike with Signals, the Workflow will keep going.
+- Reject the Update by throwing any exception from your [Validator](https://docs.temporal.io/encyclopedia/workflow-message-passing#update-validators).
+- Allow a failing Activity of Child Workflow to exhaust its retries, so that it throws an [Activity Failure](https://docs.temporal.io/references/failures#activity-failure) or [Child Workflow Failure](https://docs.temporal.io/references/failures#child-workflow-failure). Note that for Activities, this will only happen if you change the default Activity [Retry Policy](https://docs.temporal.io/encyclopedia/retry-policies), since by default they retry forever.
```
cretz, line 354:
```suggestion
- Allow a failing Activity or Child Workflow to exhaust its retries, so that it throws an [Activity Failure](https://docs.temporal.io/references/failures#activity-failure) or [Child Workflow Failure](https://docs.temporal.io/references/failures#child-workflow-failure). Note that for Activities, this will only happen if you change the default Activity [Retry Policy](https://docs.temporal.io/encyclopedia/retry-policies), since by default they retry forever.
```

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -342,16 +342,21 @@ The following content applies in every SDK except the Go SDK. See below.
...
 
 In Signal handlers, throw [Application Failures](/references/failures#application-failure) only for unrecoverable errors, because the entire Workflow will fail.
+Similarly, allowing a failing Activity or Child Workflow to exhaust its retries, so that it throws an [Activity Failure](https://docs.temporal.io/references/failures#activity-failure) or [Child Workflow Failure](https://docs.temporal.io/references/failures#child-workflow-failure) will cause the entire Workflow to fail.
+Note that for Activities, this will only happen if you change the default Activity [Retry Policy](https://docs.temporal.io/encyclopedia/retry-policies), since by default they retry forever.
 If you throw any other exception, by default, it will cause a [Workflow Task Failure](/references/failures#workflow-task-failures). This means the Workflow will get stuck and will retry the handler periodically until the exception is fixed, for example by a code change.
 
 #### Exceptions in Updates
 
-When you want to fail the Update and for the client to receive that error:
+Doing any of the following will fail the Update and cause the client to receive the error:
```
cretz, line 351: This may be missing:

* Bubble failure to send signal to child/external workflow
* Bubble any error/exception/panic configured as a workflow failure (configurable in Go, Java, Python, and .NET in different ways)

docs/encyclopedia/application-message-passing.mdx
```diff
@@ -342,16 +342,21 @@ The following content applies in every SDK except the Go SDK. See below.
...
 
-- Reject the Update by throwing any exception from your Validator, OR
-- Throw an [Application Failure](/references/failures#application-failure) from your Update handler. Unlike with Signals, the Workflow will keep going.
+- Reject the Update by throwing any exception from your [Validator](https://docs.temporal.io/encyclopedia/workflow-message-passing#update-validators).
```
cretz, line 353: > throwing

Note, most of our SDKs don't have the concept of "throw"

> exceptions

Note, some languages call these errors.

Not sure if anything needs to change, but to a reader this can appear very Java/.NET centric.

cretz (approved):
Non-blocking notes

docs/encyclopedia/application-message-passing.mdx
dandavison: That's intentional. "Throw" is the term used in Java, Typescript, and .NET. Python and Ruby use "raise" which is idiosyncratic and in the minority; they'll understand "throw". Go has a separate section below due to the difficulty of describing error handling in one sentence in a way that applies across Go and exception-based languages.

docs/encyclopedia/application-message-passing.mdx
dandavison: I'm happy leaving those out here; the previous version wasn't exhaustive either. The main aim is to get readers to understand Temporal's high-level philosophy.

docs/encyclopedia/application-message-passing.mdx
dandavison: Activity timeout is just another way for a retry to be triggered, so I think it's OK as-is. Added Child Workflows, but neither now nor previously is there a claim to comprehensiveness.

dandavison:
>  Are there any blockers or target release dates we should be aware of for the merge?

No, this is just an incremental improvement to the docs.

docs/encyclopedia/application-message-passing.mdx
drewhoskins-temporal: I'm referring to the [schedule-to-close timeout](https://docs.temporal.io/encyclopedia/detecting-activity-failures#schedule-to-close-timeout), which doesn't result in a retry.
(also, schedule-to-start timeout, though that's rarely used)
I think the other part of my feedback applies, regardless.    "Similarly" isn't the right word since what you're describing is a particular example of the phenomenon you described in the first paragraph.

PR #3246, opened 2024-12-18
EDU-2392: Removing repetitive code snippet 

jsundai:


jsundai:
@brianmacdonald-temporal Ready for review

fairlydurable:
I don't see any problems but be aware that with at least one of these patches, there was a subtle but important difference between the two code snippets that seemed to be redundant at first and, in fact, needed clarification as to what each one did.

I'm cc'ing in @axfelix for a second pair of eyes

axfelix:
This one is *actually* redundant, though it's a bit unintuitive now, it should really segue directly into this part https://docs.temporal.io/develop/python/versioning#deprecated-patches to show an example of `deprecate_patch()`.

fairlydurable:
Thanks for checking @axfelix 

Going to approve and unblock.

fairlydurable (approved):
We're good to go. Unblocking.

PR #3254, opened 2024-12-19
EDU-2392: Removes duplicate code block from Python versioning page

dgorton:
Remove duplicate code block

Addresses https://temporalio.atlassian.net/browse/EDU-2392


dgorton:
@jsundai or @brianmacdonald-temporal - ready for review.

axfelix:
Actually not a duplicate! Note the difference between `pre_patch_activity` and `post_patch_activity`. However, I agree that this is slightly harder than necessary to parse... but our Python being fairly verbose makes this challenging to emphasize without having highlighting in our code samples.

In any case, I would reject this PR, but I encourage anyone to figure out a way to make this distinction clearer here.

fairlydurable:
@axfelix Understood. Will close the ticket as won't do.

Thank you @dgorton for your help here.

PR #3255, opened 2024-12-19
EDU-3637: Match account cancellation language in pricing.mdx

webchick:
## What does this PR do?
The current [pricing page](https://docs.temporal.io/cloud/pricing) directs users to contact Support if they wish to cancel their account. However, this is actually something SSU users can now do on their own.

There is already new language for this in the _new_ [pricing](https://github.com/temporalio/documentation/blob/main/docs/evaluate/temporal-cloud/pricing.mdx) page, which launches in January. But in the meantime, we are giving incorrect information.

This PR updates the "legacy" pricing page which is currently live to use the same language as the forthcoming pricing page.

Also closes https://temporalio.atlassian.net/browse/EDU-3637

### DIFF
--- a/docs/evaluate/temporal-cloud/legacy-pricing.mdx
+++ b/docs/evaluate/temporal-cloud/legacy-pricing.mdx
@@ -440,7 +440,8 @@ We charge applicable sales tax in US jurisdictions as required.
 
 **How do I cancel my account?**
 
-To cancel your Temporal Cloud account, [contact Support](https://temporalsupport.zendesk.com/).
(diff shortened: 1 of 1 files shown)


axfelix:
@jsundai @brianmacdonald-temporal Do we still need to merge this? Now we have "Pricing" and "Future Pricing" in the sidebar alongside "Support" and "Future Support," I assume we want to resolve that ASAP anyway.

fairlydurable:
I am going to close this. We had changes this morning on the billing and pricing page and the old legacy pricing is going away shortly. The information on the new (and now active) pricing page is correct.

PR #3258, opened 2024-12-20
EDU-3673: Update .NET Determinism update documentation to be inline with the README.md

cici:
Per ticket EDU-3673, update .NET Task Determinism documentation to be inline with the README.md at https://github.com/temporalio/sdk-dotnet/edit/main/README.md


cici:
Ready for review @Jwahir Sundai, @Brian Macdonald

brianmacdonald-temporal (approved):
Changes look good.

axfelix:
FYI -- I am not 100% on whether this best belongs in "Core application" or on a separate page going into detail about SDK async/determinism/sandbox behavior like we have for Python: https://docs.temporal.io/develop/python/python-sdk-sandbox

Not a huge deal either way, just want to flag it here, whichever decision is made.

fairlydurable:
I'm going to go ahead and merge but I wanted to note that as a rule we do not mention future features as we don't want to promise things that may not be delivered.

@axfelix This page is targeted for major updates anyway so I'm not worried about where the information belongs, since the intent is to break it down and repurpose it to other sections.

PR #3274, opened 2025-01-06
EDU-3819: docs: tried to clarify the notes on failures

GSmithApps:
## What does this PR do?

Tried to clarify some parts of the Failures page. Specifically regarding the Application Failures

### DIFF
--- a/docs/encyclopedia/application-message-passing.mdx
+++ b/docs/encyclopedia/application-message-passing.mdx
@@ -51,7 +51,7 @@ Unlike Signals, Updates must be synchronous. That is, they must wait for the Wor
 Use Signals instead of Updates when:
 
 - The Workflow's clients want to quickly move on after sending an asynchronous message.
-- The clients care more about the latency of sending the message than they do about the latency of the end-to-end operation.
+- The clients are willing to "fire and forget": they don't want a result or exception from the message.
 - The clients don't want to rely on the Worker being available.
 
 Use Updates instead of Signals when:
--- a/docs/production-deployment/self-hosted-guide/index.mdx
+++ b/docs/production-deployment/self-hosted-guide/index.mdx
@@ -27,7 +27,7 @@ If you are just getting started with Temporal, we recommend our [introductory tu
 
 :::
 
-:::info Buidling an app?
+:::info Building an app?
 
 If you are building a new Temporal Application, you might only need a [development server](/cli#start-dev-server) available through the [Temporal CLI](/cli).
 Check out the [dev guide](/develop) for application development best practices.


GSmithApps:
Hi @fairlydurable thanks for marking this as in-progress a few weeks ago. It is ready for review whenever time permits. Thank you!!

Side note: I see that vercel says the deployment failed -- is that just flaky CICD, or has something gone wrong? I don't think I've changed anything that would cause deployment to fail

Lastly, please let me know if I can help out at all - cheers!

fairlydurable:
Vercel is failing because of a broken anchor: linking to `/workers#activity-task-execution`

The new link is `/tasks#activity-task-execution` 

Hope that helps

docs/encyclopedia/retry-policies.mdx
```diff
@@ -18,7 +18,7 @@ import { RelatedReadContainer, RelatedReadItem } from '@site/src/components/rela
 A Retry Policy works in cooperation with the timeouts to provide fine controls to optimize the execution experience.
 
 A Retry Policy is a collection of attributes that instructs the Temporal Server how to retry a failure of a [Workflow Execution](/workflows#workflow-execution) or an [Activity Task Execution](/tasks#activity-task-execution).
-(Retry Policies do not apply to [Workflow Task Executions](/tasks#workflow-task-execution), which always retry indefinitely.)
+(Retry Policies do not apply to [Workflow Task Executions](/tasks#workflow-task-execution), which retry until the Workflow Execution Timeout, which is unlimited by default.)
```
fairlydurable, line 21:
```
Retry Policies do not apply to [Workflow Task Executions](/tasks#workflow-task-execution), which retry until the Workflow Execution Timeout.
The Execution Timeout for Workflow Task Executions is unlimited by default.
```

Won't they apply if the default is changed? What happens then? Can we stick in a "normally", get rid of the parentheses, and break this down into a couple of sentences?

```suggestion
Retry Policies normally do not apply to [Workflow Task Executions](/tasks#workflow-task-execution) unless their Workflow Execution Timeout is changed.
Workflow Task Executions will retry until reaching their Workflow Execution Timeout.
The default value for this timeout is unlimited, so a Workflow Task Execution will normally retry forever.
```

I'm not sure if I got the semantics correct here, so please confirm.

Also, does this need a tech reviewer to confirm your changes?

docs/references/failures.mdx
```diff
@@ -52,21 +52,22 @@ This is the only type of Temporal Failure created and thrown by user code.
...
-Most types of Temporal Failures occur automatically, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity Fails.
-You can also explicitly fail the Workflow Execution by throwing an Application Failure (returning any error in Go).
+Most types of Temporal Failures are raised by the Temporal Service, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity Fails.
+In contrast, the application developer can explicitly fail the Workflow Execution by throwing an Application Failure (returning any error in Go).
 
 #### Workflow Task Failures
 
 A **Workflow Task Failure** is an unexpected situation failing to process a Workflow Task.
-This could be triggered by raising an exception in your Workflow code.
+This could be triggered by a non-Temporal exception being raised (panicking in Go) in your Workflow code.
```
fairlydurable, line 61: Maybe "a non-Temporal exception being raised, or panicking in Go, in your Workflow code"

docs/references/failures.mdx
```diff
@@ -52,21 +52,22 @@ This is the only type of Temporal Failure created and thrown by user code.
...
 Any exception that does not extend Temporal's `FailureError` exception is considered a Workflow Task Failure.
-These types of failures will cause the Workflow Task to be retried.
+These types of failures will cause the Workflow Task to be retried until the
```
fairlydurable, line 63: I'd suggest rewording (assuming I got the logic right above). "These types of failures cause a Workflow Task to be retried indefinitely unless the Workflow Execution Timeout was changed from its default of "unlimited" time." Or something like that.

docs/references/failures.mdx
```diff
@@ -52,21 +52,22 @@ This is the only type of Temporal Failure created and thrown by user code.
...
+Workflow Execution Timeout, which is unlimited by default.
 
 #### Workflow Execution Failures
 
 An `ApplicationError`, an extension of `FailureError`, can be raised in a Workflow to fail the Workflow Execution.
 Workflow Execution Failures put the Workflow Execution into the "Failed" state and no more attempts will be made in progressing this execution.
-If you are creating custom exceptions you would either need to extend the `ApplicationError` class—a child class of `FailureError`— or explicitly state that this exception is a Workflow Execution Failure by raising a new `ApplicationError`.
+If you are creating custom exceptions you would need to extend the `ApplicationError` class—a child class of `FailureError`.
```
fairlydurable, line 70: How about "When you create custom exceptions, you need ot extend the `ApplicationError` class`. This class is a child of `FailureError`."

Would it help to link to the API references for these two?

docs/references/failures.mdx
```diff
@@ -82,7 +83,7 @@ During conversion, the following Application Failure fields are set:
 - `next_retry_delay` is left unset.
 - call stack is copied.
 
-When an [Activity Execution](/activities#activity-execution) fails, the Application Failure from the last Activity Task is the `cause` field of the [ActivityFailure](#activity-failure) thrown in the Workflow.
+When an [Activity Execution](/activities#activity-execution) (an Activity Execution is the full chain of [Activity Task Executions](/tasks#activity-task-execution)) fails, the Application Failure from the last Activity Task is the `cause` field of the [ActivityFailure](#activity-failure).  This Activity Failure is thrown by the Workflow's call to execute the Activity, and this call site is where Activity Failure can be handled by the application developer.
```
fairlydurable, line 86: This is very long and combines sentences on a single line. (Let me know if you want the copy-pasta of why we use our one sentence per line rules.) Suggest rewording to simplify the concepts because I found it confusing to read.

docs/references/failures.mdx
```diff
@@ -52,21 +52,22 @@ This is the only type of Temporal Failure created and thrown by user code.
 An error in a Workflow can cause either a **Workflow Task Failure** (the Task will be retried) or a **Workflow Execution Failure** (the Workflow is marked as failed).
 
 Only Workflow exceptions that are Temporal Failures cause the Workflow Execution to fail; all other exceptions cause the Workflow Task to fail and be retried (in Go, any error returned from the Workflow fails the Workflow Execution, and a panic fails the Workflow Task).
-Most types of Temporal Failures occur automatically, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity Fails.
-You can also explicitly fail the Workflow Execution by throwing an Application Failure (returning any error in Go).
+Most types of Temporal Failures are raised by the Temporal Service, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity Fails.
+In contrast, the application developer can explicitly fail the Workflow Execution by throwing an Application Failure (returning any error in Go).
```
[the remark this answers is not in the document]
GSmithApps: Agreed on switching "application developer" to "you"!

Yeah I agree that the parenthesis felt a little odd. I did it like this to mirror the description two sentences ago, because it's a similar situation in which Go behaves differently from how the other SDKs do. 

Lol I had the same question about what exactly constitutes an error in Go, and I'm pretty darn sure the precise behavior is that any non-nil value as the last return value will cause the failure, but I think every Go programmer would describe that as returning an error. It doesn't have to be any specific instance or type... I know, it's weird

docs/references/failures.mdx
```diff
@@ -52,21 +52,22 @@ This is the only type of Temporal Failure created and thrown by user code.
 An error in a Workflow can cause either a **Workflow Task Failure** (the Task will be retried) or a **Workflow Execution Failure** (the Workflow is marked as failed).
 
 Only Workflow exceptions that are Temporal Failures cause the Workflow Execution to fail; all other exceptions cause the Workflow Task to fail and be retried (in Go, any error returned from the Workflow fails the Workflow Execution, and a panic fails the Workflow Task).
-Most types of Temporal Failures occur automatically, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity Fails.
-You can also explicitly fail the Workflow Execution by throwing an Application Failure (returning any error in Go).
+Most types of Temporal Failures are raised by the Temporal Service, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity Fails.
```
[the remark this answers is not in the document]
GSmithApps: Agreed on not capitalizing the F! Still thinking on the other things

docs/references/failures.mdx
GSmithApps: Similar to above, I was following the existing pattern for how Go one-offs were called out. Happy to do whatever on these, but I do think it's worth trying to draw the reader's attention.

docs/references/failures.mdx
GSmithApps: I like your style on making the sentence flow more easily. On my read, I slightly prefer emphasizing that the retry will be until the Workflow Execution Timeout, and calling out the default almost as an afterthought rather than stating that they're retried indefinitely then including the other things as the afterthought. Idk though I think you're the writing expert

But yep the logic I see in your note matches my understanding!

docs/references/failures.mdx
GSmithApps: I like this idea! I think links are the way to go too. In this specific case, maybe it's a little tricky because this is a general article that isn't specific to any one language. And the most language-neutral description of these two things is above in this article. So we could do an internal link if we want. Or maybe we could bullet out all of the languages and link to their API docs? Idk I figure we won't do that. 

What do you think?

fairlydurable: Thank you for clarifying! Sounds to me like it's an Event History entry name. I see that it's code-faced  on the failures reference page, although I'm having "thoughts" about that unless it is a globally defined constant across SDKs.

Would these links work? 

* https://docs.temporal.io/references/failures#workflow-execution-failures
* https://docs.temporal.io/references/failures#application-failure

docs/references/failures.mdx
GSmithApps: Ohh I see the confusion -- this is talking about workflow **tasks**, not workflow executions, which are different. workflow executions are not retried, but workflow tasks are

fairlydurable: Oh, gotcha! Thanks

docs/references/failures.mdx
```diff
@@ -52,21 +52,22 @@ This is the only type of Temporal Failure created and thrown by user code.
 An error in a Workflow can cause either a **Workflow Task Failure** (the Task will be retried) or a **Workflow Execution Failure** (the Workflow is marked as failed).
 
 Only Workflow exceptions that are Temporal Failures cause the Workflow Execution to fail; all other exceptions cause the Workflow Task to fail and be retried (in Go, any error returned from the Workflow fails the Workflow Execution, and a panic fails the Workflow Task).
-Most types of Temporal Failures occur automatically, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity Fails.
-You can also explicitly fail the Workflow Execution by throwing an Application Failure (returning any error in Go).
+Most types of Temporal Failures are raised by the Temporal Service, like a [Cancelled Failure](#cancelled-failure) when the Workflow is Cancelled or an [Activity Failure](#activity-failure) when an Activity fails.
+In contrast, you can explicitly fail the Workflow Execution by throwing an Application Failure (returning any error in Go) in Workflow Definition code.
```
GSmithApps, line 56: I cleaned up some things in response to your comment, and that seems to have wiped out your comment, so I'm just putting it back in case we want to continue discussing :)

Erica:
Consider agreement between the articles for Workflow and Activity ("when a Workflow is Cancelled", "when an Activity fails").

Should F be capitalized here in "Fails"? I assume it is correct for "Cancelled Failure" and "Activity Failure", although Failure isn't in my Temporal caps list.

If you want to get rid of passive voice, you can move Temporal Service ("The Temporal Service raises most types of Temporal failures, like a...")

Grant:
Agreed on not capitalizing the F!

Regarding the argeement on articles, yeah same here -- I felt like something was clunky when I read it too. I couldn't quite tell what it was. But as far as the article agreement goes ("when a Workflow is Cancelled", "when an Activity fails"), I actually think this is intentional. From the reader/user's perspective, the cancellation is probably a manual intervention, so it will feel like it was cancelled, whereas with an Activity Failure, it wouldn't happen because of manual intervention, so that would feel more passive and more distant or out of their control. Does that align with how the prose reads to you? (Also, I didn't change that part, but I figure you're mentioning it because we should fix it while we're here, right?)

Still thinking on the passive voice thing. I don't think I really tried to change the voice on that -- I just expanded more, but maybe that changes the voice.

docs/references/failures.mdx
GSmithApps: Okay yeah I added them. I don't think it's perfect, but I think it's probably the best we can do given that creating custom exceptions is slightly SDK-specific and this is an SDK-agnostic article. I'm happy with it if you are!

docs/references/failures.mdx
fairlydurable: p.s. I'm going to with your gut on the articles because this isn't a hill to die on

docs/references/failures.mdx
GSmithApps: That copy-pasta would be awesome!! I'm learning about technical writing, so anything like that would be awesome. I went ahead and fixed that sentence-break

Yeah it's wordy and hard to read. I can work on it a little more if we want. Honestly, the original of how we had it was correct and fine, so if we want, we can go back to that too. I just expanded a little bit, but I may have been overzealous and it might not be a necessary expansion

docs/references/failures.mdx
GSmithApps: Lol yeah I am behind on m-dashes and modal auxiliaries. Not a clue! I could look them up though if you want a fellow discussion buddy :)

I agree on the question of what specifically a Canceled Failure is. I think the way I'd describe it is our language-neutral term for what in each SDK we'd call a CancelledError, CancelledFailure, or CancelledException. I left it not codefaced because Application Failure was also not code faced, and those two behave the same

docs/references/failures.mdx
fairlydurable: This is why we put items on separate lines, nominally sentences:

* Improves version control diffs for easier line-by-line inspection.
* Focuses feedback and review discussions on single points, preventing overlap with other content that may need changes.
* Reduces overlap and merge conflicts with other tickets.
* Enhances the precision of our automated lint tooling, such as Vale (vale /path/to/file.mdx), allowing you to more easily locate and address issues by line number.

docs/references/failures.mdx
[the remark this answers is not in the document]
fairlydurable: We srsly are not going to block on passive voice at this time. Jwahir did a lovely ticket over Docscember clearing out passive voice but we just want to improve things right now and not throw obstacles and we are pretty low-friction at this point with a great new manager.

This explanation you've added is an obvious win that servers the user. I'll let you know what can be improved, what should be improved or fixed, and what I find confusing. Beyond that? ¯\_(ツ)_/¯  We'll get to it. The guiding philosophy right now as I see it is keep moving things forward to serve our users better and incremental improvements are better than blocked tickets.

docs/encyclopedia/retry-policies.mdx
GSmithApps: Workflow task executions are different from activities and workflow executions. I'm not sure precisely what their retry policy is, so I didn't change any of that in this sentence. I just added the part I was quite sure of, which is that they don't always retry forever -- they retry until the workflow execution timeout, which is by default forever

docs/references/failures.mdx
fairlydurable: It's fine. Let it be.

docs/references/failures.mdx
fairlydurable: Good enough is a great thing.

fairlydurable (approved):
All blocking issues addressed.

Approving PR

PR #3282, opened 2025-01-08
EDU-3753: Clarify SSU vs non-SSU account cancellation

fairlydurable:
Clarify SSU vs non-SSU account cancellation on the billing and payments page.

Signed off by TLo and Patrick Liao


fairlydurable:
Since there is limited personnel availability, this PR is being merged without review. I’ll continue to ensure everything aligns as needed.

PR #3288, opened 2025-01-10
darker docsearch-muted-color

axfelix:


fairlydurable (approved):
Checked in Vercel, both light and dark mode.  The Search bar issue is no longer reported. 

Approving.

PR #3294, opened 2025-01-15
GCP: Add link to the PSC process

xaf:
## What does this PR do?

Enable the Google Cloud Private Service Connect link so it appears in the documentation.


docs/evaluate/temporal-cloud/gcp-private-service-connect.mdx
```diff
@@ -11,22 +11,34 @@ keywords:
...
 It establishes a private connection between your Google Virtual Private Cloud (VPC) and Temporal Cloud.
 This one-way connection means Temporal cannot establish a connection back to your service.
 This is useful if normally you block traffic egress as part of your security protocols.
 If you use a private environment that does not allow external connectivity, you will remain isolated.
 
+:::note
+
+If you are interested in leveraging Google Cloud Private Service Connect in your Namespaces, [create a support ticket](/cloud/support#support-ticket) that includes the following information:
+
+- Google Cloud Region: The Region in which your connection will go through.
```
cludden, line 36: We don't require the region, because the consumer must create the endpoint within the same region as the producer (e.g. temporal cloud namespace region). To enable cross-region connectivity within their VPC, they can enable Global Access on their consumer endpoint if they have the `GLOBAL` routing mode configured on their VPC.

XaF: You are right, and AWS now has a global option too which we should add to the documentation on the AWS side. That being said, as mentioned in the GCP call, I think for now it's better to keep asking the same information no matter for which cloud as it standardizes things across. Once we get to self-service with PCv3 we won't need any of those informations anymore and can clean things up!!

docs/evaluate/temporal-cloud/gcp-private-service-connect.mdx
```diff
@@ -44,10 +56,10 @@ Set up Private Service Connect with Temporal Cloud with these steps:
 
   - For **Target service**, fill in the **Service name** with the Private Service Connect Service Name for the region you’re trying to connect to:
 
-      | Region                 | Private Service Connect Service Name                                            |
-      | ---------------------- | ------------------------------------------------------------------------------- |
-      | `us-central1`          | `projects/PROJECT/regions/us-central1/serviceAttachments/temporal-api`          |
-      | `australia-southeast1` | `projects/PROJECT/regions/australia-southeast1/serviceAttachments/temporal-api` |
+      | Region                 | Private Service Connect Service Name                                                      |
+      | ---------------------- | ----------------------------------------------------------------------------------------- |
+      | `asia-south1`          | `projects/prod-d5spc2sfeshws33bg33vwdef7/regions/asia-south1/serviceAttachments/pl-7w7tw` |
```
MasonEgger, line 61: @xaf Is this project string accurate? It looks like some UUIDs. Or is this an example?

MasonEgger (commented):
One question, the rest looks good.

docs/evaluate/temporal-cloud/gcp-private-service-connect.mdx
xaf: These are the correct ones!

MasonEgger (approved):

PR #3309, opened 2025-01-27
consolidate web UI pages and update to reflect yaml

axfelix:


axfelix:
Ticket is https://temporalio.atlassian.net/browse/DEVAC-107, btw (sorry, I should've made that in EDU)

jsundai (approved):
LGTM

PR #3313, opened 2025-01-30
EDU-3858: Clarify instructions regarding Typescript setHandler

dandavison:
The instructions regarding `setHandler` were hard to understand and misleading: users are not in fact free to delay calling `setHandler` for Updates until after async initialization. See e.g. https://temporalio.slack.com/archives/C01DKSMU94L/p1738161795203779.

This PR moves to a direct instruction regarding how workflows should be written.

### DIFF
--- a/docs/develop/typescript/message-passing.mdx
+++ b/docs/develop/typescript/message-passing.mdx
@@ -122,14 +122,11 @@ export async function greetingWorkflow(): Promise<string> {
 - The handler cannot return a value.
   The response is sent immediately from the server, without waiting for the Workflow to process the Signal.
 
-- Signal (and Update) handlers can be `async`.
+- Signal and Update handlers can be `async`.
   This allows you to use Activities, Child Workflows, durable [`workflow.sleep`](https://typescript.temporal.io/api/namespaces/workflow#sleep) Timers, [`workflow.condition`](https://typescript.temporal.io/api/namespaces/workflow#condition) conditions, and more.
   See [Async handlers](#async-handlers) and [Workflow message passing](/encyclopedia/workflow-message-passing) for guidelines on safely using async Signal and Update handlers.
 
-- Delay calling [`workflow.setHandler`](https://typescript.temporal.io/api/namespaces/workflow#sethandler) until the Workflow initialization needed by Signal (or Update) handlers has finished.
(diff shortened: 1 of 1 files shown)


docs/develop/typescript/message-passing.mdx
```diff
@@ -122,14 +122,11 @@ export async function greetingWorkflow(): Promise<string> {
...
   See [Async handlers](#async-handlers) and [Workflow message passing](/encyclopedia/workflow-message-passing) for guidelines on safely using async Signal and Update handlers.
 
-- Delay calling [`workflow.setHandler`](https://typescript.temporal.io/api/namespaces/workflow#sethandler) until the Workflow initialization needed by Signal (or Update) handlers has finished.
-  This is safe because the SDK buffers messages when there are no registered handlers for them.
-  Note that [`workflow.setHandler`](https://typescript.temporal.io/api/namespaces/workflow#sethandler) will immediately invoke the handler of buffered Signals (or Updates) with matching types.
-  This could lead to out-of-order processing of messages with different types.
+- If your Workflow needs to do some async initialization before handling a Signal or Update, use [`workflow.condition`](https://typescript.temporal.io/api/namespaces/workflow#condition) inside your handler to wait until initialization has completed.
```
dandavison, line 129: @mjameswh I've removed the statement about `setHandler` immediately invoking buffered messages because I think it's too technical to explain buffering and diffrerences therein between Signal and Update in the docs. However, would it make sense to say something like

> You should typically call `setHandler` at the start of your Workflow.

?

mjameswh: > However, would it make sense to say something like
>> You should typically call setHandler at the start of your Workflow.

I don't think that would be useful.

> I've removed the statement about setHandler immediately invoking buffered messages because I think it's too technical to explain buffering and differences therein between Signal and Update in the docs

I have had to explain that loophole a few times, so I would like to have that information somewhere that users can find, though I totally agree this is a bit too much for this specific section. Let's remove it from here, and I'll find some better place for that.

fairlydurable (approved):
- All changes read through
- Vercel compiles without link or anchor issues
- Technical review has been completed

fairlydurable:
@dandavison @mjameswh Thank you for your work on this enhancement.

PR #3320, opened 2025-02-05
EDU-3927: [Ready-to-roll] fixed curly braces in API keys doc

GSmithApps:
This TS call needs curly braces


fairlydurable (approved):
[Laissez les bons typescript braces rouler](https://en.wikipedia.org/wiki/Laissez_les_bons_temps_rouler)

PR #3331, opened 2025-02-10
EDU-3951: Make clear that usage-by-Namespace is restricted by role

fairlydurable:
- Audit billing-and-cost page
- Update wording
- Waiting on expert sign-off


brianmacdonald-temporal (approved):
Changes look good; approved.

PR #3333, opened 2025-02-10
EDU-3857: Image audit and organization

fairlydurable:
- Does not include image optimization/SEO
- Needs careful review


axfelix (approved):
This looks good to me! Can't find any oversights. Nice work.

PR #3346, opened 2025-02-12
EDU-3972: Adds definitions in-line

fairlydurable:
Present definitions in-line so users don’t leave the site or the page and to boost our authority.

Tested with screen readers and on mobile.

```
import ExpandableDefinition from '@site/src/components/definitions/ExpandableDefinition';

Text starts <span class="nowrap">replica<ExpandableDefinition label="Definition" definition="Something something definition Namespace becomes unavailable." /></span>, ...more text
```

- Note that you must use the span wrap around the word and its definition to keep them together.
- The definition will push the defined word away from the comma. There’s no way to delay the definition other than using, for example, a <detail>, which is overkill here.

### DIFF
--- a/src/components/definitions/ExpandableDefinition.js
+++ b/src/components/definitions/ExpandableDefinition.js
@@ -1,0 +1,44 @@
+const ExpandableDefinition = ({ label = "Definition", definition }) => {
+  return (
+    <span style={{ display: 'inline-flex', alignItems: 'center', whiteSpace: 'nowrap' }}>
+      <button
+        onClick={toggleDefinition}
+        style={{
+          background: 'none',
+          border: 'none',
+          color: '#007acc',
+          cursor: 'pointer',
+          fontSize: '0.9rem',
+          margin: '0 5px',
+          verticalAlign: 'middle',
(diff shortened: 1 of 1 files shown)


axfelix:
Sorry, where can I see an example of this? The example shared in Slack looked good, I thought, but it was a lot of serif fonts in a way that seemed to go against the style of our docs site, and it was a lot of content to be hidden under a single drop-down. Not sure if that pertains to this PR directly or if it should be a general style point, though.

axfelix:
Previous comment may have pertained more to https://github.com/temporalio/documentation/pull/3345, sorry — hard to tell where I should be looking here.

fairlydurable:
@axfelix Yeah, that's the problem. These are not incorporated into actual samples because that would affect our docs directly. I can do it for the disclosure one with pricing, which I think is not much of an issue.

As for definitions, I do want to have decent style. I put the screen-shots and a video out for review a week or so ago and got no feedback. I assume that we put things in-place and then can refine. 

I do want your input as to better design with the first merge though.

fairlydurable:
Closed. Merged with High Availability.

PR #3360, opened 2025-02-18
Fix formatting

dandavison:
WISOTT.

Viewing the PR with whitespace changes hidden shows that there are no non-whitespace changes.


jsundai:
Thanks for the contribution, @dandavison. 🙌 Looks like when I run `yarn format`, it doesn't like the `decode` format and moves it back. Other than that, LGTM.

jsundai (approved):

axfelix:
@dandavison Yeah, I can see the issue here, but you can probably see why dprint behaves that way -- it wants to dedent any given sample it sees, it can't really carry forward indentation level from one sample to the next (that would bake in some strange expectations).

PR #3372, opened 2025-02-20
fix inconsistency in docs where we reference tcld in the temporal cli section

mattkim:
## What does this PR do?
fix inconsistency in docs where we reference tcld in the temporal cli section

## Notes to reviewers
while reviewing the docs I noticed a discrepancy where we reference `tcld` in the `temporal cli` section which is inocrrect


brianmacdonald-temporal (commented):
Looks good.

jsundai (approved):
Added JIRA task [EDU-4008](https://temporalio.atlassian.net/browse/EDU-4008) and capitalized CLI. Thanks for catching that, @mattkim.

PR #3395, opened 2025-02-28
.NET update-with-start docs

cretz:
## What does this PR do?

This creates .NET documentation that matches the Python equivalent for update with start.

## Notes to reviewers

If there are any concerns with the general English herein, they may be general concerns with already-present documentation and not specific to this review, and should be handled generally. In many cases due to overloads, .NET cannot link single API doc locations like Python can, but like in similar .NET docs, we just don't link.

### DIFF
--- a/docs/develop/dotnet/message-passing.mdx
+++ b/docs/develop/dotnet/message-passing.mdx
@@ -340,6 +340,67 @@ To send an Update to a Workflow Execution, you can:
 
   For more details, see the "Async handlers" section.
 
+#### Update-with-Start {#update-with-start}
+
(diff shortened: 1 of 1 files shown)


cretz:
@fairlydurable - I see you made changes at https://github.com/temporalio/documentation/pull/3395/commits/198d0f7984e2317426165659ae3c49e61b9ac1bb, but those changed pieces were taken almost verbatim from https://github.com/temporalio/documentation/blob/main/docs/develop/python/message-passing.mdx?plain=1. Is another PR needed make these same changes to the Python equivalent?

PR #3398, opened 2025-02-28
Nexus GA polish

bergundy:
Document context propagation and tracing for Nexus Operations.

### DIFF
--- a/docs/develop/go/observability.mdx
+++ b/docs/develop/go/observability.mdx
@@ -73,7 +73,7 @@ For more information, see the [Go sample for metrics](https://github.com/tempora
 
 The Temporal Go SDK supports three tracing implementations: [Datadog](https://pkg.go.dev/go.temporal.io/sdk/contrib/datadog), [OpenTelemetry](https://pkg.go.dev/go.temporal.io/sdk/contrib/opentelemetry), and [OpenTracing](https://pkg.go.dev/go.temporal.io/sdk/contrib/opentracing).
 
-Tracing allows you to view the call graph of a Workflow along with its Activities and any Child Workflows.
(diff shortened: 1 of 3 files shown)


docs/develop/go/observability.mdx
```diff
@@ -156,6 +156,16 @@ It also uses Jaeger for tracing.
 
 Yes. Multiple context propagators help to structure code with each propagator having its own scope of responsibility.
 
+### Context Propgation Over Nexus Operation Calls
```
fairlydurable, line 159:
```suggestion
### Context Propagation Over Nexus Operation Calls
```

Quinn-With-Two-Ns (approved):

docs/develop/go/observability.mdx
```diff
@@ -156,6 +156,17 @@ It also uses Jaeger for tracing.
 
 Yes. Multiple context propagators help to structure code with each propagator having its own scope of responsibility.
 
+### Context Propagation Over Nexus Operation Calls
```
prasek, line 159: @bergundy @Quinn-With-Two-Ns can we add a similar section to the Java SDK guide?

fairlydurable (approved):
No issues. 

Typos fixed, sentences broken down and clarified for brevity, and brought into Docs alignment.

Good work. Thank you for these.

docs/develop/go/observability.mdx
bergundy: Java doesn't have a comprehensive context propagation section and I didn't want to take on adding all of this extra non-Nexus related stuff for now.

I think we should add it though, just don't think Nexus should be treated specially here.

I would open an issue and ask the SDK team to prioritize since it's not a trivial amount of work.

PR #3404, opened 2025-02-28
EDU-4047: More Nexus GA polish

bergundy:


Quinn-With-Two-Ns (approved):

prasek (approved):

fairlydurable (approved):
Made things more readable, put sentences on different lines. 
I did not do a full-file edit, only addressed `### Canceling a Nexus Operation` and the other local changes. 

Thank you for your passion, which shines through in this writing and the double-polishing. It's always a pleasure to work with your team.

Approving PR.

---

Reference:

**This is why we put items on separate lines, nominally sentences:**

* Improves version control diffs for easier line-by-line inspection.
* Focuses feedback and review discussions on single points, preventing overlap with other content that may need changes.
* Reduces overlap and merge conflicts with other tickets.
* Enhances the precision of our automated lint tooling, such as Vale (vale /path/to/file.mdx), allowing you to more easily locate and address issues by line number.

**Further, we break down CLI commands to one flag or option per line for readability.**

PR #3410, opened 2025-03-03
Reordered Glossary so that "Nexus Registry" was in the correct alphab…

brianmacdonald-temporal:
## What does this PR do?

Reorders the glossary to put the entry "Nexus Registry" in the correct alphabetical spot.


jsundai (approved):
LGTM

PR #3470, opened 2025-03-19
EDU-4126: Removes SDK references to statsd and M3DB which is misleading users to think they're supported natively

fairlydurable:
Updated with Chad's feedback


tomwheeler (approved):
LGTM. Ship it!

PR #3483, opened 2025-03-25
EDU-4199: Convert tables of instructions to tabs

fairlydurable:
@brianmacdonald-temporal and @atihkin both hated the instructions-in-tables approach.

All tables used in High Availability are now updated to use tabbed instructions: Web UI and tcld.

image
image

versus

image


brianmacdonald-temporal (approved):
Changes look good. Thanks!

PR #3485, opened 2025-03-26
Caution against blocking in WorkflowInit Java docs

drewhoskins-temporal:
## What does this PR do?

A user got confused when using Update with Start in Java, and got "not found" errors from the update.  Turns out, they were doing a blocking call from their `WorkflowInit` method, which delayed registration of the update handler.

Docs should clarify this.  (This only applies to Java, because in .NET and Python, this is statically enforced.)

## Test plan
```
yarn start
http://localhost:3000/develop/java/message-passing
```
Screenshot 2025-03-26 at 11 48 28 AM


docs/develop/java/message-passing.mdx
```diff
@@ -593,6 +593,12 @@ The SDK will then ensure that your constructor receives the Workflow input argum
 The Workflow input arguments are also passed to your `@WorkflowMethod` method -- that always happens, whether or not you use the `@WorkflowInit` annotation.
 This is useful if you have message handlers that need access to Workflow input: see [Initializing the Workflow first](/handling-messages#workflow-initializers).
 
+:::caution
+
+Do not make blocking calls from within your `@WorkflowInit` method. This could result in your Workflow being incompletely initialized at the start, meaning, for example, that Signal and Update handler registration would be delayed.
```
Quinn-With-Two-Ns, line 598:
```suggestion
Do not make blocking calls from within your `@WorkflowInit` method. This could result in your Workflow being incompletely initialized at the start, meaning, for example, that Signal, Query, and Update handler registration would be delayed.
```

Quinn-With-Two-Ns (approved):

fairlydurable (approved):
Quinn signed off.

Looks fine.

PR #3486, opened 2025-03-26
EDU-4213: [DO NOT MERGE] Creates a prototype following best practices

fairlydurable:
What if we followed the principles we are discussing, such as visibility, verbs-not-nouns, etc.


fairlydurable:
Prototype presented, discussed. Closing.

PR #3498, opened 2025-03-31
add link to Nexus GA blog post

prasek:
## What does this PR do?
- add link to the Nexus GA blog post


fairlydurable (changes requested):
I recommend removing the admonitions and using a different approach to link to the blog post.

[PR #3498](https://github.com/temporalio/documentation/pull/3498) breaks our support, stability, and dependency info practices for these reasons:
We normally remove this admonition entirely for general availability.
We only link to stage definitions, and this patch breaks that.
The "Nexus now available" blog link must now be maintained and will age.
I think posting an announcement on the Docs website using the built-in banner feature would be more appropriate, and recommend removing the support, stability, and dependency info entirely now that Temporal Nexus is Generally Available.

fairlydurable (approved):

PR #3536, opened 2025-04-17
clarify use of snipsync

axfelix:


tomwheeler (approved):
I think the link target on line 126 could suggest that we only have one samples repo (one for the Go SDK). That is at odds with the prose, whose use of the plural suggests that we have multiple of them (which is indeed the case). It would be best to eliminate that confusion, but I will approve the PR so as not to stand in the way of the greater good.

PR #3559, opened 2025-04-29
EDU-4415: Work around Docusaurus bug by setting commented files to draft

fairlydurable:
Patches visibility issues for unlinked pages
    
- Performed audit for files that do not appear in sidebars.js
- Hand-checked each of these files
- Set any file within this set that was 
    1. not in sidebars.js or commented out; and
    2. displayed with no sidebar
  
  to draft

image


axfelix (approved):
Thanks for going back and forth with me on this! Approved.

PR #3560, opened 2025-04-30
ruby sdk dev guide [draft] 

jsundai:


docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+      { greeting: 'Hello', name: },
+      start_to_close_timeout: 100
+    )
+  end
+end
+```
+
+### Workflow Futures {#workflow-futures}
+
+`Temporalio::Workflow::Future` can be used for running things in the background or concurrently.
```
axfelix: This is a bit of an abrupt explanation, I think we need another sentence here:

"Futures are a language construct that allow you to kick off multiple Activities concurrently and wait for them all to return."

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+This is basically a safe wrapper around `Fiber.schedule` for starting and `Workflow.wait_condition` for waiting.
```
axfelix: I would remove "basically" and rewrite this slightly:

```suggestion
Temporal provides Workflow-safe wrappers around some core language features in cases like these.
`Temporalio::Workflow::Future` is a safe wrapper around `Fiber.schedule` for running multiple Activities at once.
The Ruby SDK also provides `Workflow.wait_condition` for awaiting a result.
```

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+Nothing uses futures by default, but they work with all workflow code/constructs.
```
axfelix:
```suggestion
Futures are never used implicitly, but they work with all Workflow code and constructs.
```

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+# Wait for first act result or sleep fut
+act_result = Temporalio::Workflow::Future.any_of(sleep_fut, *act_futs).wait
+# Fail if timer done first
+raise Temporalio::Error::ApplicationError, 'Timer expired' if sleep_fut.done?
+# Print act result otherwise
+Temporalio::Workflow.logger.info("Act result: #{act_result}")
+```
+
+There are several other details not covered here about futures, such as how exceptions are handled, how to use a setter
+proc instead of a block, etc. See the API documentation for details.
```
axfelix: This should be a link, probably?

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+    Temporalio::Workflow.execute_activity(
+      MyActivity,
+      { greeting: 'Hello', name: },
+      start_to_close_timeout: 100
+    )
+  end
+end
+```
+
+### Workflow Futures {#workflow-futures}
```
axfelix: This H3 should be moved to an H4 at the end of the "Start Activity Execution" section imo, explaining Futures before Activities is unintuitive.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+    "#{input['greeting']}, #{input['name']}!"
+  end
+end
+```
+
+There is no explicit limit to the total number of parameters that an [Activity Definition](/activity-definition) may support.
+However, there is a limit to the total size of the data that ends up encoded into a gRPC message Payload.
+
+A single argument is limited to a maximum size of 2 MB.
+And the total size of a gRPC message, which includes all the arguments, is limited to a maximum of 4 MB.
```
axfelix, line 122: I'd add a sentence here (to all of the SDKs if the explanation is consistently this dry) that says, "it is almost never necessary to send large binary data as input and output to your Workflows. If this is ever required for your business logic, you can implement an S3 storage hook following this example: https://github.com/DataDog/temporal-large-payload-codec" or something along those lines. I think this limit is a bit confusing out of context otherwise.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+Also, keep in mind that all Payload data is recorded in the [Workflow Execution Event History](/workflow-execution/event#event-history) and large Event Histories can affect Worker performance.
+This is because the entire Event History could be transferred to a Worker Process with a [Workflow Task](/tasks#workflow-task).
```
axfelix: I would consider just removing this paragraph, I don't think it belongs this early in the SDK docs and it's a bit scary. Should be an encyclopedia concept only.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+Some SDKs require that you pass context objects, others do not.
+When it comes to your application data—that is, data that is serialized and encoded into a Payload—we recommend that you use a single hash or object as an argument that wraps the application data passed to Activities.
+This is so that you can change what data is passed to the Activity without breaking a method signature.
```
axfelix, line 126: Should have an example here using a Ruby object rather than just talking about it

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+Activity parameters are the method parameters of the `execute` method.
+These can be any data type Temporal can convert.
+Technically this can be multiple parameters, but Temporal strongly encourages a single parameter containing all input fields.
```
axfelix: This is redundant with the preceding paragraph

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+### Activity Concurrency and Executors {#activity-concurrency-and-executors}
```
axfelix, line 131: This should probably also go under "Start Activity Execution" as an H4, and maybe even in an admonition to clarify that these are advanced threading concepts that people will not need as they're starting out.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+    )
+  end
+end
+```
+
+Activity Execution semantics rely on several parameters.
+The only required value that needs to be set is either a [Schedule-To-Close Timeout](/encyclopedia/detecting-activity-failures#schedule-to-close-timeout) or a [Start-To-Close Timeout](/encyclopedia/detecting-activity-failures#start-to-close-timeout).
+These values are set as keyword parameters.
+
+### Get Activity Execution results {#get-activity-results}
```
axfelix: I would remove this H4 and just keep the sentence beneath it.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+Activities can only choose the fiber executor if the worker has been created and run in a fiber, but thread pool executor is always available.
+Currently due to [an issue](https://github.com/temporalio/sdk-ruby/issues/162), workers can only run in a fiber on Ruby versions 3.3 and newer.
+
+## Start Activity Execution {#activity-execution}
+
+Calls to spawn [Activity Executions](/activity-execution) are written within a [Workflow Definition](/workflow-definition).
+The call to spawn an Activity Execution generates the [ScheduleActivityTask](/references/commands#scheduleactivitytask) Command.
+This results in the set of three [Activity Task](/tasks#activity-task) related Events ([ActivityTaskScheduled](/references/events#activitytaskscheduled), [ActivityTaskStarted](/references/events#activitytaskstarted), and ActivityTask[Closed])in your Workflow Execution Event History.
+
+Activity implementation code should be _idempotent_.
```
axfelix: *This* should be under Develop an Activity, not Start Activity Execution, and it should have a link to another part of our docs explaining this...

docs/develop/ruby/core-application.mdx
cretz: Futures are not a language construct, more like an SDK construct, and they allow you to kick off anything concurrently, not specific to activities. Any rewording or additional context here works for me.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+The only required value that needs to be set is either a [Schedule-To-Close Timeout](/encyclopedia/detecting-activity-failures#schedule-to-close-timeout) or a [Start-To-Close Timeout](/encyclopedia/detecting-activity-failures#start-to-close-timeout).
+These values are set as keyword parameters.
+
+### Get Activity Execution results {#get-activity-results}
+
+The Activity result is the returned from the `execute_activity` call.
+
+## Run Worker Process {#run-worker-process}
+
+The [Worker Process](/workers#worker-process) is where Workflow Functions and Activity Functions are executed.
```
axfelix:
```suggestion
The [Worker Process](/workers#worker-process) is where Workflow Functions and Activity Functions are actually executed.
In a Temporal application deployment, you ship and scale as many Workers as you need to handle the load of your Workflows and Activities.
```

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+- Each [Worker Entity](/workers#worker-entity) in the Worker Process must register the exact Workflow Types and Activity Types it may execute.
+- Each Worker Entity must also associate itself with exactly one [Task Queue](/task-queue).
+- Each Worker Entity polling the same Task Queue must be registered with the same Workflow Types and Activity Types.
```
axfelix, line 247: These bullet points should be next to the little code sample down below and softened somewhat, otherwise it feels like we're giving people a lot of scary / arbitrary rules before they've seen how task queue assignment actually works.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+
+A [Worker Entity](/workers#worker-entity) is the component within a Worker Process that listens to a specific Task Queue.
+
+Although multiple Worker Entities can be in a single Worker Process, a single Worker Entity Worker Process may be perfectly sufficient.
```
axfelix: I would cut this from this guide, doesn't make any sense to anyone who hasn't already gone deeper into scaling and deploys.

docs/develop/ruby/core-application.mdx
```diff
@@ -0,0 +1,338 @@
...
+For more information, see the [Worker tuning guide](/develop/worker-performance).
+
+A Worker Entity contains a Workflow Worker and/or an Activity Worker, which makes progress on Workflow Executions and Activity Executions, respectively.
+
+To develop a Worker, create a new `Temporalio::Worker` providing the Client and worker options which include Task Queue, Workflows, and Activities and more.
```
axfelix: I think the word here shouldn't necessarily be "develop" because you don't need to do a lot of development to ship a Worker. Maybe something like:

```suggestion
Workers are implemented in each Temporal SDK, and can be deployed with just a bit of boilerplate.
To create a Worker, use `Temporalio::Worker.new()`, providing the Worker options which include Task Queue, Workflows, and Activities and more.
```

docs/develop/ruby/temporal-client.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+A [Temporal Client](/encyclopedia/temporal-sdks#temporal-client) enables you to communicate with the [Temporal Service](/temporal-service).
+Communication with a Temporal Service includes, but isn't limited to, the following:
+
+- Starting Workflow Executions.
+- Sending Signals to Workflow Executions.
+- Sending Queries to Workflow Executions.
+- Getting the results of a Workflow Execution.
+- Providing an Activity Task Token.
+
```
axfelix, line 35:
```suggestion
All applications that need to communicate directly with the Temporal Service will run inside of Temporal Clients.

Your Workers are technically Temporal Clients, because they need to poll the Temporal Service to actually do work.
Your Workflow and Activity definitions are *not* Temporal Clients -- your Worker imports these directly instead.
Each Temporal SDK includes a set of Temporal Client functions.
The `temporal` CLI is also a Temporal Client, and is generally at feature parity with the SDKs.
```

docs/develop/ruby/versioning.mdx
```diff
@@ -0,0 +1,164 @@
+---
```
axfelix, line 1: I'll be overhauling this page pretty significantly once this PR gets merged, FYI -- I need to propagate my Worker Versioning rewrites of the other SDKs to Ruby/PHP sometime hopefully next week.

docs/develop/ruby/converters-and-encryption.mdx
```diff
@@ -0,0 +1,179 @@
...
+    end
+  end
+end
+```
+
+Now if `include ActiveModelJSONSupport` is present on any ActiveModel class, on serialization `to_json` will be used which will use `as_json` which calls the super `as_json` but also includes the fully qualified class name as the JSON
+`create_id` key.
+On deserialization, Ruby JSON then uses this key to know what class to call `json_create` on.
+
+## Use a custom Payload Codec
```
axfelix: What's this doing all the way down here? Probably can get rid of it, looks like it's accidentally redundant.

jsundai:
Thank you, [axfelix](https://github.com/axfelix)! I'll go through these today/tomorrow.

docs/develop/ruby/core-application.mdx
cretz: I would like us to not recommend large payload codec, it is not a good pattern and has many problems. We admittedly lack a good blog post or docs on how best to handle this (can discuss internally on prioritizing that), but we should not encourage large payload codec.

axfelix: Fine by me, in that case I revise my feedback to clarifying that you can usually just send strings or identifiers instead of needing to send large blobs as I/O

docs/develop/ruby/core-application.mdx
cretz: The "API documentation" is actually many anchor links depending on what they are trying to do. May be able to link to https://ruby.temporal.io/Temporalio/Workflow/Future.html as a container for those other things they can do.

docs/develop/ruby/core-application.mdx
cretz: Not following. Futures have nothing to do with activities, they are a workflow utility for concurrent code running.

docs/develop/ruby/core-application.mdx
cretz: The struggle I had is that we are trying to combine "fix general SDK docs" and "make Ruby SDK docs" into one. Can we do the general SDK doc improvement that applies to multiple SDKs a separate PR? Ideally we should get Ruby like others, and fix _all_ others if there is a problem with _all_ others. This makes it easier to port.

docs/develop/ruby/core-application.mdx
cretz: We chose not to do this in other SDK docs. If we want to consistently provide this across SDKs, I think we should do so across SDKs and not make Ruby unique here.

docs/develop/ruby/core-application.mdx
cretz: This is also the case in other SDK docs. Are you sure you only want to improve here and not there?

docs/develop/ruby/core-application.mdx
axfelix: Generally yes, and that's the model I've followed with https://github.com/temporalio/documentation/pull/3598, but you might be interested to know that I've gotten just as much feedback from people who don't like *that* model either, and there are a lot of clarity improvements we can make here that would not have gotten looked at if *not* for this porting opportunity. I am happy to add the other SDK page changes in this PR but I think people would then start talking about scope creep and blocking Ruby...

docs/develop/ruby/core-application.mdx
cretz: This section is very much for implementation/definition of activities. The starting of activities is unrelated to this implementation aspect.

docs/develop/ruby/core-application.mdx
cretz: Does this only apply to Ruby docs?

docs/develop/ruby/core-application.mdx
cretz: Does this only apply to Ruby docs?

docs/develop/ruby/core-application.mdx
cretz: Does this only apply to Ruby docs? Can we do a separate SDK docs review not specific to Ruby?

docs/develop/ruby/core-application.mdx
axfelix: See https://github.com/temporalio/documentation/pull/3560#discussion_r2162118932, and tell me which method of improvement you would object to least :)

docs/develop/ruby/core-application.mdx
cretz: In general at least with code we don't like to mix refactoring and greenfield work because the refactoring gets lost in the details. If we have things we need to improve across SDK docs, it should be improved across SDK docs IMO separately from a single language migration.

Sounds like several of these things need to be improved generally and a PR should be opened for those unrelated to Ruby. The SDK docs gain more and more inconsistency if we apply certain improvements only to specific languages and leave them in others.

docs/develop/ruby/core-application.mdx
axfelix: Right -- but if someone is reading this page top to bottom for the first time and they've just encountered the concept of a Workflow (a real possibility, if they are a Ruby developer first approaching Temporal), Futures are not going to make sense to them as the very first Workflow feature they've learned about. They *do* make sense in the context of concurrently dispatching Activities, and I believe this is the context we typically bring them up in. If this isn't clean enough conceptually for you, I would strongly advocate for them to go on a different page, because I think we are reproducing a very common complaint about our docs otherwise, namely that they introduce platform features in too technically correct and unintuitive a way.

cretz: I see this as reference material, not top-to-bottom read material, and I expect hierarchical nav to represent that. Workflow utilities go with workflows, activity utilities go with activities, etc. I don't expect to have workflow utility reference material fragmented as you don't have anywhere to go to see all workflow utility pieces. I don't expect to have a futures section for child workflows, nexus operations, etc even though they apply to those too.

But these are just my opinions on how I see these docs vs something on learn.temporal.io.

axfelix: I think that's a good argument, but it's one we've been having here for years, and in practice, I think we wind up continually returning to the reality that people are approaching it 50-50 as reference material and top-to-bottom material. Notably, one place this PR *does* already deviate from the others in a way I didn't comment on today is by adding a Quickstart page to the top of the Ruby SDK section -- I thought this was a neat experiment that we'd build on -- in light of that finding. This is, imo, one of the only places where we're seriously violating an assumption that people might be top-to-bottom, hence why I'm drawing attention to this one.

docs/develop/ruby/core-application.mdx
jsundai: An admonition definitely works in this case.

docs/develop/ruby/core-application.mdx
jsundai: Agreed

PR #3562, opened 2025-04-30
EDU-4420: Adds audit-stray-pages and updates READMEs

fairlydurable:
- Adds auditing tool
- Creates UTILITIES.md
- Updates README.md to point to all read-me files
- Adds PhotoCarousel coverage to COMPONENTS.md

### DIFF
--- a/COMPONENTS.md
+++ b/COMPONENTS.md
@@ -21,6 +21,7 @@ Whether you’re using core components or experimenting with new ones, this guid
 - [Using CaptionedImage](#using-captionedimage)
 - [Using DiscoverableDisclosure](#using-discoverabledisclosure)
 - [Using DocsTable](#using-docstable)
+- [Using PhotoCarousel](#using-photocarousel)
 - [Using RelatedRead](#using-relatedread)
 - [Using ToolTipTerm](#using-tooltipterm)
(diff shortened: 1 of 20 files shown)


COMPONENTS.md
```diff
@@ -262,6 +230,44 @@ Archetypes:
...
+...
+]}
+/>
+```
+
+- If links are stored on-site, place images in static in an appropriately themed folder.
+- If links are stored off-site, use a fully specified URL
+- Please ensure the caption count and order matches that of the images
+- You may pass an optional `isDarkMode` Boolean prop
+
```
angelazhou32, line 270: Also, if you put "NA", then it skips a caption for that photo. For example, 

<PhotoCarousel
images={[
"/link/to/image-1",
"/link/to/image-2",
...
]}
captions={[
"NA",
"caption-2",
...
]}
/>

Photo 1 will not have a caption but photo 2 will have a caption. This is important because if we just don't write anything, then photo 2 will be matched with caption 1.

fairlydurable: Adding!

brianmacdonald-temporal (approved):
Changes look good; thanks very much.

PR #3564, opened 2025-05-01
EDU-4374: Change redirect for tctl content from /cli to /tctl-v1

fairlydurable:
Moves tctl redirects back from Temporal CLI and to the deprecated (but accessible) tctl-v1 pages as some users continue to use this content after deprecation


tomwheeler (approved):
I tested this by replacing `docs.temporal.io` in the redirect URL with the hostname where this PR is deployed. I then navigated to resulting URL (`https://temporal-documentation-gpldxsa00.preview.thundergun.io/docs/system-tools/tctl`) and observed that it redirected me to the expected location (`https://temporal-documentation-gpldxsa00.preview.thundergun.io/tctl-v1`).

Approved (and thank you)!

PR #3578, opened 2025-05-13
EDU-4430 python sync vs async

GSmithApps:
## What does this PR do?

Merges [this doc](https://github.com/temporalio/dev-success/blob/main/python/python-worker-and-activity-config.md) into [this docs page](https://docs.temporal.io/develop/python/python-sdk-sync-vs-async)

## Note

When we merge this, we should also merge [this](https://github.com/temporalio/dev-success/pull/5).


docs/develop/python/python-sdk-sync-vs-async.mdx
```diff
@@ -67,65 +67,48 @@ asynchronous Activity would lead to blocking your event loop. If you want to mak
...
-the greeting from the body of the response.
+Python workers have following components for executing code:
 
-```python
-import aiohttp
-import urllib.parse
-from temporalio import activity
+- Your event loop, which runs tasks from async activities **plus the rest of Temporal, such as communicating with the server**.
+- An executor for executing activity tasks from synchronous activities. A thread pool executor is recommended.
+- A thread pool executor for executing workflow tasks (see forum post [here](https://community.temporal.io/t/whats-the-workflow-task-executor-in-the-python-worker-configuration/16965)).
```
MasonEgger, line 76: Can we just port this content into the docs directly? It's best to avoid external links when able, and this content seems relevant to this page.

docs/develop/python/python-sdk-sync-vs-async.mdx
```diff
@@ -175,6 +158,86 @@ Whether to implement Activities as class methods or functions is a design choice
...
+                raise ApplicationError(
+                    f"HTTP Error {response.status}: {translation}",
+                    # We want to have Temporal automatically retry 5xx but not 4xx
+                    non_retryable=response.status < 500,
+                )
+
+            return translation
+```
+
+### Running synchronous code from an asynchronous function
```
MasonEgger, line 231: We should remove this entire section, according to the Python SDK readme, both of those options (`run_in_executor` and `to_thread`) are disabled in our event loop. https://github.com/temporalio/sdk-python?tab=readme-ov-file#asyncio-and-determinism

MasonEgger (changes requested):

docs/develop/python/python-sdk-sync-vs-async.mdx
GSmithApps: Thanks for the comment! Our event loop and that section of the readme apply to the custom event loop in workflows, not activities, so the custom event loop doesn’t apply here 👍 

I see that the section heading could be confusing though so I changed it from “**Running synchronous code from an asynchronous function**” to “**Running synchronous code from an asynchronous activity**”

docs/develop/python/python-sdk-sync-vs-async.mdx
GSmithApps: Love it. Done 👍

MasonEgger (approved):

PR #3584, opened 2025-05-14
remove extra accent character from create search attribute command

kepe-temporal:
There was an extra accent on this command at the end.

This pr simply removes that unnecessary accent


jsundai (approved):

jsundai:
LGTM

PR #3585, opened 2025-05-14
fix typescript example erroneous syntax

sdedovic:
This PR fixes the Typescript examples for API Key based connections which were malformed and unusable as-is.


jsundai (approved):

jsundai:
Thanks for the contribution!

PR #3589, opened 2025-05-16
Add latest tested postgres versions

prathyushpv:
## What does this PR do?
Postgres versions listed as supported are older versions. We now run release validation pipelines on newer versions.


yiminc (approved):

jsundai (approved):
LGTM

PR #3603, opened 2025-06-05
Update documentation to indicate activity size limit across all SDKs

billrich2001:
## What does this PR do?

correct statement regarding activity return value size limit (in the current [docs](https://docs.temporal.io/develop/go/core-application#activity-return-values) it says there is no size limit, but it should state that it is constrained by the payload limits)

## Notes to reviewers

Used ai to write correction. Happy to edit in any way needed.


jsundai:
Taking a look at this today

PR #3606, opened 2025-06-07
Revamp Java/PHP/TS Continue-as-New documentation

yuandrew:
## What does this PR do?
Update language specific CaN docs, similar to #3579

## Notes to reviewers
Docs should match Go and Python, just with the sample code and links specific to the language.

TS sample was missing the last bit of testing CaN with `shouldContinueAsNew`, so skipped that part of the doc.

### DIFF
--- a/docs/develop/typescript/continue-as-new.mdx
+++ b/docs/develop/typescript/continue-as-new.mdx
@@ -1,48 +1,79 @@
 ---
 id: continue-as-new
-title: Continue-As-New - TypeScript SDK
+title: Continue-As-New - Typescript SDK
 sidebar_label: Continue-As-New
+description: Learn how to use Temporal's Continue-As-New in Typescript to manage large Event Histories by atomically creating new Workflow Executions with the same Workflow Id and fresh parameters.
 toc_max_heading_level: 2
 keywords:
-  - continue-as-new
+  - continue-as-new workflow
+  - restart workflow
+  - fresh event history
+  - avoid large event histories
+  - temporal typescript continue-as-new
 tags:
   - Workflows
   - continue-as-new
-  - TypeScript SDK
+  - Typescript SDK
   - Temporal SDKs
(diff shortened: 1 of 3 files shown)


yuandrew:
Marking as draft to update dotnet, previously thought we were blocked for dotnet docs, but I was mistaken.

yuandrew:
dotnet has been added, ready for review

docs/develop/typescript/continue-as-new.mdx
```diff
@@ -1,48 +1,110 @@
...
-<!--SNIPEND-->
+export async function clusterManagerWorkflow(input: ClusterManagerInput = {}): Promise<ClusterManagerStateSummary> {
+
+````
+The test hook in the above snippet is covered [below](#how-to-test).
+
+Inside your Workflow, call the [`ContinueAsNew()`](https://typescript.temporal.io/api/classes/workflow.ContinueAsNew) function with the same type.
+This stops the Workflow right away and starts a new one.
+
+<div class="copycode-notice-container">
```
axfelix, line 59: One nit -- is this actually showing a `ContineuAsNew()` function being called? This looks like we're returning `wf.continueAsNew` instead of calling something. Language can sometimes get confusing here when porting between eg Go and TS, so I'm not sure if this is what happened. These are also very, very short code samples; if it were up to me I think they might be easier to read if they looked something like:

```
## Your other Workflow code
...
ContinueAsNew()

```

But I don't feel too strongly about that.

Anyway -- after clarifying this sample I think we're good to merge!

yuandrew: Ahh good callout, the issue is actually that I linked the wrong `continueAsNew` method, pushed a fix

yuandrew: As for the `Your other Workflow code` suggestion, I decided against it because this is linking directly to the typescript sample, they can click on the link to view the whole sample for how it fits into the workflow

axfelix (approved):

PR #3614, opened 2025-06-12
Improve determinism explanation

Sushisource:
## What does this PR do?
Provide the answers people are looking for up front in simpler language before diving into details.


docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+
+- Starting or cancelling a Timer
+- Scheduling or cancelling Activity executions (including local Activities)
+- Starting or cancelling Child Workflow executions
+- Signalling or cancelling signals to external Workflow executions
+- Scheduling or cancelling Nexus operations
+- Ending the Workflow Execution in any way (completing, failing, cancelling, or continuing as new)
+- Patched or GetVersion calls for versioning (though they may be added or removed according to the [patching](#patching) rules)
+- Upserting Workflow search attributes
+- Upserting Workflow memos
```
Quinn-With-Two-Ns: nit: can include `SideEffect` and `MutableSideEffect` here too

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
-
-The execution semantics of a Workflow Execution include the re-execution of a Workflow Function, which is called a [Replay](/workflow-execution#replay).
-The use of Workflow APIs in the function is what generates [Commands](/workflow-execution#command).
-Commands tell the Temporal Service which Events to create and add to the Workflow Execution's [Event History](/workflow-execution/event#event-history).
-When a Workflow Function executes, the Commands that are emitted are compared with the existing Event History.
-If a corresponding Event already exists within the Event History that maps to the generation of that Command in the same sequence, and some specific metadata of that Command matches with some specific metadata of the Event, then the Function Execution progresses.
+A critical aspect of developing Workflow Definitions is ensuring they are deterministic. Generally
+speaking, this means you must take care to ensure that any time your Workflow code is executed it
+makes the same Workflow API calls in the same sequence, given the same input. Some changes to those
+
```
tomwheeler, line 162: Seems to be an unnecessary blank line.

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+API calls are safe to make. For example, you can change:
+
+- The inputs/arguments and timeout values of Child Workflows, Activities, and Signals to external workflows
```
tomwheeler: This omits mention of the outputs for Child Workflows and Activities, which I think would be as notable as the inputs. Also, I don't think timeouts are relevant for Signals here, so I'd cover that in a separate bullet point. 

Consider replacing lines 165-166 with this:

> - The input parameters, return values, and execution timeouts of Child Workflows and Activities
>    - However, it is not safe to change the types or IDs of Child Workflows or Activities
> - The input parameters used to Signal an external Workflow

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+  - But not their type, or id!
+- The duration of timers (though changing them to 0 is not safe in all SDKs)
```
tomwheeler:
```suggestion
- The duration of Timers (although changing them to 0 is not safe in all SDKs)
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Add or remove calls to Workflow APIs that don't produce Commands (ex: get Workflow Info)
```
tomwheeler: I know what you mean by "get Workflow Info," but since that's not the actual name in any SDK (AFAIK), I think it would be better to use a specific method as the example here.

```suggestion
- Add or remove calls to Workflow APIs that don't produce Commands (e.g., `workflow.GetInfo` in the Go SDK or its equivalent in other SDKs)
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+
+The following Workflow API calls all can produce commands, and thus must not be reordered, added, or
+removed without proper [versioning techniques](#workflow-versioning):
```
tomwheeler:
```suggestion
removed without proper [Versioning techniques](#workflow-versioning):
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+makes the same Workflow API calls in the same sequence, given the same input. Some changes to those
+
+API calls are safe to make. For example, you can change:
+
+- The inputs/arguments and timeout values of Child Workflows, Activities, and Signals to external workflows
+  - But not their type, or id!
+- The duration of timers (though changing them to 0 is not safe in all SDKs)
+- Add or remove calls to Workflow APIs that don't produce Commands (ex: get Workflow Info)
+
+The following Workflow API calls all can produce commands, and thus must not be reordered, added, or
```
tomwheeler:
```suggestion
The following Workflow API calls all can produce Commands, and thus must not be reordered, added, or
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+removed without proper [versioning techniques](#workflow-versioning):
+
+- Starting or cancelling a Timer
+- Scheduling or cancelling Activity executions (including local Activities)
```
tomwheeler:
```suggestion
- Scheduling or cancelling Activity Executions (including local Activities)
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Starting or cancelling Child Workflow executions
```
tomwheeler, line 175:
```suggestion
- Starting or cancelling Child Workflow Executions
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Signalling or cancelling signals to external Workflow executions
```
tomwheeler:
```suggestion
- Signalling or cancelling signals to external Workflow Executions
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Scheduling or cancelling Nexus operations
+- Ending the Workflow Execution in any way (completing, failing, cancelling, or continuing as new)
```
tomwheeler:
```suggestion
- Ending the Workflow Execution in any way (completing, failing, cancelling, or continuing-as-new)
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Patched or GetVersion calls for versioning (though they may be added or removed according to the [patching](#patching) rules)
```
tomwheeler:
```suggestion
- `Patched` or `GetVersion` calls for Versioning (although they may be added or removed according to the [patching](#patching) rules)
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Upserting Workflow search attributes
```
tomwheeler:
```suggestion
- Upserting Workflow Search Attributes
```

docs/encyclopedia/workflow/workflow-definition.mdx
tomwheeler:
```suggestion
- Upserting Workflow Memos
```

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,36 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Upserting Workflow memos
+
+More formally, the use of certain Workflow APIs in the function is what generates
+[Commands](/workflow-execution#command). Commands tell the Temporal Service which Events to create
+and add to the Workflow Execution's [Event History](/workflow-execution/event#event-history). When a
```
tomwheeler: I think the mention of "Workflow Function" here could be confusing, partly because of the formality implied by the uppercase F and partly because the term isn't consistent for all SDKs (such as Java, where the Workflow is defined in a _method_). Using "Workflow Definition" instead of "Workflow Function" would alleviate that, but that also sounds weird to me.

So, instead of this:

> When a Workflow Function [replays](/workflow-execution#replay), the Commands that are emitted are compared
with the existing Event History.

Consider this:

> When the Workflow's code [replays](/workflow-execution#replay), the Commands that are emitted are compared
with the existing Event History.

(Or alternatively, "When the Worker [replays](/workflow-execution#replay) code in the Workflow Definition, the Commands ...")

tomwheeler (commented):
I left some comments and suggestions, mostly capitalization for conformance with house style, but a few others that I believe would clarify the documentation. In any case, I think this is much better explained than what's currently on docs.temporal.io.

docs/encyclopedia/workflow/workflow-definition.mdx
Sushisource: Yeah, I agree. I was trying to get rid of some of these old overly formal terms we insisted on having in the past but I didn't know what the feeling was on that now. Happy to do something simpler.

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,37 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+A critical aspect of developing Workflow Definitions is ensuring they are deterministic. Generally
+speaking, this means you must take care to ensure that any time your Workflow code is executed it
+makes the same Workflow API calls in the same sequence, given the same input. Some changes to those
+API calls are safe to make. For example, you can change:
+
+- The input parameters, return values, and execution timeouts of Child Workflows and Activities
+  - However, it is not safe to change the types or IDs of Child Workflows or Activities
+- The input parameters used to Signal an external Workflow
+- The duration of Timers (although changing them to 0 is not safe in all SDKs)
+- Add or remove calls to Workflow APIs that don't produce Commands (e.g., `workflow.GetInfo` in the Go SDK or its equivalent in other SDKs)
```
drewhoskins-temporal: Commands is un-introduced at this point.  Amounts to "know what's on this list that we haven't given you."

Not having brilliant ideas to fix, maybe say "certain read-only workflow APIs?"

If you keep the Commands concept, link out to the definition.

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,37 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Signalling or cancelling signals to external Workflow Executions
+- Scheduling or cancelling Nexus operations
+- Ending the Workflow Execution in any way (completing, failing, cancelling, or continuing-as-new)
+- `Patched` or `GetVersion` calls for Versioning (although they may be added or removed according to the [patching](#patching) rules)
+- Upserting Workflow Search Attributes
+- Upserting Workflow Memos
+- Running a SideEffect or MutableSideEffect
+
+More formally, the use of certain Workflow APIs in the function is what generates
+[Commands](/workflow-execution#command). Commands tell the Temporal Service which Events to create
```
drewhoskins-temporal: As mentioned above, this is the first link but third mention.

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,37 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
-
-The execution semantics of a Workflow Execution include the re-execution of a Workflow Function, which is called a [Replay](/workflow-execution#replay).
-The use of Workflow APIs in the function is what generates [Commands](/workflow-execution#command).
-Commands tell the Temporal Service which Events to create and add to the Workflow Execution's [Event History](/workflow-execution/event#event-history).
-When a Workflow Function executes, the Commands that are emitted are compared with the existing Event History.
-If a corresponding Event already exists within the Event History that maps to the generation of that Command in the same sequence, and some specific metadata of that Command matches with some specific metadata of the Event, then the Function Execution progresses.
+A critical aspect of developing Workflow Definitions is ensuring they are deterministic. Generally
+speaking, this means you must take care to ensure that any time your Workflow code is executed it
+makes the same Workflow API calls in the same sequence, given the same input. Some changes to those
```
drewhoskins-temporal: How about a paragraph at the top that's something like "The two most common sources of non-determinism are when workflows interact with data sources without memoizing their results, and when users deploy new code that gets executed by an existing workflow."

And then get into the details.

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,37 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- The input parameters, return values, and execution timeouts of Child Workflows and Activities
+  - However, it is not safe to change the types or IDs of Child Workflows or Activities
+- The input parameters used to Signal an external Workflow
+- The duration of Timers (although changing them to 0 is not safe in all SDKs)
+- Add or remove calls to Workflow APIs that don't produce Commands (e.g., `workflow.GetInfo` in the Go SDK or its equivalent in other SDKs)
+
+The following Workflow API calls all can produce Commands, and thus must not be reordered, added, or
+removed without proper [Versioning techniques](#workflow-versioning):
+
+- Starting or cancelling a Timer
```
drewhoskins-temporal, line 173: We have https://docs.temporal.io/references/commands so this is somewhat duplicative.  (though this is more concisely readable).  Wonder if we should consolidate.

docs/encyclopedia/workflow/workflow-definition.mdx
```diff
@@ -156,13 +156,37 @@ We strongly recommend that you write a Workflow Definition in a language that ha
...
+- Upserting Workflow Search Attributes
+- Upserting Workflow Memos
+- Running a SideEffect or MutableSideEffect
+
+More formally, the use of certain Workflow APIs in the function is what generates
+[Commands](/workflow-execution#command). Commands tell the Temporal Service which Events to create
+and add to the Workflow Execution's [Event History](/workflow-execution/event#event-history). When
+the Workflow's code [replays](/workflow-execution#replay), the Commands that are emitted are
+compared with the existing Event History. If a corresponding Event already exists within the Event
+History that matches that command, then the Execution progresses.
```
angelazhou32: Not sure if you'd like to link to this for a detailed walkthrough of how this works: https://docs.temporal.io/encyclopedia/event-history/

docs/encyclopedia/workflow/workflow-definition.mdx
Sushisource: I think we need to have the concept generally still. Added the link.

docs/encyclopedia/workflow/workflow-definition.mdx
Sushisource: ~I can add something about the two sources. Users don't think of Temporal as "memoizing" though, even if that is what it does.~

Actually, this is addressed at the end of the section and I'm not sure how much it helps earlier before going over the concepts.

docs/encyclopedia/workflow/workflow-definition.mdx
Sushisource: Hmm... good point. I think both formats serve a purpose. I'll link to there from here.

docs/encyclopedia/workflow/workflow-definition.mdx
drewhoskins-temporal: I think it helps to have a "when/why I should care"/"give me a concrete example" section at the top to help orient folks.  But up to you.

drewhoskins-temporal (approved):
lgtm, one optional thought.

PR #3620, opened 2025-06-14
Java - Add docs for Nexus cancellation type

Quinn-With-Two-Ns:
## What does this PR do?

Java - Add docs for Nexus cancellation type

## Notes to reviewers

Currently this is in draft since the Java SDK is not released yet


brianmacdonald-temporal (approved):
Changes look good

PR #3623, opened 2025-06-18
added 2 year default to the two notes on the `Generate API Key` button

GSmithApps:
## What does this PR do?

Adds a note that when the `Generate API Key` button in the UI is clicked, the expiration time is the default max, which is 2 years. 

## Notes to reviewers

This is in response to a customer ticket


brianmacdonald-temporal (approved):
Looks good to me.

PR #3636, opened 2025-06-30
add link for workflowcheck README in Go SDK landing page

yuandrew:
## What does this PR do?

workflowcheck is only mentioned in on a page in learn.temporal.io, for now add a link to the tool's README.

### DIFF
--- a/docs/develop/go/index.mdx
+++ b/docs/develop/go/index.mdx
@@ -65,6 +65,12 @@ Set up the testing suite and test Workflows and Activities.
 - [Testing Workflows](/develop/go/testing-suite#test-workflows)
 - [How to Replay a Workflow Execution](/develop/go/testing-suite#replay)
 
+## [workflowcheck Determinism Checker]
+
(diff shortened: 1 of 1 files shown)


axfelix:
I think I'd rather see this on the Go Versioning or Testing page, both of which are under heavy development from the Worker Versioning launch, so sit tight on this please! The landing pages don't get a ton of attention and it would be good to give this more prominence.

jsundai:
@axfelix Perhaps, this could go under the Workflow Definitions page in the Encyclopedia?

axfelix:
It's Go specific, so I think it probably makes sense to keep it under Go -- it might actually deserve its own page like we have for Python async stuff. The only 2 pages in the Go dev guide I'm aware of right now that are totally unique to Go are Selectors and Multithreading, and I can see *maybe* adding it to multithreading as an example of "here's how to do things safely with Temporal in Go" because I think that's something the tool would catch.

axfelix:
Closing this because https://github.com/temporalio/documentation/pull/3644 was merged instead!

PR #3641, opened 2025-07-01
Add some docs on eager tasks

Sushisource:
## What does this PR do?
Added some docs on eager activities / workflows

### DIFF
--- a/docs/develop/worker-performance.mdx
+++ b/docs/develop/worker-performance.mdx
@@ -101,6 +101,14 @@ When a Task Poller receives a Task, it delivers to the appropriate Executor Slot
 Task Pollers enable efficient load balancing across multiple Worker processes.
 The number of Task Pollers can be configured using `WorkerOptions` when creating a new Worker instance.
 
+### Eager task execution
+
+Workers may eagerly execute activity and workflow tasks under the right circumstances.
+
+Eager Activity execution may happen automatically if the Worker processing a Workflow Task also has the Activity definition being called registered. If it does, it may try to reserve an Activity Slot for the execution of the Activity, and the server may respond to the Workflow Task completion with the Activity Task for the worker to execute immediately.
+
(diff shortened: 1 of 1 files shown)


docs/develop/worker-performance.mdx
```diff
@@ -101,6 +101,14 @@ When a Task Poller receives a Task, it delivers to the appropriate Executor Slot
...
 
+### Eager task execution
+
+Workers may eagerly execute activity and workflow tasks under the right circumstances.
+
+Eager Activity execution may happen automatically if the Worker processing a Workflow Task also has the Activity definition being called registered. If it does, it may try to reserve an Activity Slot for the execution of the Activity, and the server may respond to the Workflow Task completion with the Activity Task for the worker to execute immediately.
+
+Eager Workflow execution is opt-in, and requires the Client which is starting the Workflow to be located in the same process as a Worker. When making the Start Workflow call, you can set the `request_eager_start` (or similar name) to true. If it is, and the Worker has a Workflow Task slot available and the Workflow Definition registered, the Worker can execute the first task of the Workflow locally without first making a round-trip to the Temporal Server. This is typically most useful in combination with a Local Activity executing in the first Workflow Task, since other Workflow API calls that require waiting on something will force a round-trip.
```
bergundy:
```suggestion
Eager Workflow execution is opt-in, and requires the Client which is starting the Workflow to be located in the same process as a Worker. When making the Start Workflow call, you can set the `request_eager_start` (or similar name) to true. When set, and the Worker has a Workflow Task slot available and the Workflow Definition registered, the Worker can execute the first task of the Workflow locally without first making a round-trip to the Temporal Server. This is typically most useful in combination with a Local Activity executing in the first Workflow Task, since other Workflow API calls that require waiting on something will force a round-trip.
```

bergundy (approved):

jsundai (approved):

PR #3653, opened 2025-07-08
update documentation for activity pause,unpause,reset

spkane31:
## Notes to reviewers

Update the documentation for `temporal activity {pause,unpause,reset}` commands to include missing flags and requirements.


docs/cli/activity.mdx
```diff
@@ -394,16 +394,24 @@ Use the following options to change the behavior of this command.
 
 **--activity-id**, **-a** _string_
 
-Activity ID to pause.
+Activity ID to pause. Either `activity-id` or `activity-type` must be provided, but not both.
```
ychebotarev, line 397: there is also match-all flag for pause/unpause. It also mutually exclusive with activity id and activity type.
It works like "pause/unpuase whatever is currently running"

spkane31: `--match-all` does not exist on `pause`, will update the docs for `unpause`

PR #3662, opened 2025-07-09
Propagate Versioning edits to Ruby and PHP

axfelix:
Also did a bit of style tidying on the others

### DIFF
--- a/docs/develop/ruby/versioning.mdx
+++ b/docs/develop/ruby/versioning.mdx
@@ -24,41 +24,65 @@ tags:
   - Patching
 ---
 
-This page shows how to do the following:
+import { CaptionedImage } from '@site/src/components';
 
-- [Use the Ruby SDK patching API](#ruby-sdk-patching-api)
-- [Patch in new code safely](#patching-in-new-code)
-- [Understand deprecated patches](#deprecated-patches)
-- [Safely deploy `post_patch_activity`](#safe-deployment-of-postpatchactivity)
+Since Workflow Executions in Temporal can run for long periods — sometimes months or even years — it's common to need to make changes to a Workflow Definition, even while a particular Workflow Execution is in progress.
 
-## Introduction to Versioning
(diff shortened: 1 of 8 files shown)


axfelix:
It may make sense to hold this for a bit since I don't know what the current support expectations are for WV in Ruby and PHP (there *are* code samples for Ruby on the WV page so I assume this is planned).

docs/develop/ruby/versioning.mdx
```diff
@@ -24,41 +24,65 @@ tags:
   - Patching
 ---
 
-This page shows how to do the following:
+import { CaptionedImage } from '@site/src/components';
```
angelazhou32: Doesnt seem to be used

docs/develop/ruby/versioning.mdx
```diff
@@ -88,21 +112,21 @@ class MyWorkflow < Temporalio::Workflow::Definition
 end
 ```
 
-**Problem: You cannot deploy `PostPatchActivity` directly until you're certain there are no more running Workflows created using the `PrePatchActivity` code, otherwise you are likely to cause a nondeterminism error.**
-
-Instead, you'll need to deploy `PostPatchActivity` and use the [patched](https://ruby.temporal.io/Temporalio/Workflow.html#patched-class_method) method to determine which version of the code to execute.
+The problem is that you cannot deploy this new revision directly until you're certain there are no more running Workflows created using the `PrePatchActivity` code, otherwise you are likely to cause a nondeterminism error.
+Instead, you'll need to use the [`patched`](https://ruby.temporal.io/Temporalio/Workflow.html#deprecate_patch-class_method) function to check which version of the code should be executed.
```
angelazhou32: This link instead? https://ruby.temporal.io/Temporalio/Workflow.html#patched-class_method

axfelix: Great catch

angelazhou32 (approved):
lgtm

PR #3668, opened 2025-07-10
Improve value prop for Worker Versioning

drewhoskins-temporal:
## What does this PR do?
* Move the value proposition first, above the description of types of deployments.
* @axfelix recommended talking up the benefits for blue/green deploys, which is crystallized here.  This will hopefully increase adoption among those who don't want to move to Rainbow Deploys.
* Describe types of deployments more systematically.
* Recommend CLI v1.4.0, which has breaking changes from v1.3
* Try out capitalized term of art "Workflow Pinning"

## Notes to reviewers
Test plan
yarn start
http://localhost:3000/production-deployment/worker-deployments/worker-versioning

Click all the links.

### DIFF
--- a/docs/production-deployment/worker-deployments/worker-versioning.mdx
+++ b/docs/production-deployment/worker-deployments/worker-versioning.mdx
@@ -68,15 +68,17 @@ Self-hosted users:
 
 To get started with Worker Versioning, you should understand some concepts around versioning and deployments.
 
-- A **Worker Deployment** is a deployment or service across multiple versions. In a rainbow deployment, multiple active Deployment Versions can run at once.
-- A **Worker Deployment Version** is a version of a deployment or service. It can have multiple Workers, but they all run the same build.
+- A **Worker Deployment** is a deployment or service across multiple versions. In a rainbow deployment, more than two active Deployment Versions can run at once.
+- A **Worker Deployment Version** is a version of a deployment or service. It can have multiple Workers polling on multiple Task Queues, but they all run the same build.
 - A **Build ID**, in combination with a Worker Deployment name, identifies a single Worker Deployment Version.
+- When a versioned worker polls on a task queue, that task queue becomes part of that Worker's version. That version's Worker Deployment controls how the task queue matches Workflow Tasks with Workers.
 - Using **Workflow Pinning**, you can declare each Workflow type to have a **Versioning Behavior**, either Pinned or Auto-Upgrade.
-  - A Pinned Workflow is guaranteed to complete on a single Worker Deployment Version.
-  - An Auto-Upgrade Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows won't be restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
+  - A **Pinned** Workflow is guaranteed to complete on a single Worker Deployment Version.
+  - An **Auto-Upgrade** Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows are not restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
+  - Both Pinned and Auto-Upgrade Workflows are guaranteed to start only on the Current or Ramping Version of their Worker Deployment.
   - Workflow Pinning is designed for use with rainbow deployments. See [Deployment Systems](#deployment-systems).
-- Each Worker Deployment has a single **Current Version** which is where workflows are routed to unless they were previously pinned on a different version. Other versions are still around to allow pinned Workflows to finish executing, or in case you need to roll back. If no current version is specified, the default is unversioned.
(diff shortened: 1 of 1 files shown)


docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -58,25 +59,35 @@ Minimum versions:
...
 - Each Worker Deployment has a single **Current Version** which is where workflows are routed to unless they were previously pinned on a different version. Other versions are still around to allow pinned Workflows to finish executing, or in case you need to roll back. If no current version is specified, the default is unversioned.
 - Each Worker Deployment can have a **Ramping Version** which is where a configurable percentage of workflows are routed to unless they were previously pinned on a different version. The ramp percentage can be in the range [0, 100]. Workflows that don't go to the Ramping Version will go to the Current Version. If no ramping version is specified, 100% of auto-upgrade workflows will go to the Current Version.
 
-Your deployment system should support rainbow deployments so that you can keep versions alive to drain existing workflows.
+Worker Versioning requires that you are using blue-green deployments.
+Using Workflow Pinning requires rainbow deployments.
```
axfelix: I think this part will be confusing for a first time reader for a couple reasons:
- How is using Workflow Pinning different from adopting Worker Versioning in the first place? imo, we haven't made this clear enough, this makes it sound like it's a subset of the feature when in practice it's kind of *the whole feature* right now. It comes across like you can adopt Worker Versioning in a staggered or half way which doesn't really achieve anything.
- I don't think there is a large audience of users who are all-in on two and only two (ie blue and green) deployments -- hence I'm not sure that it's helpful to separate the framing of blue-green and rainbow in this way. I would rather frame it as "Worker Versioning means *not* using rolling deployments, this supports blue-green deploys, or any other strategy that makes use of rainbow deployments."

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -58,25 +59,35 @@ Minimum versions:
...
+- Minimum Temporal CLI version v1.4.0
 - Minimum Temporal Server version: v1.27.1
 - Minimum Temporal UI Version v2.36.0
   :::
 
+## Three deployment strategies {#deployment-strategies}
+
+When preparing your Workers for production, you'll choose among three common deployment strategies:
+
+- A **rolling deployment** strategy upgrades Workers in place with little control over how quickly they cut over and only a slow ability to roll Workers back. We do not recommend rolling deploys for high-availability Worker services.
```
axfelix: I like the first sentence (I think it's descriptive and sufficiently negative to motivate adopting Worker Versioning). For the second sentence, instead of "We do not recommend rolling deploys", can we say something like "Without Worker Versioning and Workflow Pinning, rolling deploys are the default with Temporal. By adopting Worker Versioning, you get more fine-grained control." This way, instead of saying "this is bad," we're making clear what this new feature adds.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -58,25 +59,35 @@ Minimum versions:
 
 Self-hosted users:
 
-- Minimum Temporal CLI version v1.3.0
+- Minimum Temporal CLI version v1.4.0
 - Minimum Temporal Server version: v1.27.1
 - Minimum Temporal UI Version v2.36.0
```
carlydf:
```suggestion
- Minimum Temporal CLI version [v1.4.0](https://github.com/temporalio/cli/releases/tag/v1.4.0)
- Minimum Temporal Server version: [v1.28.0](https://github.com/temporalio/temporal/releases/tag/v1.28.0)
- Minimum Temporal UI Version [v2.38.0](https://github.com/temporalio/ui/releases/tag/v2.38.0)
```

carlydf (approved):
approved pending my mininum server version suggested change!

docs/production-deployment/worker-deployments/worker-versioning.mdx
drewhoskins-temporal: > I don't think there is a large audience of users who are all-in on two and only two (ie blue and green) deployments 

You might be surprised.  Blue/green has like 10x mindshare to rainbow based on my customer interviews.  And one customer explicitly told me they had to stick to b/g because their managed deployment stack is oriented around it.

However, we do want to nudge people toward rainbow since it's more effective.  I'll wordsmith this.

drewhoskins-temporal: > It comes across like you can adopt Worker Versioning in a staggered or half way which doesn't really achieve anything.

You _can_ adopt Worker Versioning without pinning workflows.  It accomplishes the three goals I set up at the top of the document.

axfelix (approved):
👍 let's do it!

docs/production-deployment/worker-deployments/worker-versioning.mdx
axfelix: I'm not convinced this won't prove confusing (this was why I was enthusiastic to see this whole workstream and related features grouped under "safe deploys" instead), but I'm happy with the changes for now. 👍👍

PR #3679, opened 2025-07-11
[Sdk-Go] warn users not to register struct methods as workflows

yuandrew:
## What does this PR do?

This is to address https://github.com/temporalio/sdk-go/issues/1986, the team has decided adding a note in the docs is more suitable than adding a warning log in the SDK itself.

## Notes to reviewers

Linked work item and closed SDK PR should have all of the context for this


docs/develop/go/core-application.mdx
```diff
@@ -409,6 +409,14 @@ Activities written as struct methods can use shared struct variables, such as:
```
yuandrew, line 409: In the paragraph above starting with "In the Temporal Go SDK programming model," (GH won't let me add a comment that high up), I noticed "struct" seems to be punctuated inconsistently, that paragraph alone has `struct`, Struct, and struct in different sentences. Not sure if that's something we want to make consistent

Quinn-With-Two-Ns (approved):

axfelix (approved):

PR #3686, opened 2025-07-15
Add Python interceptors documentation

dandavison:


docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+tags:
+  - Interceptors
+  - Python SDK
+  - Temporal SDKs
+---
+
+### Interceptors
+
+The behavior of the Python SDK can be customized in many useful ways by modifying inbound and outbound calls using
+interceptors. This is similar to the use of middleware in other frameworks.
```
axfelix: We usually do one sentence per line — won't fuss with it here, but an FYI if you are porting this to other SDKs after this (which I would strongly encourage)!

More pertinent — I'm not sure that saying that Interceptors are "similar to the use of middleware in other frameworks" is very descriptive. Can we analogize to adding hooks instead?

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+There are five categories of inbound and outbound calls that you can modify in this way:
+
+1. Outbound Client calls, such as `start_workflow()`, `signal_workflow()`, `list_workflows()`, `update_schedule()`, etc.
```
axfelix: These categories are not widely used elsewhere in our documentation or other educational materials, so I don't think these lists or the use of "etc." are sufficient for users to understand them...

How's this format?

```suggestion
1. Outbound calls made by a Temporal client:
  - `start_workflow()`
  - `signal_workflow()`
  - `list_workflows()`
  - `update_schedule()`
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+2. Inbound Workflow calls: `execute_workflow()`, `handle_signal()`, `handle_update_handler()`, etc
+
+3. Outbound Workflow calls: `start_activity()`, `start_child_workflow()`, `start_nexus_operation()`, etc
+
+4. Inbound call to execute an Activity: `execute_activity()`
+
+5. Outbound Activity calls: `info()` and `heartbeat()`
+
+To modify outbound Client calls, define a class inheriting from
```
axfelix: Since this deals with outbound Client calls (the first category out of 5), we should probably pull this up to right after the list of outbound client calls above (and maybe make these 5 categories each different H3s or H4s rather than a list).

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+[`client.Interceptor`](https://python.temporal.io/temporalio.client.Interceptor.html), and implement the method
+`intercept_client()` to return an instance of
+[`OutboundInterceptor`](https://python.temporal.io/temporalio.client.OutboundInterceptor.html) that implements the
+subset of outbound Client calls that you wish to modify.
+
+Then, pass a list containing an instance of your `client.Interceptor` class as the
```
axfelix: Need to see a sample here imo — talking about class instances in the abstract is not helpful. I suspect this is why we haven't had progress on these docs :)

docs/develop/python/interceptors.mdx
dandavison: Sorry, I'll change to one-sentence per line. I copied this from the README and forgot about that.

> More pertinent — I'm not sure that saying that Interceptors are "similar to the use of middleware in other frameworks" is very descriptive.

Hm -- are you sure? Middleware is the term used by Django and Flask I know:

https://docs.djangoproject.com/en/5.2/topics/http/middleware/
https://flask.palletsprojects.com/en/stable/lifecycle/#middleware

and for ones I didn't know what term they use I'm also seeing it used

https://www.starlette.io/middleware/

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+`interceptors` argument of [`Client.connect()`](https://python.temporal.io/temporalio.client.Client.html#connect).
+
+The purpose of the interceptor framework is that the methods you implement on your interceptor classes can perform
```
axfelix:
```suggestion
The methods you implement on your interceptor classes can perform
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+arbitrary side effects and/or arbitrary modifications to the data, before it is received by the SDK's "real"
```
axfelix: Is "data" what we mean here? I think it'd be best to focus on the side effects case first, otherwise someone might get mixed up about these being primarily useful for data transformations a la data converters.

docs/develop/python/interceptors.mdx
dandavison: I don't think that edit is an improvement -- it loses the sense that this is not just something they _can_ do, but rather that it's the entire point of the existence of the framework.

axfelix: Then let's restate it for emphasis in the following sentence (eg "This is the purpose of the Interceptor framework.") — this was mostly a readability edit to lose that first clause.

docs/develop/python/interceptors.mdx
axfelix: Ah, OK, it seems like it's accurate to how "middleware" is used in Python frameworks, but not to my understanding of "middleware" as a generic concept. Might need to use those specific examples here for clarity and make a note to change this when porting.

docs/develop/python/interceptors.mdx
dandavison: I think the snippet at the bottom shows gives a good intro. You're right that it's a tough area to explain in words, and it's also a more advanced topic -- they'll need to click through to the samples to really be able to apply it.

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+The remaining four categories are Worker calls. To modify these, define a class inheriting from
+[`worker.Interceptor`](https://python.temporal.io/temporalio.worker.Interceptor.html) and implement methods on that
+class to define the
+[`ActivityInboundInterceptor`](https://python.temporal.io/temporalio.worker.ActivityInboundInterceptor.html),
+[`ActivityOutboundInterceptor`](https://python.temporal.io/temporalio.worker.ActivityOutboundInterceptor.html),
+[`WorkflowInboundInterceptor`](https://python.temporal.io/temporalio.worker.WorkflowInboundInterceptor.html), and
+[`WorkflowOutboundInterceptor`](https://python.temporal.io/temporalio.worker.WorkflowOutboundInterceptor.html) classes
+that you wish to use to effect your modifications. Then, pass a list containing an instance of your `worker.Interceptor`
+class as the `interceptors` argument of the [`Worker()`](https://python.temporal.io/temporalio.worker.Worker.html)
+constructor.
```
axfelix: An inline sample would be great here too!

docs/develop/python/interceptors.mdx
dandavison: The data modifications are important, and they're what the sample that I'm using demonstrates. I believe the context propagation samples was chosen and implemented in multiple languages because it's a typical use case.

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+It often happens that your Worker and Client interceptors will share code because they implement closely related logic.
+For convenience, you can create an interceptor class that inherits from _both_ `client.Interceptor` and
+`worker.Interceptor` (their method sets do not overlap). You can then pass this in the `interceptors` argument of
+`Client.connect()` when starting your Worker _as well as_ in your client/starter code. If you do this, your Worker will
+automatically pick up the interceptors from its underlying Client (and you should not pass them directly to the
+`Worker()` constructor).
```
axfelix: The use of parenthesis here is going to be hard for a novice user to follow. I'd like to see inline code samples here too, and then maybe an admonition (like the note on https://docs.temporal.io/develop/python/cancellation#cancel-activity) to make the clarification in parenthesis clearer and more emphatic.

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+This is best explained by example. The [Context Propagation Interceptor Sample](https://github.com/temporalio/samples-python/tree/main/context_propagation) is a good starting point. In
```
axfelix: Good — but we generally don't require users to click through to samples repos from docs. I would remove the first 2 sentences here, move the code sample ahead of this explanation, and then move the rest of this paragraph, starting with "In this example, a class is defined..." to below the sample.

axfelix:
Added to sidebar :)

Really excited to have this content added to docs at last. Right now I think it does a little bit too much abstract explanation of the concept (ie without inline samples) and leans too heavily on the samples repo rather than using smaller inline samples, but I think it's 80% of the way there and can be used as a model for the other SDKs once this is done.

docs/develop/python/interceptors.mdx
axfelix: Hm, OK. I think I might still prefer to see the emphasis start clearer, something like:
- Interceptors can be used for arbitrary side effects
- these can include data modifications, e.g. for context propagation

To keep this focused. Mind if I have a try at rewriting this paragraph?

axfelix:
I would recommend not force pushing, btw — looks like you may have lost the sidebar addition.

docs/develop/python/interceptors.mdx
dandavison: Right, this is an advanced topic, and those categories are being defined here. What I've done now is linked each one to its respective interceptor, and that page contains the complete list of calls being referred to. I've selected the subset here so that a sparky reader will get it ("Hm, start_activity, start_child_workflow, start_nexus_operation... Oh OK, I get it, all the things I can do from a workflow"). But in the cases where there are too many to list I've used "etc" to indicate that the list is incomplete. I've also added  a link to https://docs.temporal.io/encyclopedia/temporal-sdks#temporal-client.

docs/develop/python/interceptors.mdx
dandavison: I prefer introducing the 5 categories without breaking up their flow with explanatory detail. I know from experience (I wrote this documentation because I found I didn't have good understanding of the subject) that what made it start to be clear was to see this taxonomy, so that there weren't so many unanchored concepts and class names floating around.

axfelix: Having a taxonomy is great, but I wonder if we need some encyclopedia page additions to deep link in that case. Mostly I just want to see these categories better substantiated if we're just doing a conceptual gloss here, because we don't draw much attention to inbound vs outbound calls anyplace else.

docs/develop/python/interceptors.mdx
dandavison: Well, OK, I've just gone with your original sentence. I agree it's not essential to reinforce that this is the purpose of the framework, and I couldn't see an elegant way to do it other than the way I had.

docs/develop/python/interceptors.mdx
dandavison: I struggled a lot with this parenthetic comment! Be my guest and improve this para!

docs/develop/python/interceptors.mdx
dandavison: OK, I copied some of the code across, but I'd personally strongly vote for leaving the link to the sample in the docs -- seeing a full sample is going to be essential to anyone wanting to implement their own interceptors.

docs/develop/python/interceptors.mdx
dandavison: Yes, please go ahead and rewrite.

docs/develop/python/interceptors.mdx
dandavison: I added the links to the three Python web frameworks above

axfelix:
It does help to fix indentation and such, and clobbering work is no fun — please pull before pushing in the future, this doesn't have to happen ASAP.

docs/develop/python/interceptors.mdx
dandavison: That could be good. This will be obvious to you but just for the record for anyone writing encyclopedia content let me just say that talking of "calls" in all cases is pretty questionable. In the case of a workflow these are "commands" of course; a workflow is not allowed to make a network call, and that it "calls" an activity is obviously an illusion that we create, not an accurate understanding!

axfelix:
@dandavison what do you think of the most recent commit changes?

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+  - Interceptors
+  - Python SDK
+  - Temporal SDKs
+---
+
+The behavior of the Python SDK can be customized in many useful ways by modifying inbound and outbound calls using Interceptors.
+This is similar to the use of "middleware" in web frameworks such as [Django](https://docs.djangoproject.com/en/5.2/topics/http/middleware/), [Starlette](https://www.starlette.io/middleware/), and [Flask](https://flask.palletsprojects.com/en/stable/lifecycle/#middleware).
+
+The methods you implement on your Interceptor classes can perform arbitrary side effects.
+Interceptors can also perform arbitrary modifications to data before it is received by the SDK's "real" implementation, e.g. for context propagation.
```
dandavison: I think "e.g. for context propagation" will go over a lot of readers' heads and so we shouldn't do that so early on in the doc.

So, just a suggestion, but:

```suggestion
Interceptors can also perform arbitrary modifications to incoming and outgoing data before it is received by the SDK's "real" implementation.
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+- `signal_workflow()`
+- `list_workflows()`
+- `update_schedule()`
+
+This is not an exhaustive list; refer to the [Python SDK methods](https://python.temporal.io/temporalio.client.OutboundInterceptor.html) for more.
+
+#### [Inbound Workflow calls](https://python.temporal.io/temporalio.worker.WorkflowInboundInterceptor.html)
+
+- `execute_workflow()`
+  - Note that `execute_workflow()` is considered an Inbound Workflow call unlike `start_workflow()` because it blocks and waits for a result to be returned.
```
dandavison: `execute_workflow` here is not really related to the `start_workflow` vs `execute_workflow` distinction in the client API. `execute_workflow` is just the name given to the task that a workflow has to do when it is told to execute a workflow.

It's like "handle a request to execute user workflow code".

Mentioning the client methods will confuse readers. Could maybe be something like


```suggestion
- `execute_workflow()`
  - I.e. handle a Workflow Task that is starting a new Workflow Execution.
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+- `start_child_workflow()`
+- `signal_child_workflow()`
+- `signal_external_workflow()`
+- `start_nexus_operation()`
+- `start_local_activity()`
+
+#### [Inbound Activity calls](https://python.temporal.io/temporalio.worker.ActivityInboundInterceptor.html)
+
+- `execute_activity()`
+  - This is technically the only Inbound Activity call, with the same distinction as `execute_workflow()`.
```
dandavison:
```suggestion
- `execute_activity()`
  - I.e. handle a task to execute an Activity (this is the only Inbound Activity call)
```

dandavison: (leaving parens or sub-bullet up to you)

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+#### [Outbound Activity calls](https://python.temporal.io/temporalio.worker.ActivityOutboundInterceptor.html)
+
+- `info()`
+- `heartbeat()`
+
+Technically, the first of these categories is a Client call, and the remaining 4 are Worker calls.
+It often happens that your Worker and Client interceptors will share code because they implement closely related logic.
+In the Python SDK, you will typically want to create an interceptor class that inherits from _both_ `client.Interceptor` and `worker.Interceptor`, since their method sets do not overlap.
```
dandavison: We haven't introduced anything about interceptor classes, so I think this sentence has to come later.

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+- `execute_activity()`
+  - This is technically the only Inbound Activity call, with the same distinction as `execute_workflow()`.
+
+#### [Outbound Activity calls](https://python.temporal.io/temporalio.worker.ActivityOutboundInterceptor.html)
+
+- `info()`
+- `heartbeat()`
+
+Technically, the first of these categories is a Client call, and the remaining 4 are Worker calls.
```
dandavison:
```suggestion
The first of these categories is a Client call, and the remaining 4 are Worker calls.
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+It often happens that your Worker and Client interceptors will share code because they implement closely related logic.
+In the Python SDK, you will typically want to create an interceptor class that inherits from _both_ `client.Interceptor` and `worker.Interceptor`, since their method sets do not overlap.
+
+## Client call Interceptors
+
+To modify outbound Client calls, define a class inheriting from [`client.Interceptor`](https://python.temporal.io/temporalio.client.Interceptor.html), and implement the method `intercept_client()` to return an instance of [`OutboundInterceptor`](https://python.temporal.io/temporalio.client.OutboundInterceptor.html) that implements the subset of outbound Client calls that you wish to modify:
+
```
dandavison, line 67: How about we move this sentence:

> This example implements an Interceptor on outbound Client calls that sets a certain key in the outbound `headers` field.
A User ID is context-propagated by being sent in a header field with outbound requests.

to here, so that they understand that the code they're about to look at is an example illustating a specific application of interceptors?

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+- `info()`
+- `heartbeat()`
+
+Technically, the first of these categories is a Client call, and the remaining 4 are Worker calls.
+It often happens that your Worker and Client interceptors will share code because they implement closely related logic.
+In the Python SDK, you will typically want to create an interceptor class that inherits from _both_ `client.Interceptor` and `worker.Interceptor`, since their method sets do not overlap.
+
+## Client call Interceptors
+
+To modify outbound Client calls, define a class inheriting from [`client.Interceptor`](https://python.temporal.io/temporalio.client.Interceptor.html), and implement the method `intercept_client()` to return an instance of [`OutboundInterceptor`](https://python.temporal.io/temporalio.client.OutboundInterceptor.html) that implements the subset of outbound Client calls that you wish to modify:
```
dandavison: Nice

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+```python
+client = await Client.connect(
+    "localhost:7233",
+    interceptors=[interceptor.ContextPropagationInterceptor()],
+)
+```
+
+The `interceptors` list can contain multiple interceptors.
+In this case they form a chain: a method implemented on an interceptor instance in the list can perform side effects, and modify the data, before passing it on to the corresponding method on the next interceptor in the list.
+Your interceptor classes need not implement every method; the default implementation is always to pass the data on to the next method in the interceptor chain.
```
dandavison, line 128:
```suggestion
Your interceptor classes need not implement every method; the default implementation is always to pass the data on to the next method in the interceptor chain, eventually being handled by the "real" SDK method. 
```

Maybe? To reinforce what's going on.

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+## Worker call Interceptors
+
+At the top level, there's a `worker.Interceptor` class.
+This level is an interface with two methods named `intercept_activity` and `workflow_interceptor_class` which you can use to configure interceptions of Activity and Workflow calls, respectively.
```
dandavison: This two-level structure applies to `client.Interceptor` also.

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+`intercept_activity` returns an `ActivityInboundInterceptor`.
+
+During execution, when the SDK encounters an Inbound Activity call, it will look to the first Interceptor instance, get hold of the appropriate intercepted method, and call it.
+Ihe intercepted method will perform its function then call the same method on the next Interceptor in the chain.
```
dandavison, line 135:
```suggestion
The intercepted method will perform its function then call the same method on the next Interceptor in the chain.
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+At the end of the chain the SDK will call the "real" SDK method.
+
+This example demonstrates using an interceptor to measure [Schedule-To-Start](/encyclopedia/detecting-activity-failures#schedule-to-start-timeout) latency:
```
dandavison: This conceptual / mechananical understanding applies to client and worker interceptors, so I think it's coming too far down in the document here. It would confuse me to see it here.

```suggestion
During execution, when the SDK encounters an Inbound Activity call, it will look to the first Interceptor instance, get hold of the appropriate intercepted method, and call it.
The intercepted method will perform its function then call the same method on the next Interceptor in the chain.
At the end of the chain the SDK will call the "real" SDK method.

This example demonstrates using an interceptor to measure [Schedule-To-Start](/encyclopedia/detecting-activity-failures#schedule-to-start-timeout) latency:
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,192 @@
...
+    client,
+    # ...
+)
+```
+
+:::note
+
+You should not pass your Interceptors directly to the `Worker()` constructor.
+Although an interceptor class may _inherit_ from both `client.Interceptor` and `worker.Interceptor`, you should only ever pass it to `Client.connect()`.
+This will allow a Worker to use them from the underlying Client.
```
dandavison, line 188: EDIT: I think we can ignore my comment here

I don't think this is quite right. The point we're trying to get across is that the user has the following options:

- Inherit from `client.Interceptor` and pass to `client.Connect()` in your starter and/or worker code
- Inherit from `worker.Interceptor` and pass to `Worker()`
- Inherit from both and pass to `client.Connect()` in your starter and/or worker code. The client itself will ignore any of the worker methods, but, the worker will automatically pick up the same interceptor set from its underlying client and use the worker methods. (And if you choose this one then don't pass it directly to your worker! Because then it'll get the same interceptor twice.)

The SDK README takes a stab at explaining this also

> It often happens that your worker and client interceptors will share code because they implement closely related logic. For convenience, you can create an interceptor class that inherits from both client.Interceptor and worker.Interceptor (their method sets do not overlap). You can then pass this in the interceptors argument of Client.connect() when starting your worker as well as in your client/starter code. If you do this, your worker will automatically pick up the interceptors from its underlying client (and you should not pass them directly to the Worker() constructor).

dandavison:
> @dandavison what do you think of the most recent commit changes?

Definite improvement! I did another pass; hopefully helpful rather than merely iterating. Remarkably difficult subject to explain 😅

docs/develop/python/interceptors.mdx
dandavison: OK I see you have it below; so combine this comment with my comment below https://github.com/temporalio/documentation/pull/3686#discussion_r2211652304

axfelix: Cool — moved this up a bit higher.

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+    interceptors=[SimpleWorkerInterceptor()],
+    # ...
+)
+```
+
+:::note
+
+If you are inheriting methods from both `client.Interceptor` and `worker.Interceptor`, you should not pass your Interceptors directly to the `Worker()` constructor — instead, just pass it to `Client.connect()`.
+This will allow a Worker to use them from the underlying Client.
+In other words, only inherit from `worker.Interceptor` and pass the methods to the `Worker()` if you are not using Client methods.
```
dandavison:
```suggestion
In other words, only pass the Interceptors to the `Worker()` if you are not using Client methods.
```

dandavison: I think we can ignore my longer comment above. Thisn section is good to go with this change IMO. (the "only inherit from worker.Interceptor" added confusion since they _do_ have the option of doing that when passing to `Client.connect()`)

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+Your interceptor classes need not implement every method; the default implementation is always to pass the data on to the next method in the interceptor chain.
+During execution, when the SDK encounters an Inbound Activity call, it will look to the first Interceptor instance, get hold of the appropriate intercepted method, and call it.
+The intercepted method will perform its function then call the same method on the next Interceptor in the chain.
+At the end of the chain the SDK will call the "real" SDK method.
+
+## Worker call Interceptors
+
+At the top level, there's a `worker.Interceptor` class.
+This level is an interface with two methods named `intercept_activity` and `workflow_interceptor_class` which you can use to configure interceptions of Activity and Workflow calls, respectively.
```
dandavison: Let's make this consistent with the client calls section. The two level structure applies to both so let's not mention it unless we're mentioning it in both. Here's a version that makes them the same and doesn't mention two-level:

```suggestion
To modify inbound and outbound Workflow and Activity calls, define a class inheriting from `worker.Interceptor`.
This is an interface with two methods named `intercept_activity` and `workflow_interceptor_class` which you can use to configure interceptions of Activity and Workflow calls, respectively.
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+- `start_child_workflow()`
+- `signal_child_workflow()`
+- `signal_external_workflow()`
+- `start_nexus_operation()`
+- `start_local_activity()`
+
+#### [Inbound Activity calls](https://python.temporal.io/temporalio.worker.ActivityInboundInterceptor.html)
+
+- `execute_activity()`
+  - I.e. handle a task to execute an Activity (this is the only Inbound Activity call)
```
axfelix:
```suggestion
  - i.e. handle a task to execute an Activity (this is the only Inbound Activity call)
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+worker = Worker(
+    client,
+    interceptors=[SimpleWorkerInterceptor()],
+    # ...
+)
+```
+
+:::note
+
+If you are inheriting methods from both `client.Interceptor` and `worker.Interceptor`, you should not pass your Interceptors directly to the `Worker()` constructor — instead, just pass it to `Client.connect()`.
```
brianmacdonald-temporal:
```suggestion
If you are inheriting methods from both `client.Interceptor` and `worker.Interceptor`, you should not pass your Interceptors directly to the `Worker()` constructor — instead, pass it to `Client.connect()`.
```

docs/develop/python/interceptors.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+Your interceptor classes need not implement every method; the default implementation is always to pass the data on to the next method in the interceptor chain.
+During execution, when the SDK encounters an Inbound Activity call, it will look to the first Interceptor instance, get hold of the appropriate intercepted method, and call it.
+The intercepted method will perform its function then call the same method on the next Interceptor in the chain.
+At the end of the chain the SDK will call the "real" SDK method.
+
+## Worker call Interceptors
+
+To modify inbound and outbound Workflow and Activity calls, define a class inheriting from `worker.Interceptor`.
+This is an interface with two methods named `intercept_activity` and `workflow_interceptor_class` which you can use to configure interceptions of Activity and Workflow calls, respectively.
```
brianmacdonald-temporal:
```suggestion
This is an interface with two methods named `intercept_activity` and `workflow_interceptor_class`, which you can use to configure interceptions of Activity and Workflow calls, respectively.
```

brianmacdonald-temporal (approved):
A couple of tiny changes.

PR #3693, opened 2025-07-17
MCR updates

bechols:
Adds mentions of multi-cloud replication (MCR) in relevant areas
Describes MCR networking
Removes potentially misleading regions callout


brianmacdonald-temporal (approved):
Changes are good; thanks!

PR #3703, opened 2025-07-25
update audit log doc

bechols:
## What does this PR do?

- Adds some audit log events
- Adds references to GCP setup.
- Updates formatting a bit

### DIFF
--- a/docs/production-deployment/cloud/audit-logging.mdx
+++ b/docs/production-deployment/cloud/audit-logging.mdx
@@ -18,62 +18,74 @@ tags:
   - Logging
 ---
 
-Audit Logging is a feature of [Temporal Cloud](/cloud/overview) that provides forensic access information at the account level, the user level, [Namespace](/namespaces) level, and the Nexus Endpoint level.
+Audit Logging is a feature of [Temporal Cloud](/cloud/overview) that provides forensic access information for a variety of operations in the Temporal Cloud control plane.
 
 Audit Logging answers "who, when, and what" questions about Temporal Cloud resources.
 These answers can help you evaluate the security of your organization, and they can provide information that you need to satisfy audit and compliance requirements.
 
+:::info
+
+Audit Logging does NOT capture data plane events, like Workflow Start, Workflow Terminate, Schedule Create, etc.
+Instead, explore the [Export](/cloud/export) feature, which does let you send closed Workflow Histories to external storage.
+
+:::
+
 ## Which integrations are supported by Audit Logging? {#supported-integrations}
 
-Audit Logging supports the [Amazon Kinesis](https://docs.aws.amazon.com/kinesis/) streaming-data platform.
+Audit Logging supports both [Amazon Kinesis](/cloud/audit-logging-aws) and [Google Cloud Pub/Sub](/cloud/audit-logging-gcp) streaming-data platforms.
 
 ## Which events are supported by Audit Logging? {#supported-events}
 
-Audit Logging supports [Admin Operation events](#admin-operation-events) as well as [API Key Create, Update, and Delete events](#admin-operation-events).
-
-Please note that Audit Logging only captures operations in the Temporal Cloud control plane and not the data plane.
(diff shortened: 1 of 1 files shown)


docs/production-deployment/cloud/audit-logging.mdx
```diff
@@ -18,62 +18,74 @@ tags:
...
-- Create API Key: `CreateAPIKey`
-- Delete API Key: `DeleteAPIKey`
-- Update API Key: `UpdateAPIKey`
+  - `CreateNexusEndpoint`: Create Nexus Endpoint
+  - `DeleteNexusEndpoint`: Delete Nexus Endpoint
+  - `UpdateNexusEndpoint`: Update Nexus Endpoint
+- API Keys
+  - `CloudCreateAPIKey`: Cloud Create API Key
+  - `CloudDeleteAPIKey`: Cloud Delete API Key
+  - `CloudUpdateAPIKey`: Cloud Update API Key
```
bechols: @alice-yin what are Cloud API keys vs regular API keys?

docs/production-deployment/cloud/audit-logging.mdx
```diff
@@ -18,62 +18,74 @@ tags:
...
+  - `DeleteServiceAccount`: Delete Service Account
+  - `UpdateServiceAccount`: Update Service Account
+- Export Sinks
+  - `CreateExportSink`: Create Export Sink
+  - `CreateNamespaceExportSink`: Create Namespace Export Sink
+  - `DeleteExportSink`: Delete Export Sink
+  - `DeleteNamespaceExportSink`: Delete Namespace Export Sink
+  - `UpdateExportSink`: Update Export Sink
+  - `UpdateNamespaceExportSink`: Update Namespace Export Sink
+  - `ValidateNamespaceExportSink`: Validate Namespace Export Sink
```
bechols: @alice-yin what's the difference between NamespaceExport and Export?

alice-yin: NamespaceExport -- This is the new cloud ops API
Export -- This is the legacy API

UI/tcld is already using the new cloud ops API. However, earlier this year I checked netflix is still using Export legacy API that blocks us from completely getting rid of this API.

alice-yin: On the doc page, we may just expose only new APIS: e.g CreateNamespaceExportSink to avoid confusion? That's majority of the customer will see. 

There's some history on this. But I think we should set up a strategy to stop supporting for legacy APIs. It's very painful to maintain two.

docs/production-deployment/cloud/audit-logging.mdx
alice-yin: I think they are one for legacy api and one for new Cloud-ops api. However, I don't know what's the usage story on these two apis. I think showing 2 types of operation on public doc is a bit confusing. 

@mattkim owns API key?

docs/production-deployment/cloud/audit-logging.mdx
bechols: Agree with only showing the new APIs - I'll remove the legacy ones.

brianmacdonald-temporal (approved):
Changes look fine to me.

docs/production-deployment/cloud/audit-logging.mdx
captainbeardo: I don't know the history, but it does look like it's for two separate APIs. The body of the message is slightly different. In the types the only difference is the capitalization: https://github.com/temporalio/saas-control-plane/blob/main/cmn/auditlog/types.go#L56-L61

alice-yin: Are we comfortable just showing either regular or cloud api keys here? Showing 2 could be a bit confusing. 

(Assuming customer just want to have an idea of about what kind of operations show in auditlog, they won't do a 1 to 1 mapping on their side.)

bechols: I'm going to remove the cloud version for now, IAM can advise if we need to change.

PR #3707, opened 2025-07-27
fix gtag console error

bechols:
## What does this PR do?

Today in prod:
Google Chrome-Temporal Cloud guide  Temporal Platform Documentation-2025-07-27 at 14 13 47@2x

This PR adds a gtag value to the docusaurus config that matches the config in static/scripts/googletag.js, which prevents the error.

If we're not using Google Tag Manager any more, we should remove both the docusaurus config and the googletag.js file to remove the error.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -254,11 +254,11 @@ module.exports = async function createConfigAsync() {
           theme: {
             customCss: require.resolve("./src/css/custom.css"),
           },
-          // gtag: {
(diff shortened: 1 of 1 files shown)


axfelix:
I do not believe we are still using Google Tag Manager in any way, but just in case someone at the org *is* using it and will miss it, going to merge these out of an abundance of caution.

axfelix (approved):

PR #3709, opened 2025-07-28
Upgrade to React 19

bechols:
## What does this PR do?

React 18 -> React 19. Immediate value: much clearer hydration error messages (see comparison below). This is also future-proofing because Docusaurus v4 will drop React 18 support.

React 19
image

React 18
image

Tested by building locally and sampling pages - no build warnings, no visual changes.


axfelix:
No objections here and Vercel looks happy. Thanks for the housekeeping.

PR #3712, opened 2025-07-28
update directories + file extensions for vale

bechols:
## What does this PR do?

All the `lint` commands in package.json are no-ops. Seems like they were targeting an outdated folder structure (the only mention of docs-src in the repo is a comment saying something was migrated from it).

They were also only targeting .md files, while most new docs files are .mdx.

This PR updates the `lint` command to run on all md and mdx files in /docs. The language-specific commands run on md and mdx files in /docs/develop/{language}.

I'm not sure what the desired behavior actually is. There's potentially....a lot to fix.
Cursor-index mdx (Working Tree) (index mdx) — documentation-2025-07-28 at 15 20 33@2x

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -1049,15 +1049,27 @@ h1 {
   margin-right: 40px;
 }
 
+.navbar__items--right {
+  flex-direction: row-reverse;
+}
+
+@media (max-width: 1241px) {
+  .navbar__link {
+    display: none;
+  }
+
+  a[href^="/ai-cookbook"] {
+    display: inline;
+  }
+}
+
 @media only screen and (min-width: 1400px) {
   .navbar__items--right {
-    width: 35vw;
-    flex-direction: row-reverse;
     justify-content: space-between;
   }
 
   .DocSearch {
-    width: 30vw;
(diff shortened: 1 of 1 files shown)


axfelix:
Yeah, I have a feeling I know what's going on here -- we used to generate .mdx from .md files that were directly edited, we moved away from that approach a while ago, and we probably never updated the Vale rules, so we haven't been following them super closely as a consequence. We were never heavily, heavily all-in on Vale, so this is a legacy process, but it's one that we should bring back. Flagging @brianmacdonald-temporal here -- if we merge this, there's going to be a lot of boring Vale fix PRs that we might not even fully agree with.

.vale.ini
```diff
@@ -8,10 +8,6 @@ Vocab = Temporal
 [*]
```
lennessyy: Can we change this to 

```
[*.{md,mdx}]
```

It doesn't make sense for it to lint the code files i think. If it's to lint the comments, I don't think we need to apply the same standard as our public docs. This generates a lot of noise for the TS files.

flippedcoder: Made that update! Thanks for pointing it out.

PR #3716, opened 2025-07-29
Versioning inheritance semantics

carlydf:
## What does this PR do?

Document versioning inheritance semantics

## Notes to reviewers

I'm not sure where to put this section, and am very open to feedback saying this is confusing. 
We have reasons for all of these decisions, I just didn't include the reasons because it's already a bit wordy and I didn't want to summarize an entire design doc worth of scenarios.

### DIFF
--- a/docs/production-deployment/worker-deployments/worker-versioning.mdx
+++ b/docs/production-deployment/worker-deployments/worker-versioning.mdx
@@ -497,6 +497,7 @@ When workflows start new runs, versioning behavior and version information may b
 #### Ways Workflows Start New Runs
 
 A workflow can start a new run through:
+
 - Starting a [Child Workflow](https://docs.temporal.io/child-workflows)
 - Invoking [Continue-As-New](https://docs.temporal.io/workflow-execution/continue-as-new)
 - Retrying per its [Retry Policy](https://docs.temporal.io/encyclopedia/retry-policies)
@@ -512,10 +513,12 @@ A workflow can start a new run through:
 ##### Child Workflows
 
 **When Parent is AutoUpgrade:**
+
 - Child inherits no initial Versioning Behavior
 - Child starts on the Current Version of its Worker Deployment like all new workflow executions
 
 **When Parent is Pinned:**
+
 - Child inherits the parent's version if the child's Task Queue belongs to that version
 - Child's first workflow task executes in the same version as its parent
 - If child is also Pinned: child remains Pinned to the inherited version for its lifetime
@@ -525,31 +528,38 @@ A workflow can start a new run through:
 ##### Continue-As-New
 
 **When Original Workflow is Pinned:**
+
 - The Pinned version is inherited across the Continue-As-New chain
 - If the new run's Task Queue is not in the same Worker Deployment as the original workflow: no inheritance occurs, new run starts on Current Version of its task queue
 
 **When Original Workflow is AutoUpgrade:**
+
 - No version inheritance occurs
 
 ##### Retries
 
 **Inheritance Conditions (all must be met):**
+
 - The retried run is effectively pinned at the time of retry
 - The retried run inherited a pinned version when it started (i.e., it is a child of a pinned parent, or a Continue-As-New of a pinned run)
 - The retried run is running on a Task Queue in the inherited version
 
 **When Conditions Not Met:**
+
 - No version inheritance occurs
 
 ##### Cron Jobs
+
 - **Never inherit** versioning behavior or version
 
 ##### Resets
+
 - Inheritance behavior follows the same rules as the execution type that would result from the reset
 
 #### Versioning Override vs. Non-Override Inheritance
 
 ##### Versioning Override Inheritance
+
 - Children, crons, retries, and continue-as-new inherit the source run's override **if**:
   - The override is pinned, **AND**
   - The new workflow's Task Queue belongs to the override version
--- a/docs/production-deployment/worker-deployments/worker-versioning.mdx
+++ b/docs/production-deployment/worker-deployments/worker-versioning.mdx
@@ -492,49 +492,65 @@ We are continuing to improve this feature, and we welcome any feedback or featur
 
 ### Inheritance Semantics
 
-There are various ways in which one run of a workflow can start other workflow runs.
-The following section names each of those scenarios and describes how versioning behavior and/or version are inherited.
+When workflows start new runs, versioning behavior and version information may be inherited depending on the scenario. This section explains how inheritance works across different workflow execution patterns.
 
-A workflow can start a new run by:
+#### Ways Workflows Start New Runs
+
+A workflow can start a new run through:
 - Starting a [Child Workflow](https://docs.temporal.io/child-workflows)
 - Invoking [Continue-As-New](https://docs.temporal.io/workflow-execution/continue-as-new)
 - Retrying per its [Retry Policy](https://docs.temporal.io/encyclopedia/retry-policies)
 - Being [Reset](https://docs.temporal.io/workflow-execution/event#reset)
-- Starting another iteration of a [Cron Job](https://docs.temporal.io/cron-job) (superseded by [Schedules](https://docs.temporal.io/schedule)
+- Starting another iteration of a [Cron Job](https://docs.temporal.io/cron-job) (superseded by [Schedules](https://docs.temporal.io/schedule))
 
-When an AutoUpgrade workflow starts a new run, the version is never inherited.
+#### Inheritance Rules Overview
 
-When a Pinned Parent workflow starts a child, the child will inherit the parent's version. So the first workflow task of
-the child will execute in the same version as its parent. If the child is also Pinned, the user can enjoy the guarantee
-that the parent-child tree will run in the same version. If the child workflow is annotated as Auto-Upgrade, the workflow's
-behavior will change to Auto-Upgrade after the first task is completed.
+**Key Principle**: AutoUpgrade workflows never inherit versions, while Pinned workflows may inherit under specific conditions.
 
-If the child's task queue is not in the same Worker Deployment as the parent, then the child will not inherit, and it will
-start on the Current Version of its task queue.
+#### Inheritance by Scenario
 
-When an Auto-Upgrade Parent workflow starts a child, the child will inherit no initial versioning behavior, and will start
-on the current version of its Worker Deployment like all other new workflow executions.
+##### Child Workflows
 
----
+**When Parent is AutoUpgrade:**
+- Child inherits no initial Versioning Behavior
+- Child starts on the Current Version of its Worker Deployment like all new workflow executions
 
-When a Pinned workflow Continues-As-New, the Pinned version is inherited across the Continue-As-New chain.
+**When Parent is Pinned:**
+- Child inherits the parent's version if the child's Task Queue belongs to that version
+- Child's first workflow task executes in the same version as its parent
+- If child is also Pinned: child remains Pinned to the inherited version for its lifetime
+- If child is Auto-Upgrade: child's behavior changes to Auto-Upgrade after the first task completes
+- If child's Task Queue is not in the same Worker Deployment as parent: no inheritance occurs, child starts on Current Version of its task queue
 
----
-Versioning Override:
-    // Children, crons, retries, and continue-as-new will inherit source run's override if the override is pinned
-    // and if the new workflow's Task Queue belongs to the override version.
+##### Continue-As-New
 
-Non-Override Pinned Behavior and Version:
-    // Child of pinned parent will inherit the parent's version if the Child's Task Queue belongs to that version.
-    //
-    // New run initiated by workflow ContinueAsNew of pinned run, will inherit the previous run's version if the
-    // new run's Task Queue belongs to that version.
-    //
-    // New run initiated by workflow Cron will never inherit.
-    //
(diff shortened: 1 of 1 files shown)


axfelix:
Am leaning to thinking that this is the point at which we add an Encyclopedia page for this stuff, so that way we can include explanations as well. Want me to re-tool this PR a bit along those lines?

carlydf:
> Am leaning to thinking that this is the point at which we add an Encyclopedia page for this stuff, so that way we can include explanations as well. Want me to re-tool this PR a bit along those lines?

yes that would be amazing! lmk what you need from me. definitely feel free to check out this branch and do whatever you want to it!

For what it's worth, I think it would be nice to have Auto-Upgrade and Pinned definitions in the encyclopedia so that those can be linked to in places where those terms are used. Right now they're only defined in the "getting started with worker versioning" section.

Also, the "version statuses" (Inactive -> Active -> Draining -> Drained) are good candidates for encyclopedia IMO

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+tags:
+  - versioning
+  - safe deploys
+---
+
+This page defines some of the underlying concepts used in [Worker Versioning](/production-deployment/worker-deployments/worker-versioning):
+
+- [Versioning Behaviors](#versioning-behaviors)
+- [Versioning Statuses](#versioning-statuses)
+- [Inheritance Semantics](#inheritance-semantics)
```
drewhoskins-temporal: Not sure what this means at this stage, so it should be contextualized.  Inheritance of what?  At this point I'm thining of OOP.

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+A An **Auto-Upgrade** Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows are not restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
+
+### Versioning Statueses {#versioning-statuses}
+
+A Worker Deployment Version moves through the following states:
+
+1. **Inactive**: The version exists because a Worker with that version has polled the server. If this version never becomes Active, it will never be Draining or Drained.
+2. **Active**: The version is either Current or Ramping, so it is accepting new Workflows and existing auto-upgrade Workflows.
+3. **Draining**: The version stopped being Current or Ramping, and it has open pinned Workflows running on it. It is possible to be Draining and have no open pinned Workflows for a short time, since the drainage status is updated periodically.
+4. **Drained**: The version was draining and now all the pinned Workflows that were running on it are closed.
```
drewhoskins-temporal: mention queries?

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+#### Auto-Upgrade Workflows {#auto-upgrade}
+
+A An **Auto-Upgrade** Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows are not restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
+
+### Versioning Statueses {#versioning-statuses}
+
+A Worker Deployment Version moves through the following states:
+
+1. **Inactive**: The version exists because a Worker with that version has polled the server. If this version never becomes Active, it will never be Draining or Drained.
+2. **Active**: The version is either Current or Ramping, so it is accepting new Workflows and existing auto-upgrade Workflows.
```
drewhoskins-temporal, line 43: Should we mention the CLI command that triggers this?  Or link out to the appropriate section?

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+3. **Draining**: The version stopped being Current or Ramping, and it has open pinned Workflows running on it. It is possible to be Draining and have no open pinned Workflows for a short time, since the drainage status is updated periodically.
```
drewhoskins-temporal: Let's give a hint as to why this would happen to help people get a feel for what might have caused it.  If I'm like "why is my deployment draining?" this could answer that question.  Something like "the version stopped being Current or Ramping, either because A or B."

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+4. **Drained**: The version was draining and now all the pinned Workflows that were running on it are closed.
+
+### Inheritance Semantics {#inheritance-semantics}
+
+When Workflows start new runs, versioning behavior and version information may be inherited depending on the scenario. This section explains how inheritance works across different Workflow execution patterns.
```
drewhoskins-temporal: nit: was briefly confused.  inherited by what?
Active voice might help here
"When Workflows start new runs, the new run may inherit their versioning behavior."

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+A Workflow can start a new run through:
+
+- Starting a [Child Workflow](https://docs.temporal.io/child-workflows)
+- Invoking [Continue-As-New](https://docs.temporal.io/workflow-execution/continue-as-new)
+- Retrying per its [Retry Policy](https://docs.temporal.io/encyclopedia/retry-policies)
+- Starting another iteration of a [Cron Job](https://docs.temporal.io/cron-job) (superseded by [Schedules](https://docs.temporal.io/schedule))
+
+#### Inheritance Rules Overview
+
+**Key Principle**: Auto-upgrade Workflows never inherit versions, while Pinned workflows may inherit under specific conditions.
```
drewhoskins-temporal: "Pinned workflows may inherit under specific conditions"

I'd like to give a more accessible summary here for the folks who don't want to read the whole thing.


And my hope is we can generalizee

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+#### Inheritance by Scenario
+
+##### Child Workflows
+
+**When Parent is Pinned:**
```
drewhoskins-temporal, line 71: I think this section can be way shorter.
Since, for example, for a child workflow, the default task queue is the parent's, we can start people off with the default.  

"By default, PINNED workflows will pass their version to any PINNED children.  If you set the task queue on such a child workflow to one in a Different deployment, it will be pinned to that Deployment's current version."

I also wonder if this can be generalized (i.e. task queue semantics are the same for continue as new)

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+- Child inherits the parent's version if the child's Task Queue belongs to that version
+- Child's first Workflow task executes in the same version as its parent
+- If child is also Pinned: child remains Pinned to the inherited version for its lifetime
+- If child is Auto-Upgrade: child's behavior changes to Auto-Upgrade after the first task completes
+- If child's Task Queue is not in the same Worker Deployment as parent: no inheritance occurs, child starts on Current Version of its task queue
```
drewhoskins-temporal, line 77: Do we have a succinct way of describing the "version that the workflow will go to--current if current and with some probability N if it's ramping at N"?

carlydf: Ramping and Current versions are both the "Active" versions of the Deployment, so we could start using that. I agree it would be good to have a succinct way of saying that fact

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+#### Pinned Workflows {#pinned}
+
+A **Pinned** Workflow is guaranteed to complete on a single Worker Deployment Version. You can mark a Workflow Type as pinned when you register it by adding an additional Pinned parameter. If you need to move a pinned Workflow to a new version, use [`temporal workflow update-options`](https://docs.temporal.io/cli/workflow#update-options).
+
+#### Auto-Upgrade Workflows {#auto-upgrade}
+
+A An **Auto-Upgrade** Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows are not restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
+
+### Versioning Statueses {#versioning-statuses}
```
carlydf: "Version Statuses"

docs/encyclopedia/workers/worker-versioning.mdx
carlydf: Continue-as-new, Retry, and Child Workflow Semantics

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: Changed to "Continue-as-new, Child Workflow, and Retry Semantics" as below.

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: Added:

Closed Workflows may still re-run some code paths if they are [Queried](https://docs.temporal.io/sending-messages#sending-queries) within their [Retention Period](https://docs.temporal.io/temporal-service/temporal-server#retention-period) and Workers with that version are still polling.

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: Leaning toward no here — I don't think it follows directly in the Encyclopedia.

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: How's:

3. **Draining**: The version has open pinned Workflows running on it, but stopped being Current or Ramping, usually because a newer version has been deployed. It is possible to be Draining and have no open pinned Workflows for a short time, since the drainage status is updated only periodically.

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: @carlydf? I'm admittedly having a hard time figuring out how to summarize this into a consistent rule that doesn't rely on reading through the various conditions.

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+Using **Workflow Pinning**, you can declare each Workflow type to have a **Versioning Behavior**, either Pinned or Auto-Upgrade.
+
+#### Pinned Workflows {#pinned}
+
+A **Pinned** Workflow is guaranteed to complete on a single Worker Deployment Version. You can mark a Workflow Type as pinned when you register it by adding an additional Pinned parameter. If you need to move a pinned Workflow to a new version, use [`temporal workflow update-options`](https://docs.temporal.io/cli/workflow#update-options).
+
+#### Auto-Upgrade Workflows {#auto-upgrade}
+
+A An **Auto-Upgrade** Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows are not restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
```
angelazhou32:
```suggestion
An **Auto-Upgrade** Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows are not restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
```

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: From meeting, to add to summary: "Pinned Workflows will inherit the version of their parent by default. If in a separate deployment, it will be pinned according to the version of that deployment."

Mostly what we want is to get rid of the "under specific conditions" clause -- and possibly replace with the suggestion in the below comment.

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: We won't block this PR for now -- but should resolve use of "Active" term later on.

docs/encyclopedia/workers/worker-versioning.mdx
axfelix: Added the first sentence of your comment to the top of the section and explained that what follows covers specific scenarios.

axfelix:
OK! based on our conversation this morning I think this is good to merge.

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+**When Conditions Not Met:**
+
+- No version inheritance occurs
+
+##### Cron Jobs
+
+- **Never inherit** versioning behavior or version
+
+#### Versioning Override vs. Non-Override Inheritance
```
brianmacdonald-temporal: Since this section only contains one bullet list, it seems like one of these headings is redundant. That is, there's a section on "versioning override inheritance," but no specific content related to "non-override inheritance" (I assume because non-override is the default behavior, so there's nothing to talk about). If that's the case, I think you can delete the "versus" heading and bump up "Versioning Override Inheritance" by one level.

PR #3722, opened 2025-07-31
clarify Python and Java Worker Versioning imports

axfelix:
Python and Java need a lot of non-obvious imports to configure Versioned Workers, here they are


jsundai (approved):
Agreed and LGTM!

PR #3728, opened 2025-08-01
HA cleanup

bechols:
## What does this PR do?

Make each page clearer and reduce duplicate language. Delete the HA regions page in favor of the overall regions page (they had the same content).


brianmacdonald-temporal (approved):
Changes look good. Well reorganized!

PR #3729, opened 2025-08-01
Add cancellation docs for Java SDK

Quinn-With-Two-Ns:
## What does this PR do?

Add cancellation docs for Java SDK. Tried to cover similar content to https://docs.temporal.io/develop/python/cancellation


angelazhou32 (approved):
LGTM

PR #3743, opened 2025-08-06
Added await to async calls in dotnet test suite examples

flippedcoder:


brianmacdonald-temporal (approved):
Good community change. Approved.

PR #3755, opened 2025-08-11
Updated doc to use ts instead of js on snippet

flippedcoder:
## Notes to reviewers

[This page](https://docs.temporal.io/develop/typescript/timers) had some weird formatting on this code snippet.
Screenshot 2025-08-11 at 09 18 05

Changing the block to `ts` makes it look like this:
Screenshot 2025-08-11 at 09 18 50


brianmacdonald-temporal (approved):
Good catch; thanks!

PR #3760, opened 2025-08-12
CLI: Fix auto-gen comment placement

yuandrew:
## What does this PR do?
Moved the top-of-file comment to below the header, this was causing formatting issues. CLI fix is https://github.com/temporalio/cli/pull/842, but we want a quicker fix before waiting for the next CLI release


yuandrew:
yeah, just noticed it's still not right, I'll keep tweaking it until I get it back to where it was, since I broke it originally haha

axfelix (approved):

PR #3763, opened 2025-08-13
Python - Enriching the User Interface docs

THardy98:
## What does this PR do?

Adds documentation for "Enriching the User Interface", detailing usage of the user metadata feature in Python SDK

### DIFF
--- a/docs/evaluate/temporal-cloud/actions.mdx
+++ b/docs/evaluate/temporal-cloud/actions.mdx
@@ -30,8 +30,9 @@ The following result in an action on Temporal Cloud:
 ## Workflows
 
 - **Workflow started**.
-  Occurs via client start, [Continue-As-New](/workflow-execution/continue-as-new), [Child Workflow](/child-workflows) start. Also occurs with [Update-With-Start](/sending-messages#update-with-start) if the Workflow was not already started.
+  Occurs via client start, [Continue-As-New](/workflow-execution/continue-as-new), [Child Workflow](/child-workflows) start.
   If a Workflow start fails, an Action is not recorded.
+  De-duplicated Workflow starts that share a Workflow ID do _not_ count as an Action.
 - **Workflow reset**.
   Occurs when a [Workflow](/workflows) is reset.
   (Actions that occur before a [Reset](/workflow-execution/event#reset) are counted even if they are no longer visible in [Event History](/workflow-execution/event#event-history).)
@@ -51,6 +52,8 @@ The following result in an action on Temporal Cloud:
   De-duplicated Updates that share an Update ID do _not_ count as an Action.
 - **Side Effect recorded**.
   For a mutable [Side Effect](/workflow-execution/event#side-effect), an Action occurs only when the value changes.
+- **Workflow Execution Options updated.** An Action occurs for every [Workflow-Execution-Options-Updated](/references/events#workflowexecutionoptionsupdated) event.
+  This includes attaching a Workflow completion callback or modifying a Workflow versioning override.
 
 ## Child Workflows
 


docs/develop/python/user-metadata.mdx
cretz, line 1: We need to clarify that static summary, static details, summary, and current details on all of these are usually markdown and that is how the UI works with them. Summaries are single-line markdown, details are multi-line. Can get with UI team to confirm exact details of what facets of markdown are supported.

docs/develop/python/user-metadata.mdx
```diff
@@ -0,0 +1,259 @@
...
+    id="your-workflow-id",
+    task_queue="your-task-queue",
+    static_summary="Order processing for customer #12345",
+    static_details="Processing premium order with expedited shipping"
+)
+```
+
+#### Inside the Workflow
+
+Within a workflow, you can get and set the _current workflow details_. These details are distinct from the static details provided to workflow start/execute, and are included in what is returned from the `__temporal_workflow_metadata` query.
```
cretz: This is an internal query and its name is not important. It's just important to explain "Static summary and static details are fixed at workflow start, however, current details can be changed throughout the workflow. Current details are fetched by interfaces like the UI via query."

docs/develop/python/user-metadata.mdx
```diff
@@ -0,0 +1,259 @@
...
+    @workflow.run
+    async def run(self, input: str) -> str:
+        # Create a timer with a summary
+        await workflow.sleep(timedelta(minutes=5), summary="Waiting for payment confirmation")
+        return "Timer completed"
+```
+
+### Signal Handler
+
+When defining a signal handler in a workflow, you can include a description:
```
cretz: So while "descriptions" were done as part of the user metadata project, I am not sure they should be considered "user metadata" and documented as such. It was just part of the project because we were getting workflow information via query anyways. If we do want to document that you can set descriptions on these handlers, we should do so where we discuss the defining of those handlers (but that can be separate from this PR).

docs/develop/python/user-metadata.mdx
```diff
@@ -0,0 +1,259 @@
...
+    async def run(self, input: str) -> str:
+        # Workflow logic
+        return f"Order shipped to: {self._shipping_address}"
+
+    @workflow.update(description="Update the shipping address for the order")
+    def change_shipping_address(self, new_address: str) -> None:
+        self._shipping_address = new_address
+```
+
+## Retrieving Metadata
```
cretz: I don't think retrieving metadata really needs to be documented tbh. This is mostly only done by Temporal interfaces, the user need to retrieve this information is almost non-existent. (could be argued the static summary/details stuff has value, but really it's just a part of everything else you can describe, doesn't deserve special docs)

THardy98: removed

docs/develop/python/user-metadata.mdx
THardy98: removed, we can do so in a separate PR

docs/develop/python/user-metadata.mdx
THardy98: Updated to reflect this

docs/develop/python/enriching-ui.mdx
```diff
@@ -0,0 +1,111 @@
...
+also takes Markdown format (again, excluding images, html, and scripts) and can span multiple lines. 
+
+```python
+@workflow.defn
+class YourWorkflow:
+    @workflow.run
+    async def run(self, input: str) -> str:
+        # Get the current details
+        current_details = workflow.get_current_details()
+        print(f"Current details: {current_details}")
```
cretz, line 74: Would be ok if we removed the getter from this sample code. It's there, but not that important IMO. But not a bit deal

docs/develop/python/enriching-ui.mdx
```diff
@@ -0,0 +1,111 @@
...
+        
+        # Set/update the current details
+        workflow.set_current_details("Updated workflow details with new status")
+        
+        return "Workflow completed"
+```
+
+### Adding Summary to Activities and Timers
```
cretz, line 82: The 3-# heading is a bit strange on here. Any reason why it's heavier than 4-# "Starting a Workflow"?

THardy98: no, overlooked that, changed to 4#

PR #3768, opened 2025-08-14
TCLD page nits

shivam-ajmera:
- NamespaceID consists of the accountID, to keep it consistent with other samples remove the `<account_id>`
- Escaping the string causes the italics formatting to be printed literally: we print `_example_` when in reality we want _example_


flippedcoder (approved):
Thanks for the clean up!

PR #3794, opened 2025-08-27
added note that you can't configure nexus callback limits

GSmithApps:
## What does this PR do?

Adds note that you can't configure nexus callback limits

## Notes to reviewers

currently draft


brianmacdonald-temporal (approved):
Change looks good

PR #3803, opened 2025-08-30
update task dispatching

bechols:
## What does this PR do?

Update description of the order of task dispatching


docs/encyclopedia/workers/task-queues.mdx
```diff
@@ -136,15 +136,14 @@ There are five places where the name of the Task Queue can be set by the develop
...
-Task Queues with multiple partitions do not have any ordering guarantees.
-Once there is a backlog of Tasks that have been written to disk, Tasks that can be dispatched immediately (“sync matches”) are delivered before tasks from the backlog (“async matches”).
-This approach optimizes throughput.
+By [default](/references/dynamic-configuration#service-level-rps-limits) each Task Queue has 4 partitions.
 
 Task Queues with a single partition are almost always first-in, first-out, with rare edge case exceptions.
 However, using a single partition limits you to low- and medium-throughput use cases.
 
+In Task Queues with multiple partitions, each task is assigned to a random partition.
+Generally partitions will act as FIFO queues, so even sync matches (Tasks that can be dispatched immediately) are not delivered before async matches (older tasks).
```
yiminc: Generally partitions will act as FIFO queues, so once a task queue build up backlog, the sync match rate will drop significantly to almost zero because task queue will dispatch tasks from backlog first.

bechols:
```suggestion
Generally partitions will act as FIFO queues, so once a task queue builds up a backlog, the sync match (tasks that can be dispatched immediately) rate will drop to nearly zero because the task queue will instead dispatch tasks from the backlog (i.e. async matches) first.
```

yiminc (approved):

PR #3814, opened 2025-09-04
Remove promql-dd docs

dustin-temporal:
## What does this PR do?
Removes the documentation on how to scrape metrics from promQL and send them to Datadog. This is a deprecated methodology and customers should use the supported Datadog integration that does not require them to run a service. This page now points to Datadog documentation for the integration in order to keep documentation up to date in only one place.


brianmacdonald-temporal (approved):
Looks good!

PR #3830, opened 2025-09-08
Update incorrect metric names

jtamagnan:
### Problem

I noticed that there are two metrics that are not named properly in the documentation:
 - `service_errors_with_type` is actually `service_error_with_type`. [ref](https://github.com/temporalio/temporal/blob/16b1e29b2a86eaf1a95e026e49e16f31f8c414aa/common/metrics/metric_defs.go#L595)
 - `task_latency_userlatency` doesn't exist but `service_latency_userlatency` does. [ref](https://github.com/temporalio/temporal/blob/main/common/metrics/metric_defs.go#L615)

### Solution

Update the documentation


lennessyy:
Verified in the following files
- https://github.com/search?q=org:temporalio%20service_error_with_type&type=code
- https://github.com/temporalio/temporal/blob/707b2aa2bc3e90ba675b5fcc3626616f1878ab1b/common/metrics/metric_defs.go#L616

PR #3837, opened 2025-09-11
best practices guide section (learn to docs)

jsundai:
## What does this PR do?
Backport Best Practice Guides from Learn Site to Docs

## Notes to reviewers
Putting this as a back porting PR for now. This may become a migration from learn instead of a back port so there's not duplicate content.


axfelix:
Looks good and I'm in favor of a straight port -- but we should do something about the header levels (there are full sentences in the right hand sidebar now)

PR #3845, opened 2025-09-16
docs: rewrite Typescript activity execution section

lennessyy:
## What does this PR do?

This is a mini-rewrite of the type script activity execution section. I tighten the wording, use simpler language, and explain the code more thoroughly. The original text was concept heavy and does not do enough to educate users on the syntax to execute an activity in TypeScript.


axfelix (approved):
Big improvement imo! This was needlessly complex (the TS docs deviate from the way we frame the "core application" in the other languages in not always useful ways) and now it isn't.

PR #3855, opened 2025-09-18
add additional python error handling details from course

axfelix:
Added some missing context from the error handling course to explain Activity vs Workflow Failures and Non-Retryable Errors. This should be ported to the other languages if we like it (and note that Go works a bit differently than the others).


docs/develop/python/failure-detection.mdx
```diff
@@ -24,12 +24,108 @@ description: Learn how to set Workflow and Activity timeouts, retries, retry pol
...
+        # Your activity logic goes here
+    except Exception as e:
+        raise ApplicationError(
+            type="MyCustomError",
+            message=f"Error encountered on attempt {attempt}",
+        ) from e
+```
+
+Depending on your implementation, you may decide to use either method.
+One reason to use the Temporal `ApplicationError` class is because it allows you to set an additional `non_retryable` parameter.
```
brianmacdonald-temporal, line 81:
```suggestion
An additional reason to use the Temporal `ApplicationError` class is because it allows you to set an additional `non_retryable` parameter.
```

Suggested this change since you already gave one valid reason; so this is another.

docs/develop/python/failure-detection.mdx
```diff
@@ -24,12 +24,108 @@ description: Learn how to set Workflow and Activity timeouts, retries, retry pol
...
+            message=f"Error encountered on attempt {attempt}",
+            non_retryable=True,
+        ) from e
+```
+
+You can alternately specify a list of errors that are non-retryable in your Activity [Retry Policy](#activity-retries).
+
+## Failing Workflows {#workflow-failure}
+
+One of the core design principles of Temporal is that an Activity Failure will never directly cause a Workflow Failure — a Workflow should never return as Failed unless deliberately.
```
brianmacdonald-temporal, line 104:
```suggestion
One of the core design principles of Temporal is that an Activity Failure will never directly cause a Workflow Failure — a Workflow should never return as Failed unless it is deliberately returned that way.
```

brianmacdonald-temporal (approved):
Really good additions; just did a little wordsmithing.

docs/develop/python/failure-detection.mdx
axfelix: Hmmmm, did I already give another reason?

docs/develop/python/failure-detection.mdx
axfelix: Hmmmm. Good pointing out this wasn't clear, can we just change "deliberately" to "intentional"? Fewer words.

docs/develop/python/failure-detection.mdx
```diff
@@ -24,12 +24,108 @@ description: Learn how to set Workflow and Activity timeouts, retries, retry pol
...
+        # Your activity logic goes here
+    except Exception as e:
+        raise ApplicationError(
+            type="MyNonRetryableError",
+            message=f"Error encountered on attempt {attempt}",
+            non_retryable=True,
+        ) from e
+```
+
+You can alternately specify a list of errors that are non-retryable in your Activity [Retry Policy](#activity-retries).
```
angelazhou32, line 100: Quick code sample for this here would be nice!

docs/develop/python/failure-detection.mdx
```diff
@@ -24,12 +24,108 @@ description: Learn how to set Workflow and Activity timeouts, retries, retry pol
...
+
+## Failing Workflows {#workflow-failure}
```
angelazhou32, line 102: Since we have a section on Failing Workflows, I think it would be good to mention the difference between a Workflow Task Failure vs Workflow Execution Failure

docs/develop/python/failure-detection.mdx
```diff
@@ -24,12 +24,108 @@ description: Learn how to set Workflow and Activity timeouts, retries, retry pol
...
+except ActivityError as e:
+	workflow.logger.error(f"Unable to process credit card {e.message}")
+	raise ApplicationError(
+		"Unable to process credit card", "CreditCardProcessingError"
+	)
+```
+
+This works differently in a Workflow than raising exceptions from Activities.
+In an Activity, any Python exceptions or custom exceptions are converted to a Temporal `ApplicationError`.
+In a Workflow, any exceptions that are raised other than an explicit Temporal `ApplicationError` will only fail that particular [Workflow Task](https://docs.temporal.io/tasks#workflow-task-execution) and be retried.
```
angelazhou32, line 125: Might be good to specifically mention:
- the difference between a Workflow Task Failure vs Workflow Execution Failure. I know you're doing this here, but calling it Workflow Task Failure might help people understand this error when it appears in the Web UI. 
- Also that Workflow Task Failures retry by default.

docs/develop/python/failure-detection.mdx
axfelix: This is anchored down to the rest of the same doc, I didn't want to overload it here.

docs/develop/python/failure-detection.mdx
axfelix: Personal bias, but imo I think trying to explain "Workflow Task vs Workflow Execution" gets really into the weeds (it's my least favorite aspect of the actual course right now), and I deliberately tried to avoid that here. A Workflow Task is a very marginal aspect of most people's understanding of Temporal, I want to keep the emphasis on "retry" vs "return".

PR #3857, opened 2025-09-19
Make examples more likely to work as-is when pasted.

robholland:
Remove variable references except for $namespace, where there isn't a sane default.

Fix a syntax error.


axfelix (approved):
Yup, makes sense to me!

PR #3861, opened 2025-09-19
feat: ai cookbook tiles

lennessyy:
## What does this PR do?

This PR adds the tiles component to be used for ai cookbooks. The location where the cookbook is right now is not relevant and is only for demo purposes.


lennessyy:
it's a mystery to me what got messed up in the development process. i think it's one of the typescript config files that i changed. but i cannot fix the CSS here. I nuked my repo and started another PR that does not have the CSS issues.

PR #3885, opened 2025-10-01
Port Failure handling / Non-retryable error content across languages

axfelix:
Doing these one at a time after Python, probably don't merge until we have each of the other 5.


axfelix:
Actually let's review and merge this here for now — Go and Java work differently enough that I will take marginally more time with them.

PR #3897, opened 2025-10-06
Add step for verify in audit log sink setup

david049:
## What does this PR do?
Adds a step for new verify in UI

### DIFF
--- a/docs/production-deployment/cloud/audit-logging-aws.mdx
+++ b/docs/production-deployment/cloud/audit-logging-aws.mdx
@@ -61,6 +61,8 @@ If you chose the **Manual** access method, continue with the following steps:
 1. Select **Next** on this page and on the next two pages.
 1. On the **Review** page, select **Create stack**.
 
+To ensure that audit logs can flow into the Kinesis stream, you can use the Verify button to confirm it is set up correctly.
+
 ## Consume an Audit Log {#consume-an-audit-log}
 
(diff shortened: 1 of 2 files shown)


docs/production-deployment/cloud/audit-logging-aws.mdx
```diff
@@ -74,6 +76,11 @@ Subsequent logs arrive every 2 minutes if any actions occurred during that 2-min
 You must configure and implement your own consumer of the Kinesis stream.
 For an example, see [Example of consuming an Audit Log](#example-of-consuming-an-audit-log).
 
+:::
+:::warning
+
+Certain fields in the audit log format are planned to be changed in the near future.
```
jumbogo: I'll leave this to @dustin-temporal

david049: I'll remove the warning from this diff then

david049:
> Thanks for this change; UI changes are always tricky to keep up with.
> What does the Verify process do? Does it send a test to the stream to ensure that it's set up properly?
> What sorts of errors might users expect if they don't have it set up properly?
cc: @jumbogo as the implementer
To my knowledge it does write a test (value "t") to the sink to make sure our control plane can reach it. If not set up properly the user will see an error saying `failed to write to stream:<error message>`

PR #3910, opened 2025-10-06
create JsonTable component and alter privatelink connectivity docs to…

dspinhirne-temporal:
## What does this PR do?

Creates JsonTable component that allows a source json file to be rendered as an html table. Alters the PrivateLink docs to use the component. Creates 2 json data files for PrivateLink region->service mapping. This structure will allow raw json data to be used as a single source of truth for both docs as well as potential external automation that wishes to read this data.


src/components/formatting/JsonTable.js
```diff
@@ -0,0 +1,79 @@
...
+
+    fetchData();
+  }, [filename]);
+
+  if (loading) {
+    return <div>Loading...</div>;
+  }
+
+  if (error) {
+    return <div>Error: {error}</div>;
```
flippedcoder: Since this is public-facing, we probably don't want to show an error directly on the docs pages. What do you think about this?

```suggestion
    return <div className='theme-admonition theme-admonition-warning alert alert--danger admonition_WoCw'>Please reload the page to see this table.</div>;
```

flippedcoder: That'll make it look like this:

Screenshot 2025-10-07 at 10 51 34

PR #3918, opened 2025-10-07
Claim check AI cookbook recipe

robholland:
## What does this PR do?

Adds a recipe for the claim check pattern recipe from https://github.com/temporalio/ai-cookbook/pull/9.


lennessyy (commented):
Looks good. Please merge this after the PR is merged on the cookbook repo. 

I will put up a PR that syncs the content from that repo soon so we don't need to be doing this manually.

PR #3927, opened 2025-10-08
add dns override to privatelink json data. update ha-connectivity to …

dspinhirne-temporal:
## What does this PR do?
I found another table in an ha doc that had hard-coded privatelink service data. Its table also contained dns override data. Added the additional data to the json source file and updated the ha doc to utilize the json source.


brianmacdonald-temporal (approved):
These changes look good; sorry about the delay.

PR #3928, opened 2025-10-08
Nexus .NET documentation

cretz:
## What does this PR do?

Adds Nexus .NET documentation akin to Java and Python and others before it. Almost everything is structured like those others, there is nothing very unique with the English or the structure, only code snippets were replaced basically.

### DIFF
--- a/docs/develop/dotnet/temporal-nexus.mdx
+++ b/docs/develop/dotnet/temporal-nexus.mdx
@@ -207,7 +207,7 @@ Workflow IDs should typically be business-meaningful IDs and are used to dedupe 
 
 #### Map a Nexus Operation input to multiple Workflow arguments
 
-A Nexus Operation can only take one input parameter. If you want a Nexus Operation to start a Workflow that takes multiple arguments, simple pass in different arguments.
+A Nexus Operation can only take one input parameter. If you want a Nexus Operation to start a Workflow that takes multiple arguments, simply pass in different arguments using `RunAsync`.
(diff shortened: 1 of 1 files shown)


docs/develop/dotnet/temporal-nexus.mdx
```diff
@@ -0,0 +1,502 @@
...
+
+:::tip RESOURCES
+
+[Attach multiple Nexus callers to a handler Workflow](/nexus/operations#attaching-multiple-nexus-callers) with a Conflict-Policy of Use-Existing.
+
+:::
+
+#### Map a Nexus Operation input to multiple Workflow arguments
+
+A Nexus Operation can only take one input parameter. If you want a Nexus Operation to start a Workflow that takes multiple arguments, simple pass in different arguments.
```
brianmacdonald-temporal: The end of this sentence seems vague. Extrapolating from the other SDKs, I think this is intended to be something like  "...takes multiple arguments, use the `WorkflowRunOperationHandler.FromHandleFactory` method." ?

prasek:
```suggestion
A Nexus Operation can only take one input parameter. If you want a Nexus Operation to start a Workflow that takes multiple arguments, simply pass in different arguments using `RunAsync`.
```

as shown in the example below: `(HelloHandlerWorkflow wf) => wf.RunAsync(input.Language, input.Name),`

PR #3929, opened 2025-10-08
visibility(search attributes): add documentation for aliasing support

spkane31:
## What does this PR do?

Update the [list filter](https://docs.temporal.io/list-filter) page with information about new search attribute aliasing feature.

### DIFF
--- a/docs/develop/go/set-up.mdx
+++ b/docs/develop/go/set-up.mdx
@@ -181,12 +181,12 @@ Create an Activity file (activity.go):
 package greeting
 
 import (
-    "context"
-    "fmt"
+	"context"
+	"fmt"
 )
 
 func Greet(ctx context.Context, name string) (string, error) {
-    return fmt.Sprintf("Hello %s", name), nil
+	return fmt.Sprintf("Hello %s", name), nil
 }
 ```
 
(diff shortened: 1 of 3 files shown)


docs/encyclopedia/visibility/list-filter.mdx
```diff
@@ -186,18 +186,45 @@ ExecutionTime < '2021-08-28T15:04:05+00:00' or ExecutionTime > '2021-08-22T15:04
...
+
+#### How alias resolution works
+
+When resolving a Search Attribute in a List Filter, Temporal Server checks for matches in the following order:
+
+1. Custom Search Attributes defined in the current Namespace
+
+2. Default Search Attributes
+
+This means that if you define a custom Search Attribute with the same name as a default Search Attribute, the unprefixed name will refer to your custom attribute. You can still search by the default Search Attribute by using the Temporal prefix.
```
spkane31:
```suggestion
This means that if you define a custom Search Attribute with the same name as a default Search Attribute, the non `Temporal` prefixed name will refer to your custom attribute. You can still search with the default Search Attribute by using the `Temporal` prefix.
```

docs/encyclopedia/visibility/list-filter.mdx
```diff
@@ -186,18 +186,45 @@ ExecutionTime < '2021-08-28T15:04:05+00:00' or ExecutionTime > '2021-08-22T15:04
...
+
+:::
+
+#### How alias resolution works
+
+When resolving a Search Attribute in a List Filter, Temporal Server checks for matches in the following order:
+
+1. Custom Search Attributes defined in the current Namespace
+
+2. Default Search Attributes
```
bergundy, line 216: Do we consistently call the "system" or "pre-defined" search attributes "default" in the docs? That's not what we call them in the server. Let's make sure we are consistent with the public facing terminology.

docs/encyclopedia/visibility/list-filter.mdx
```diff
@@ -186,18 +186,45 @@ ExecutionTime < '2021-08-28T15:04:05+00:00' or ExecutionTime > '2021-08-22T15:04
...
+
+This means that if you define a custom Search Attribute with the same name as a default Search Attribute, the unprefixed name will refer to your custom attribute. You can still search by the default Search Attribute by using the Temporal prefix.
+
+For example, if you have a custom Search Attribute named `WorkflowId`, List Filters using the following Search Attributes will return different results:
 
 ```sql
-order by StartTime desc, CloseTime asc
+-- If you have a custom Search Attribute named 'WorkflowId'
+-- This will use your custom attribute, not the default Search Attribute
+WorkflowId = 'custom-workflow-id-123'
```
bergundy: Use `WorkflowVersioningBehavior` for the example, we will block creating `WorkflowId` as a search attribute because it predates the `Temporal` prefix convention.

docs/encyclopedia/visibility/list-filter.mdx
lennessyy: Yes. After searching through the docs I see that we consistently call them "Default Search Attributes"

lennessyy (approved):

docs/encyclopedia/visibility/list-filter.mdx
lennessyy: Do we have a full list of the search attributes that will be blocked? I think that could be useful here to the users

PR #3932, opened 2025-10-08
docs: add worker deployments best practice

lennessyy:
## What does this PR do?

- Adds a starting page for worker deployment best practices
- [Preview](https://temporal-documentation-git-worker-best-practices.preview.thundergun.io/best-practices/worker)

### DIFF
--- a/docs/develop/python/python-sdk-sandbox.mdx
+++ b/docs/develop/python/python-sdk-sandbox.mdx
@@ -200,7 +200,7 @@ my_worker = Worker(..., workflow_runner=SandboxedWorkflowRunner(restrictions=my_
 
 The sandbox's import notification policy specifies how the sandbox behaves when it imports modules in a way that may be unintentional. It covers two common scenarios: dynamic imports and enforcing module passthrough. Each can be controlled independently.
 
-A dynamic import occurs when a module is imported after the Workflow is loaded into the sandbox. These imports are often invisible and, if they don't do anything restricted by the sandbox, cause memory overhead. By default the [`WARN_ON_DYNAMIC_IMPORT`](https://python.temporal.io/temporalio.workflow.SandboxImportNotificationPolicy.html#WARN_ON_DYNAMIC_IMPORT) policy setting is enabled and a warning will be emitted when one of these imports occurs.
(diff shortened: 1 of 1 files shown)


docs/best-practices/worker.mdx
```diff
@@ -0,0 +1,248 @@
...
+          imagePullPolicy: Always
+      enableServiceLinks: false
+```
+{/* SNIPEND */}
+
+### Separate Task Queues logically
+
+Use separate Task Queues for distinct workloads. This isolation allows you to control rate limiting, prioritize certain
+workloads, and prevent one workload from starving another. For each Task Queue, ensure you configure at least two
+Workers to poll the Task Queue.
```
jsundai, line 114: Doesn't hurt to add a bit about mismatching Task Queues. Could be some version of this. 
"A mismatch between Task Queue names on the Worker and the Client prevents the Temporal Service from dispatching Tasks to the correct Workers, effectively halting Workflow progress. The Task Queue name is specified when configuring the Worker so that it knows which Task Queue to poll on the Temporal Service, and it is also provided to the Client when starting a Workflow Execution so that the Temporal Service knows which queue to use when scheduling related Tasks. Because Task Queues are created dynamically when first used, mismatched names result in separate Task Queues with no coordination between them."

docs/best-practices/worker.mdx
```diff
@@ -0,0 +1,248 @@
...
+
```
jsundai, line 115: Another suggestion. Could add a bit of how task queues don't provide ordering guarantees. 
"Task Queues do not provide ordering guarantees across different Workflow Executions. Tasks may be executed in an order different from when they were enqueued, depending on Worker availability and load. Within a single Workflow Execution, however, ordering is fully controlled by the Workflow logic itself. Similarly, Signals sent to a specific Workflow Execution are always delivered in the order they were received."

docs/best-practices/worker.mdx
```diff
@@ -0,0 +1,248 @@
...
+	w.RegisterWorkflow(Charge)
+	w.RegisterActivity(&Activities{FraudCheckURL: config.FraudURL})
+
+	return w.Run(temporalutil.WorkerInterruptFromContext(ctx))
+}
+```
+{/* SNIPEND */}
+
+### Use Worker Versioning to safely deploy new Workflow code
```
jsundai, line 154: There could be a short section on rate limiting after the “Separate Task Queues logically” section such as "Control Worker Throughput with Rate Limiting". But this document is comprehensive enough right now!

docs/best-practices/worker.mdx
lennessyy: Adding this and a recommendation about using a constant to configure task queue names.

docs/best-practices/worker.mdx
lennessyy: This is a good suggestion, but it feels less relevant to worker deployment? Since this information is already in the Task Queue section i feel like it doesn't need to be in the worker best practice doc

PR #3942, opened 2025-10-15
Update Worker Versioning to include section on migrating from pinned to auto-upgrade workflows

flippedcoder:
## What does this PR do?

There have been questions about to to migrate a pinned workflow to an auto-upgrade workflow and we want to fill this gap before GA in mid-late Dec.


docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,23 @@ temporal workflow reset with-workflow-update-options \
     --versioning-override-build-id "$TARGET_BUILD_ID"
 ```
 
+## Migrating a Workflow from Pinned to Auto-Upgrade
+
+There may be times when you need to migrate your Workflow from Pinned to Auto-Upgrade because you support rainbow deployments now or there has been another update and you no longer want a specific Workflow using the pinned behavior.
```
drewhoskins-temporal: Thanks for clarifying the scenarios, let me try to explain when users would need this.

The biggest reason somebody might want to do this is when they configured their workflow type with the wrong annotation.  Perhaps they pinned a really long-running workflow by mistake.
 
Since pinned workflows get stuck on versions, and workflows prevent versions from draining, such long-running workflows could be preventing drainage.  They could move the workflow to a new build, but that would just push the problem to the next build.

So likely what they'll do is change their annotation in code and then "backfill" that using this command.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,23 @@ temporal workflow reset with-workflow-update-options \
...
+
+There are a few steps you need to take in order to make that change. Since Auto-Upgrade Workflows need to be replay-safe, you need to [patch](/workflow-definition#workflow-patching) your Workflow code to ensure it supports multiple versions.
```
drewhoskins-temporal: So, patching will be necessary going forward once they move to auto-upgrade, but they don't need to patch the code to make this particular change.  [Patching](https://docs.temporal.io/patching#patching) is used for certain changes to control flow in the workflow code but is not necessary in this case because these annotations don't affect the code and can't be checked within code.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,23 @@ temporal workflow reset with-workflow-update-options \
...
+
+Next, you need to update the versioning behavior for your Workflow from Pinned to Auto-Upgrade. You can use `temporal workflow update-options` for this.
+
+```bash
+temporal workflow update-options \
```
drewhoskins-temporal, line 407: Let's instead/in addition show a command where they are moving all workflows of a certain type, as given that they misconfigured it in code, they probably want to do it to all their workflows.

`--query WorkflowType=Foo`

In addition, they may just want to filter on a certain build ID to limit the number of workflows they apply it to.

`--query "WorkflowType=Foo AND --TemporalWorkerDeploymentVersion=$TARGET_DEPLOYMENT:$OLD_VERSION"`

(I didn't run this; please work with eng to verify syntax.)
See https://docs.temporal.io/search-attribute for the search attributes they can use to filter workflows.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,23 @@ temporal workflow reset with-workflow-update-options \
...
+    --workflow-id "$WORKFLOW_ID" \
+    --versioning-override-behavior auto_upgrade \
+    --versioning-override-deployment-name "$TARGET_DEPLOYMENT"
+```
+
+If there are still Tasks running in the Workflow, you should consider waiting for them to drain from the Workflow before doing this migration to avoid version issues.
```
drewhoskins-temporal: You can drop this since patching isn't necessary.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,45 @@ temporal workflow reset with-workflow-update-options \
...
+
+```bash
+temporal workflow update-options \
+    --query "WorkflowType="$WORKFLOW_TYPE" AND --TemporalWorkerDeploymentVersion=$TARGET_DEPLOYMENT:$OLD_VERSION" \
+    --workflow-id "$WORKFLOW_ID" \
+    --versioning-override-behavior auto_upgrade \
+    --versioning-override-deployment-name "$TARGET_DEPLOYMENT"
+```
+
+:::note
```
ShahabT, line 428: Maybe we want to clarify the following point?

Upon changing the behavior to auto-upgrade, the workflow will resume work on the Current version. So if the Current version is different from the previous pinned version of the workflow, they should be compatible (i.e. patching should have been use as needed). Otherwise, the workflow will be stuck on the Current version.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,45 @@ temporal workflow reset with-workflow-update-options \
...
+    --versioning-override-behavior auto_upgrade \
+    --versioning-override-deployment-name "$TARGET_DEPLOYMENT"
+```
+
+If you want to move all of your Workflows of a certain type to this new configuration, you can do it with this command:
+
+```bash
+temporal workflow update-options \
+    --query="WorkflowType="$WORKFLOW_TYPE" \
+    --workflow-id "$WORKFLOW_ID" \
```
ShahabT: --workflow-id and --query cannot be passed together. (remove --workflow-id)

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,45 @@ temporal workflow reset with-workflow-update-options \
...
+    --versioning-override-behavior auto_upgrade \
+    --versioning-override-deployment-name "$TARGET_DEPLOYMENT"
+```
+
+You can also filter on a certain build ID to limit the number of Workflows you apply it to:
+
+```bash
+temporal workflow update-options \
+    --query "WorkflowType="$WORKFLOW_TYPE" AND --TemporalWorkerDeploymentVersion=$TARGET_DEPLOYMENT:$OLD_VERSION" \
+    --workflow-id "$WORKFLOW_ID" \
```
ShahabT: same here.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,45 @@ temporal workflow reset with-workflow-update-options \
     --versioning-override-build-id "$TARGET_BUILD_ID"
 ```
 
+## Migrating a Workflow from Pinned to Auto-Upgrade
+
+There may be times when you need to migrate your Workflow from Pinned to Auto-Upgrade because you configured your Workflow Type with the wrong behavior or you've pinned a really long-running Workflow by mistake.
+
+Pinned Workflows get stuck on versions and Workflows prevent versions from draining, long-running Workflows could be preventing drainage. You could move the Workflow to a new build, but that would just push the problem to the next build.
```
Shivs11: I think "Pinned Workflows get stuck on versions and Workflows prevent versions from draining, long-running Workflows could be preventing drainage." sounds a bit difficult to follow imho

How about something like:
"Pinned workflows may block version drainage, especially when they run for a long time.”

Open to also hearing your suggestions if you have any!

docs/production-deployment/worker-deployments/worker-versioning.mdx
flippedcoder: Thanks for helping me add more detail to this part!

docs/production-deployment/worker-deployments/worker-versioning.mdx
flippedcoder: I like this wording. I think it does help clarify what's happening more here.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,43 @@ temporal workflow reset with-workflow-update-options \
...
+    --workflow-id "$WORKFLOW_ID" \
+    --versioning-override-behavior auto_upgrade \
+    --versioning-override-deployment-name "$TARGET_DEPLOYMENT"
+```
+
+If you want to move all of your Workflows of a certain type to this new configuration, you can do it with this command:
+
+```bash
+temporal workflow update-options \
+    --query="WorkflowType="$WORKFLOW_TYPE" \
```
drewhoskins-temporal: Syntax issue.  There are three quotes on this line.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,43 @@ temporal workflow reset with-workflow-update-options \
...
+    --versioning-override-behavior auto_upgrade \
+    --versioning-override-deployment-name "$TARGET_DEPLOYMENT"
```
drewhoskins-temporal: I believe this isn't necessary, here or below.   I tried without it and it seemed fine.  @ShahabT can confirm that there's no advantage to using it.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,43 @@ temporal workflow reset with-workflow-update-options \
...
+```
+
+You can also filter on a certain build ID to limit the number of Workflows you apply it to:
+
+```bash
+temporal workflow update-options \
```
drewhoskins-temporal, line 423: This gives me syntax errors as well when I try to run it.

docs/production-deployment/worker-deployments/worker-versioning.mdx
drewhoskins-temporal: Actually it would be on the Workflow's Target Version, a term the team aligned on earlier this week.  I'm prepping a docs change to add that concept to our encyclopedia.

docs/production-deployment/worker-deployments/worker-versioning.mdx
```diff
@@ -395,6 +395,43 @@ temporal workflow reset with-workflow-update-options \
...
+    --query "WorkflowType="$WORKFLOW_TYPE" AND --TemporalWorkerDeploymentVersion=$TARGET_DEPLOYMENT:$OLD_VERSION" \
+    --versioning-override-behavior auto_upgrade \
+    --versioning-override-deployment-name "$TARGET_DEPLOYMENT"
+```
+
+:::note
+When you change the behavior to Auto-Upgrade, the Workflows will resume work on the Current version. So if the Current version is different from the previous Pinned version of the Workflows, you should make sure you [patch](/patching#patching) the Workflows. Patching ensures your Workflows are replay-safe to support multiple versions after the migration. Otherwise, the Workflow will can get stuck on the Current version.
```
drewhoskins-temporal: > Patching ensures your Workflows are replay-safe to support multiple versions after the migration.  Otherwise, the Workflow will can get stuck on the Current version.

I found this wording a bit confusing.  I would vote we don't re-explain this here, as it's already explained/wordsmithed in the link.

If you want to keep it, suggest

> Patching ensures your Workflows are replay-safe to support the new version.

docs/production-deployment/worker-deployments/worker-versioning.mdx
flippedcoder: Went through and tested this locally. Made updates to get it working.

docs/production-deployment/worker-deployments/worker-versioning.mdx
flippedcoder: Went through and tested this locally. Made updates to get it working. We didn't need the `--versioning-override-deployment-name "$TARGET_DEPLOYMENT"` part of the command.

docs/production-deployment/worker-deployments/worker-versioning.mdx
flippedcoder: Went through and tested this locally. Made updates to get it working. We didn't need the `--versioning-override-deployment-name "$TARGET_DEPLOYMENT"` part of the command.

docs/production-deployment/worker-deployments/worker-versioning.mdx
flippedcoder: I'll get rid of it to keep things clear.

drewhoskins-temporal (approved):
lovely, thanks!

PR #3967, opened 2025-10-28
Fixing a typo about multi-cloud replication being GA

lukeknep:
Multi-cloud Replication is still in preview, as per the previous comment.

Multi-region Replication is GA.


jsundai (approved):

jsundai:
Thanks, @lukeknep!

PR #3970, opened 2025-10-31
Go: Add details on Heartbeating after context cancellation

yuandrew:
## What does this PR do?

Add a blurb on what happens when you try to heartbeat after the activity context has been cancelled. Because we return an error, it's not obvious that the heartbeat has still been sent to server.

This came up from https://github.com/temporalio/sdk-go/issues/2064


brianmacdonald-temporal (commented):
Change looks good

PR #3973, opened 2025-11-03
Python: Add sandbox import notification policy documentation

VegetarianOrc:
## What does this PR do?

Adds documentation about the new Python sandbox setting `import_notification_policy` that was added in temporalio/sdk-python#1187.

## Notes to reviewers

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,8 @@
 * @temporalio/education
+
+# PM owners of sensitive pages
+docs/evaluate/temporal-cloud/pricing.mdx @tlotemporal
+docs/evaluate/temporal-cloud/support.mdx @tlotemporal
+docs/evaluate/temporal-cloud/sla.mdx @bechols
+docs/evaluate/temporal-cloud/limits.mdx @bechols
+docs/evaluate/temporal-cloud/service-availability.mdx @bechols
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-yarn run lint && yarn run fix-vale
+# yarn run lint && yarn run fix-vale
--- a/docs/cli/config.mdx
+++ b/docs/cli/config.mdx
@@ -78,7 +78,7 @@ The command execution timeout. 0s means no timeout.
 
 **--config-file** _string_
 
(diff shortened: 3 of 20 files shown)


docs/develop/python/python-sdk-sandbox.mdx
```diff
@@ -195,6 +196,55 @@ my_restrictions = dataclasses.replace(
 my_worker = Worker(..., workflow_runner=SandboxedWorkflowRunner(restrictions=my_restrictions))
 ```
 
+### Import Notification Policy
+
+The sandbox's import notification policy specifies how the sandbox behaves when it imports modules in a way that may be unintentional. It covers two common scenarios, dynamic imports and enforcing module passthrough, and allows for controlling each separately.
```
jsundai:
```suggestion
The sandbox's import notification policy specifies how the sandbox behaves when it imports modules in a way that may be unintentional. It covers two common scenarios: dynamic imports and enforcing module passthrough. Each can be controlled independently.
```
Perhaps breaking up this sentence into something like this?

docs/develop/python/python-sdk-sandbox.mdx
```diff
@@ -195,6 +196,55 @@ my_restrictions = dataclasses.replace(
...
+# my_worker_file.py
+
+from temporalio.worker import Worker
+from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner, SandboxRestrictions
+
+my_worker = Worker(
+  ...,
+  workflow_runner=SandboxedWorkflowRunner(
+    restrictions=SandboxRestrictions.default.with_import_notification_policy( 
+      workflow.SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH | workflow.SandboxImportNotificationPolicy.WARN_ON_UNINTENTIONAL_PASSTHROUGH
```
jsundai: This appears twice or maybe it's intentional?

VegetarianOrc: It wasn't intentional! Thanks for catching. I've updated the example to use the two warning types instead of the same one twice.

jsundai: Ok perfect. Taking a look at this today.

PR #3984, opened 2025-11-12
Remove test scope from spring maven dependency

samiconductor:
## What does this PR do?

Removes `test` scope from Spring maven dependency. It should be the default `compile` scope.


brianmacdonald-temporal (approved):
Good catch; thanks.

PR #3986, opened 2025-11-12
Update GCP documentation so it can be referenced for the api

david049:
## What does this PR do?
adds instructions for setup on GCP audit log sink

### DIFF
--- a/docs/production-deployment/cloud/audit-logging-gcp.mdx
+++ b/docs/production-deployment/cloud/audit-logging-gcp.mdx
@@ -40,12 +40,12 @@ Before configuring the manual Audit Log sink, complete the following steps in Go
 1. If you are using the UI, follow the instructions in the Temporal Cloud UI to configure the permissions for that service account.
 1. If you are using the API, you should set up the permissions as follows, or use the [Terraform Module](https://github.com/temporalio/terraform-modules/tree/main/modules/auditlog-sa):
    - Grant the role **Pub/Sub Publisher** to the service account from the PubSub topic.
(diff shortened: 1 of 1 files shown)


docs/production-deployment/cloud/audit-logging-gcp.mdx
```diff
@@ -36,8 +36,18 @@ If you use Terraform for your deployment, you don't need to complete these prere
 Before configuring the manual Audit Log sink, complete the following steps in Google Cloud:
 
 1. Create a Pub/Sub topic and make a note of its topic name, such as `test-auditlog`.
-1. Set up a service account in the same project in Google Cloud and follow the instructions in the
-   Temporal Cloud UI to configure the permissions for that service account.
+1. Set up a service account in the same project in Google Cloud.
+1. If you are using the UI, follow the instructions in the Temporal Cloud UI to configure the permissions for that service account.
+1. If you are using the API, you should set up the permissions as follows, or use the [Terraform Module](https://github.com/temporalio/terraform-modules/tree/main/modules/auditlog-sa):
+   - Grant the role **Pub/Sub Publisher** to the service account from the PubSub topic.
```
alice-yin, line 42: I would suggest not adding detailed instruction on the doc. We also don't want to leak our internal accounts such as prod-vs4padk9dkczcey6d6st9c58e.iam.gserviceaccount.com. 

It's not sensitive information but also not good just expose on the internet.

alice-yin: Currently, only temporal customers could access it.

david049: hm okay, we are already leaking in the image on the page 😓- in this case unsure how we should convey what information an API user needs

PR #3996, opened 2025-11-13
docs: add TS to envconfig

lennessyy:
## What does this PR do?

- Add TS to envconfig documentation

### DIFF
--- a/docs/develop/environment-configuration.mdx
+++ b/docs/develop/environment-configuration.mdx
@@ -27,7 +27,7 @@ refer to [Temporal Client Environment Configuration Reference](../references/cli
 
 Environment configuration is in
 [Public Preview](../evaluate/development-production-features/release-stages.mdx#public-preview) in the Temporal Go,
-Python, Ruby, and .NET SDKs, as well as the Temporal CLI.
+Python, Ruby, TypeScript, and .NET SDKs, as well as the Temporal CLI.
 
 :::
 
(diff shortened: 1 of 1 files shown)


docs/develop/environment-configuration.mdx
```diff
@@ -330,6 +330,61 @@ public static class LoadFromFile
...
+    profile: profileName,
+    configSource: { path: configFile },
+  });
+
+  // Override the target host to the correct address.
+  // This is the recommended way to override configuration values.
+  config.connectionOptions.address = 'localhost:7233';
+
+  console.log(`\nLoaded '${profileName}' profile from ${configFile} with overrides.`);
+  console.log(`  Address: ${config.connectionOptions.address} (overridden from localhost:9999)`);
```
THardy98: I don't think we need to do this override bit, I would follow the simpler sample here, imo.

docs/develop/environment-configuration.mdx
```diff
@@ -570,4 +625,62 @@ public static class LoadProfile
...
+
+  // Load the 'staging' profile.
+  const config = loadClientConnectConfig({
+    profile: profileName,
+    configSource: { path: configFile },
+  });
+
+  // Override the target host to the correct address.
+  // This is the recommended way to override configuration values.
+  config.connectionOptions.address = 'localhost:7233';
```
THardy98: Same here - unless we explicitly want to document this and have precedent
That being said - I still think we don't need to demonstrate this twice, I think we typically have one simpler sample than a more complex one with the override?

docs/develop/typescript/temporal-client.mdx
```diff
@@ -0,0 +1,831 @@
...
+  console.log('\nAttempting to connect to client...');
+  try {
+    const connection = await Connection.connect(config.connectionOptions);
+    const client = new Client({ connection, namespace: config.namespace });
+    console.log('✅ Client connected successfully!');
+    await connection.close();
+  } catch (err) {
+    console.log(`❌ Failed to connect: ${err}`);
+  }
+}
```
THardy98, line 166: This is what I would use as a simpler sample, for my previous comment

PR #4004, opened 2025-11-18
Added definitions for worker deployment versions in encyclopedia

flippedcoder:
## What does this PR do?

This moves the bigger explanations of the three worker deployment version types to the encyclopedia page.

### DIFF
--- a/docs/production-deployment/worker-deployments/worker-versioning.mdx
+++ b/docs/production-deployment/worker-deployments/worker-versioning.mdx
@@ -81,11 +81,9 @@ To get started with Worker Versioning, you should understand some concepts aroun
   - Pinned Workflows are designed for use with rainbow deployments. See [Deployment Systems](#deployment-systems).
   - Pinned Workflows don't need to be patched, as they run on the same worker and build until they complete.
   - If you expect your Workflow to run longer than you want your Worker Deployment Versions to exist, you should mark your Workflow Type as Auto-Upgrade.
-- Each Worker Deployment has a single **Current Version** which is where Workflows are routed to unless they were previously pinned on a different version.
-- Each Worker Deployment can have a **Ramping Version** which is where a configurable percentage of Workflows are routed to unless they were previously pinned on a different version.
-- For a given Workflow, its **Target Worker Deployment Version** is the version it will move to next.
-
-You can find more details on the versioning types above in the [Worker Versioning encyclopedia article](/worker-versioning#versioning-definitions).
+- Each Worker Deployment has a single [**Current Version**](/worker-versioning#versioning-definitions) which is where Workflows are routed to unless they were previously pinned on a different version.
+- Each Worker Deployment can have a [**Ramping Version**](/worker-versioning#versioning-definitions) which is where a configurable percentage of Workflows are routed to unless they were previously pinned on a different version.
+- For a given Workflow, its [**Target Worker Deployment Version**](/worker-versioning#versioning-definitions) is the version it will move to next.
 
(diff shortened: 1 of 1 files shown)


docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -48,6 +49,12 @@ A **Pinned** Workflow is guaranteed to complete on a single Worker Deployment Ve
 
 An **Auto-Upgrade** Workflow will move to the latest Worker Deployment Version automatically whenever you change the current version. Auto-upgrade Workflows are not restricted to a single Deployment Version and need to be kept replay-safe manually, i.e. with [patching](/workflow-definition#workflow-versioning).
 
+## Versioning Definitions
+
+- **Current Worker Deployment Version**: the version where Workflows are routed to unless they were previously pinned on a different version. Other versions can continue polling to allow pinned Workflows to finish executing or in case you need to roll back. If no current version is specified, the default is unversioned.
```
jsundai:
```suggestion
- **Current Worker Deployment Version**: The version where Workflows are routed to unless they were previously pinned on a different version. Other versions can continue polling to allow pinned Workflows to finish executing or in case you need to roll back. If no current version is specified, the default is unversioned.
```

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -48,6 +49,12 @@ A **Pinned** Workflow is guaranteed to complete on a single Worker Deployment Ve
...
+- **Ramping Worker Deployment Version**: the version where a configurable percentage of Workflows are routed to unless they were previously pinned on a different version. The ramp percentage can be in the range [0, 100]. Workflows that don't go to the Ramping Version will go to the Current Version. If no Ramping Version is specified, 100% of new Workflows and Auto-Upgrade Workflows will go to the Current Version.
```
jsundai:
```suggestion
- **Ramping Worker Deployment Version**: The version where a configurable percentage of Workflows are routed to unless they were previously pinned on a different version. The ramp percentage can be in the range [0, 100]. Workflows that don't go to the Ramping Version will go to the Current Version. If no Ramping Version is specified, 100% of new Workflows and Auto-Upgrade Workflows will go to the Current Version.
```

docs/encyclopedia/workers/worker-versioning.mdx
```diff
@@ -48,6 +49,12 @@ A **Pinned** Workflow is guaranteed to complete on a single Worker Deployment Ve
...
+- **Target Worker Deployment Version**: the version your Workflow will move to next. This could be the Deployment's Current Version or the Ramping Version. For example, if an Auto-Upgrade Workflow was running on Version A, the Current Version is B, and there is a 5% ramp to C, there is a 95% chance that its Target Version is B and 5% that it's C.
```
jsundai:
```suggestion
- **Target Worker Deployment Version**: The version your Workflow will move to next. This could be the Deployment's Current Version or the Ramping Version. For example, if an Auto-Upgrade Workflow was running on Version A, the Current Version is B, and there is a 5% ramp to C, there is a 95% chance that its Target Version is B and 5% that it's C.
```

docs/encyclopedia/workers/worker-versioning.mdx
jsundai: I'd link off the bolded definitions to the encyclopedia pages too!

flippedcoder: Ooooo! I like that suggestion!

PR #4024, opened 2025-11-26
Add Worker heartbeating explanation

dustin-temporal:
## What does this PR do?
Document the new Worker heartbeating feature. It is only available in the Python SDK for now, so adding examples for that. As new SDKs are added, we'll add examples here.

### DIFF
--- a/docs/production-deployment/cloud/worker-health.mdx
+++ b/docs/production-deployment/cloud/worker-health.mdx
@@ -397,6 +397,12 @@ rate(temporal_sticky_cache_total_forced_eviction_total{namespace="$namespace"}[5
 
 ## Manage Worker Heartbeating {#manage-worker-heartbeating}
 
+:::tip SUPPORT, STABILITY, and DEPENDENCY INFO
+
+This feature is currently in [Public Preview](/evaluate/development-production-features/release-stages#public-preview).
+
+:::
+
 Workers send a heartbeat to Temporal Server every 60 seconds by default. This heartbeat serves to provide liveness and configuration data from the Worker to the Server.
 Specific data sent can be found in the [API](https://github.com/temporalio/api/blob/master/temporal/api/worker/v1/message.proto). By providing a consistent heartbeat from
 Workers, the Server can obtain an accurate count of Workers, understand Worker performance, and respond to Worker heartbeats with commands. Some examples of how this is useful:


docs/production-deployment/cloud/worker-health.mdx
```diff
@@ -394,3 +394,43 @@ max_over_time(temporal_sticky_cache_size{namespace="$namespace"}[10m])
...
+
+## Manage Worker Heartbeating {#manage-worker-heartbeating}
+
+Workers send a heartbeat to Temporal Server every 60 seconds by default. This heartbeat serves to provide liveness and configuration data from the Worker to the Server.
+Specific data sent can be found in the [API](https://github.com/temporalio/api/blob/master/temporal/api/worker/v1/message.proto). By providing a consistent heartbeat from
+Workers, the Server can obtain an accurate count of Workers, understand Worker performance, and respond to Worker heartbeats with commands. Some examples of how this is useful:
+
+- understanding the difference between a Worker that is down and a Worker that is processing tasks for a long time
+- enabling Activity cancellation without the need to implement bespoke heartbeat logic
+- identifying a Worker with high CPU usage fom the Server point of view
```
yuandrew:
```suggestion
- identifying a Worker with high CPU usage from the Server point of view
```

docs/production-deployment/cloud/worker-health.mdx
```diff
@@ -394,3 +394,43 @@ max_over_time(temporal_sticky_cache_size{namespace="$namespace"}[10m])
...
+
+If you wish to disable Worker heartbeating or set heartbeating to be more frequent than every 60 seconds, set the configuration relevant to your SDK.
```
yuandrew: Might be worth mentioning the current limits to heartbeating are 1s and 60s

docs/production-deployment/cloud/worker-health.mdx
```diff
@@ -394,3 +394,43 @@ max_over_time(temporal_sticky_cache_size{namespace="$namespace"}[10m])
...
 ```
+
+## Manage Worker Heartbeating {#manage-worker-heartbeating}
+
+Workers send a heartbeat to Temporal Server every 60 seconds by default. This heartbeat serves to provide liveness and configuration data from the Worker to the Server.
+Specific data sent can be found in the [API](https://github.com/temporalio/api/blob/master/temporal/api/worker/v1/message.proto). By providing a consistent heartbeat from
+Workers, the Server can obtain an accurate count of Workers, understand Worker performance, and respond to Worker heartbeats with commands. Some examples of how this is useful:
+
+- understanding the difference between a Worker that is down and a Worker that is processing tasks for a long time
+- enabling Activity cancellation without the need to implement bespoke heartbeat logic
```
yuandrew: Should we mention this is not implemented yet?

yuandrew (approved):
small things, but looks good otherwise!

docs/production-deployment/cloud/worker-health.mdx
dustin-temporal: I will drop it from here in case it is trickier than we think it will be actually.

PR #4030, opened 2025-12-02
Contact Temporal Support to remove a replica from Namespace with API keys

lukeknep:
## What does this PR do?

If a Namespace has API keys and replication enabled, and wants to remove the replica, they must switch to regional endpoints and contact Temporal Support to validate that the switch has been done. This was not called out in the docs before. This PR calls it out and explains _why_ this is required.

PR #4033, opened 2025-12-02
added clarifications for API key + high avail

GSmithApps:
Adds notes to doc [here](https://docs.temporal.io/cloud/namespaces#access-namespaces) and [here](https://docs.temporal.io/cloud/api-keys#namespace-authentication)

Piggybacks on [this pr](https://github.com/temporalio/documentation/pull/4030)

### DIFF
--- a/docs/production-deployment/cloud/get-started/api-keys.mdx
+++ b/docs/production-deployment/cloud/get-started/api-keys.mdx
@@ -341,11 +341,9 @@ Workflow access secure.
 ## API keys for Namespace authentication {#namespace-authentication}
 
 Create a Namespace with API key authentication as an alternative to mTLS-based authentication by selecting "Allow API
-key authentication" during setup. The gRPC endpoint format for the Namespace depends on the authentication method and whether or not High Availability features are enabled:
+key authentication" during setup. The gRPC endpoint format for the Namespace depends on the authentication method and whether or not High Availability features are enabled.
 
-- For API key connections on Namespaces without High Availability features, use the gRPC regional endpoint `<region>.<cloud_provider>.api.temporal.io:7233`.
-- For API Key connections on [Namespaces with High Availability features](/cloud/high-availability), use the
-gRPC Namespace endpoint: `<namespace>.<account>.tmprl.cloud:7233`. This allows automated failover without needing to configure worker(s) with both endpoints. These endpoints will still work, but in the event of a failover, there is a brief period of unavailability between your workers and Temporal server. 
+See the following documentation for [accessing Namespaces](/cloud/namespaces#access-namespaces) for more information.
 
 :::info
 
@@ -354,7 +352,13 @@ See [Disable High Availability](/cloud/high-availability/enable#disable) for mor
 
 :::
 
-See the following documentation for [accessing Namespaces](/cloud/namespaces#access-namespaces) for more information.
+### Without High Availability features
+
+Use the gRPC regional endpoint `<region>.<cloud_provider>.api.temporal.io:7233`.
+
+### With High Availability features
+
+Use the gRPC Namespace endpoint: `<namespace>.<account>.tmprl.cloud:7233`. This allows automated failover without needing to configure worker(s) with both endpoints. These endpoints will still work, but in the event of a failover, there is a brief period of unavailability between your workers and Temporal server. 
 
 ## Use API keys to authenticate {#using-apikeys}
 


docs/production-deployment/cloud/get-started/api-keys.mdx
```diff
@@ -341,17 +341,17 @@ Workflow access secure.
 ## API keys for Namespace authentication {#namespace-authentication}
 
 Create a Namespace with API key authentication as an alternative to mTLS-based authentication by selecting "Allow API
-key authentication" during setup. The gRPC endpoint format for the Namespace depends on the authentication method:
+key authentication" during setup. The gRPC endpoint format for the Namespace depends on the authentication method and whether or not High Availability features are enabled:
```
laniehei: It might be cleaner to completely separate "HA" and "Normal Namespaces" instead of grouping them together. Single tenant namespaces will not be concerned with what happens with HA and vice versa when reading the docs

docs/production-deployment/cloud/get-started/api-keys.mdx
```diff
@@ -341,17 +341,17 @@ Workflow access secure.
...
 
-- For API key connections, use the gRPC regional endpoint `<region>.<cloud_provider>.api.temporal.io:7233`.
-
-Use this gRPC endpoint in the Temporal CLI or SDK to connect to Temporal Cloud with an API key.
+- For API key connections on Namespaces without High Availability features, use the gRPC regional endpoint `<region>.<cloud_provider>.api.temporal.io:7233`.
+- For API Key connections on [Namespaces with High Availability features](/cloud/high-availability), use the
+gRPC Namespace endpoint: `<namespace>.<account>.tmprl.cloud:7233`. This allows automated failover without needing to
+switch endpoints.
```
laniehei:
```suggestion
gRPC Namespace endpoint: `<namespace>.<account>.tmprl.cloud:7233`. This allows automated failover without needing to configure worker(s) with both endpoints. These endpoints will still work, but in the event of a failover, there is a brief period of unavailability between your workers and Temporal server. 
```

docs/production-deployment/cloud/get-started/namespaces.mdx
```diff
@@ -251,14 +251,29 @@ Sample workflowId convention:
...
 
-- For the API key authentication method, use the gRPC regional endpoint `<region>.<cloud_provider>.api.temporal.io:7233`.
-- For the mTLS authentication method, use the gRPC Namespace endpoint `<namespace>.<account>.tmprl.cloud:7233`.
-- For [Namespaces with High Availability features](/cloud/high-availability) use the gRPC Namespace endpoint `<namespace>.<account>.tmprl.cloud:7233`, regardless of your authentication method.
-  This allows automated failover without needing to switch your endpoint.
+|                        | Not High Availability | High Availability                                                                |
+|------------------------|-----------------------|----------------------------------------------------------------------------------|
+| mTLS Authentication    | Namespace             | Namespace                                                                        |
+| API Key Authentication | Regional              | Both work, but Namespace is recommended because it allows for automated failover |
```
laniehei:
```suggestion
| API Key Authentication | Regional             | Both work, but Namespace is recommended because it reduces the unavailability window during a failover event |
```

docs/production-deployment/cloud/get-started/api-keys.mdx
GSmithApps: @laniehei are you okay if we leave it as it is? I feel like they’re already separated in the bullets

If you have a different preference, please feel free to suggest a change

laniehei: if you have strong opinions or would like to just not do this in this step, comfortable with leaving for now. 

as a user of docs, though, I don't want to have to read a usecase that doesn't pertain to me (and if someone doesn't know what HA is this is also confusing because they have to determine whether it pertains to them by reading the entire bullet point), so cleaner to separate them into "For single namespace" and "For HA" imo

GSmithApps: Figured out a nice way to implement this comment, and discussed with Lanie off-pr -- lgth 👍

docs/production-deployment/cloud/get-started/namespaces.mdx
```diff
@@ -251,14 +251,29 @@ Sample workflowId convention:
 Temporal Cloud normally supports authentication to Namespaces using [API keys](/cloud/api-keys) _or_ [mTLS](/cloud/certificates).
 If you need to migrate from one authentication method to another, or you require both API key and mTLS authentication to be enabled on your Namespace, please contact [Support](https://docs.temporal.io/cloud/support#support-ticket).
 
+:::info
+
+Requiring both API key and mTLS is in [public preview](/evaluate/development-production-features/release-stages), and doesn't currently support [High Availability features](/cloud/high-availability).
```
lukeknep, line 256: private preview, actually. I will fix after merge

PR #4038, opened 2025-12-02
Banner indicating SCIM and SAML are paid features

bechols:
## What does this PR do?

Adds banners indicating SCIM and SAML are paid features, and links to the pricing page.

Linking to the pricing page rather than adding details to the actual docs page so we keep the definitions in one place.


flippedcoder (approved):
Thanks for the updates!

PR #4092, opened 2025-12-23
fix foreground/background logic in operation list

bechols:
## What does this PR do?

Reverses the logic so that operations are correctly displayed as foreground or background.

### DIFF
--- a/docs/encyclopedia/retry-policies.mdx
+++ b/docs/encyclopedia/retry-policies.mdx
@@ -49,11 +49,36 @@ Try out the [Activity retry simulator](/develop/activity-retry-simulator) to vis
 
 ## Default behavior
 
-- **Workflow Execution:** When a Workflow Execution is spawned, it is not associated with a default Retry Policy and thus does not retry by default.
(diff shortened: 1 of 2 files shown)


src/components/formatting/OperationsTable.js
```diff
@@ -2,114 +2,114 @@ import React from "react";
...
-    }
\ No newline at end of file
+export default function OperationsTable() {
+    return (
+        <DocsTable Columns={["Operation", "Description", "Priority", "Effect of Throttling"]} sortable={true}>
+            {operations.map((operation) => (
+                <DocsTableRow key={operation.operation}>
+                    <NewDocsCell>{operation.operation}</NewDocsCell>
+                    <NewDocsCell>{operation.description}</NewDocsCell>
+                    <NewDocsCell>{operation.is_background == "TRUE" ? "Background" : "Foreground"}</NewDocsCell>
```
bechols, line 109: This is the relevant change - everything else is just spacing the editor auto-applied.

PR #4109, opened 2026-01-09
Update APS guide with TRUs

bechols:
## What does this PR do?

Add Provisioned Capacity/TRU info on the APS limits best practice page


tlotemporal (approved):
makes sense

PR #4080, opened 2025-12-17
feat(task queues): created initial draft for priority and fairness

flippedcoder:
## What does this PR do?

This adds a new page to explain Task Queue priority and fairness.

PR #4123, opened 2026-01-14
Add Nexus adoption decision tree to evaluate nexus page

jsundai:

### DIFF
--- a/docs/develop/dotnet/temporal-nexus.mdx
+++ b/docs/develop/dotnet/temporal-nexus.mdx
@@ -375,15 +375,24 @@ This will show the two workflows started and their results.
 
 ### Canceling a Nexus Operation {#canceling-a-nexus-operation}
 
-<!--TODO: Add .NET cancellation type information (like Java) once available -->
-
 To cancel a Nexus Operation from within a Workflow, cancel the cancellation token passed to the operation call. Only asynchronous operations can be canceled in Nexus, since cancellation is sent using an operation token.
 The Workflow or other resources backing the operation may choose to ignore the cancellation request.
 If ignored, the operation may enter a terminal state.
 
+When a Nexus operation is started, the caller can specify different cancellation types that control how the caller reacts to cancellation:
+
+- `Abandon` - Do not request cancellation of the operation.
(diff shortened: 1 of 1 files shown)


docs/evaluate/development-production-features/temporal-nexus.mdx
```diff
@@ -49,6 +49,14 @@ Nexus supports cross-team, cross-domain, cross-namespace, multi-region, and mult
 
 Unlike other forms of inter-service communication, Nexus combines a familiar programming model with the resiliency of the Temporal Platform and its queue-based Worker architecture.
 
+Use the following decision tree to help determine if Nexus is right for your use case:
+
+<div style={{textAlign: 'center', margin: '2rem 0'}}>
```
flippedcoder, line 55: What do you think about making this a component? We could definitely reuse this!

static/diagrams/nexusadoptiontree.png
flippedcoder, line 1: Couple comments on the image:
- Does `namespaces` need to be capitalized to match other parts of the docs?
- I don't quite understand what `Is cross-x support is a future goal and not time critical?` means? Could this be rephrased?

lennessyy, line 1: i don't think we should use excalifont -- they are hard to read and not super accessible. do we have a style guide for diagrams? we should make one if we don't to make sure they are consistent across the site 

this diagram is already complex, so i think we should opt for cleaner fonts

jsundai: Yeah I agree. I put in a request to design about it awhile ago especially for a light/dark mode version.

docs/evaluate/development-production-features/temporal-nexus.mdx
flippedcoder: Just fyi, I went ahead and made [this a component](https://github.com/temporalio/documentation/pull/4080/commits/a7cc7600abe74b4efe809c2e24aa51094058efe8) in one of my PRs because it was really useful!

jsundai: Added the component into the PR

static/diagrams/nexusadoptiontree.png
jsundai: I'll keep what the original has but yeah I think we should have a style guide for diagrams for future guidance to folks!

PR #4144, opened 2026-01-23
Clarify message idempotency section

dandavison:
## What does this PR do?
Clarifies the section on idempotency keys and CAN. There were reports that for the following sentence

> Inside your message handler, you can check your idempotency key--the Update ID or the one you provided to the Signal--to check whether the Workflow has already handled the update.

it was not obvious that it referred to the situation when the workflow is using CAN.


docs/encyclopedia/workflow-message-passing/handling-messages.mdx
```diff
@@ -109,19 +109,15 @@ If you don’t need to ensure that your handlers complete, you may specify your
...
 
-When your workflow Continues-As-New, you should handle deduplication yourself in your message handler. This is because Temporal's built-in deduplication doesn't work across [Continue-As-New](/workflow-execution/continue-as-new) boundaries, meaning you would risk processing messages twice for such Workflows if you don't check for duplicate messages yourself.
+However, if you are using Updates with [Continue-As-New](/workflow-execution/continue-as-new) you should implement the deduplication in your Workflow code, since Update ID deduplication by the server is per Workflow run.
 
-To deduplicate in your message handler, you can use an idempotency key.
+(In addition to these application-level identifiers, both Signals and Updates automatically use request IDs to deduplicate retried client calls. You do not need to do anything to enable this.)
```
stephanos: Don't mind dropping the parentheses.

stephanos: On second thought, if the user can't change/do anything about it, do we need to explain this here at all?

docs/encyclopedia/workflow-message-passing/handling-messages.mdx
```diff
@@ -109,19 +109,15 @@ If you don’t need to ensure that your handlers complete, you may specify your
 
 See the links below for how to ensure handlers are finished in your SDK.
 
-#### Ensuring your messages are processed exactly once {#exactly-once-message-processing}
+#### Message IDs and handling Continue-As-New {#exactly-once-message-processing}
 
-Many developers want their message handlers to run exactly once--to be idempotent--in cases where the same Signal or Update is delivered twice or sent by two different call sites. Temporal deduplicates messages for you on the server, but there is one important case when you need to think about this yourself when authoring a Workflow, and one when sending Signals and Updates.
+Usually, you'll want your message handlers to run exactly once--to be idempotent--in cases where the same Signal or Update is delivered twice. For Updates, Temporal handles this for you on the server, by deduplicating according to the Update ID. For Signals, you should use a custom idempotency key that you send as part of your own signal inputs, implementing the deduplication in your Workflow code. The Update ID is set automatically to a UUID, but you can set it yourself.
```
stephanos: I'd move the last sentence about UUID next to the Update sentence. The Signal sentence breaks the flow there in the middle.

docs/encyclopedia/workflow-message-passing/handling-messages.mdx
```diff
@@ -109,19 +109,15 @@ If you don’t need to ensure that your handlers complete, you may specify your
 
 See the links below for how to ensure handlers are finished in your SDK.
 
-#### Ensuring your messages are processed exactly once {#exactly-once-message-processing}
+#### Message IDs and handling Continue-As-New {#exactly-once-message-processing}
```
stephanos, line 112: I like that this is focused on CaN now. The part about the same Update/Signal coming in from different clients is now only implied. That might be fine.

docs/encyclopedia/workflow-message-passing/handling-messages.mdx
drewhoskins-temporal: I have the same question.

docs/encyclopedia/workflow-message-passing/handling-messages.mdx
drewhoskins-temporal: It may be simpler to just ask people to check idempotency.  And then explain the reasoning that it's necessary for signals and update plus CaN .  
That way, if they add CaN later, they won't cause a bug.  And I don't think there is much harm to a redundant check.

docs/encyclopedia/workflow-message-passing/handling-messages.mdx
dandavison: So, the two topics are deduplication of framework-level retries, and deduplication of messages that are duplicates according to an application-domain definition of equivalence.

We could get rid of it. The reason I put it there is because I have personally confused them in the past, which leads me to suspect others will. Put differently, whenever request IDs are discussed, idempotency is also. If we tell people that they are on the hook for signal idempotency then I think many people will confuse that with a statement that Temporal doesn't use request IDs for signal.

docs/encyclopedia/workflow-message-passing/handling-messages.mdx
flippedcoder:
```suggestion
Usually, you'll want your message handlers to run exactly once--to be idempotent--in cases where the same Signal or Update is delivered twice. For Updates, Temporal handles this for you on the server, by deduplicating according to the Update ID. The Update ID is set automatically to a UUID, but you can set it yourself.

For Signals, you should use a custom idempotency key that you send as part of your own signal inputs, implementing the deduplication in your Workflow code.
```

docs/encyclopedia/workflow-message-passing/handling-messages.mdx
flippedcoder:
```suggestion
:::info

In addition to these application-level identifiers, both Signals and Updates automatically use request IDs to deduplicate retried client calls. You do not need to do anything to enable this.

:::
```

flippedcoder: What if we have it in an info box? That way the user still gets the info and it's in a callout instead of the main content flow.

PR #4166, opened 2026-01-29
PHP Quickstart: config and autoloading (follow-up to @xepozz's #4151)

jsundai:


docs/develop/php/set-up.mdx
```diff
@@ -70,14 +82,32 @@ Now that you have your build tool installed, create a project to manage your dep
...
 </>
 }>
 
-## Add Temporal PHP SDK Dependencies
+## Add Temporal PHP SDK and Configure Autoloading
+
+Install the Temporal SDK, then add PSR-4 autoloading to your `composer.json` so PHP can find your Workflow and Activity classes.
 
-Now add the Temporal SDK dependencies to your project configuration file.
+Your final `composer.json` should look like this. After updating, run `composer dump-autoload` to regenerate the autoloader.
```
xepozz, line 110: Actually, it is not necessary. Only when "composer optimizations" were applied in the previous run: `composer i --optimize-autoload`
But we did not so, so it's redundant.

PR #4171, opened 2026-01-30
Updating Linux installation instructions

AustinGil:
## What does this PR do?

Replaces references to Homebrew with Snap in the Linux installation documentation.

## Notes to reviewers

To my knowledge, Homebrew is not a common tool found in Linux distros. Snap is much more common, and comes pre-installed in many, so it could remove an extra barrier to entry and feels more familiar to Linux users.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213014066663714">EDU-5850 Updating Linux installation instructions</a>


jsundai:
Thanks for the contribution! Let's keep Homebrew in here in addition to Snap for the instructions since a lot of folks still use it. You can just add the Snap instructions right below the Homebrew one on the pages. Once that’s in, I’m good to approve!

PR #4213, opened 2026-02-13
feat: add cookbook pagination

lennessyy:
## What does this PR do?

- Adds pagination for the cookbook

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213272768514844">EDU-5917 feat: add cookbook pagination</a>


Duncanma:
@lennessyy , my thought so far is "pagination would be useful, but probably not until we hit > 20" ... scrolling isn't that hard, and right now since we are at 13, a page_size of 6 would bring us to 3 pages, which I worry would really reduce discovery of anything past the first 6

lennessyy:
@Duncanma That makes sense. When I opened this PR we had rapidly added 6 recipes in 2 weeks so I was anticipating a larger increase than materialized. We can close this for now and reopen when necessary? Or we can just let it hang around until we have to do pagination

Duncanma:
Closing for now, easy enough to bring back later if we need to

PR #4216, opened 2026-02-19
Go: Note that workers should run in dedicated terminal instances

yuandrew:
## What does this PR do?

Make a note that workers should run in dedicated terminal instances. This was something I remember I misunderstood when I first learned about Temporal. Probably worth having this in the `Core application` section of the docs. I'm also not 100% if other SDKs have this same flow, but also probably worth porting this to other SDKs that share this requirement.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213351077319440">EDU-5920 Go: Note that workers should run in dedicated terminal instances</a>


brianmacdonald-temporal:
Resolved by #4560 ; closing.

PR #4247, opened 2026-02-26
Add all languages to plugin guide

tconley1428:
## What does this PR do?

Adds additional languages to the plugin guide, and removes advanced topics specific to Python.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213457700323410">EDU-5958 Add all languages to plugin guide</a>

### DIFF
--- a/docs/develop/plugins-guide.mdx
+++ b/docs/develop/plugins-guide.mdx
@@ -64,15 +64,14 @@ do.
 ### Built-in Activity
 
 You can provide built-in Activities in a Plugin for users to call from their Workflows. Check out the
-[Activities doc](/activities) for more detail on how these work.
+[Activities](/activities) page for more details on how they work.
 
-You should refer to the
-[best practices for creating Activities](/activity-definition#best-practices-for-defining-activities) when you are
-making Activity plugins.
+Refer to the [best practices for creating Activities](/activity-definition#best-practices-for-defining-activities) when
+you are making Activity plugins.
 
 #### Timeouts and retry policies
 
-Temporal's Activity retry mechanism gives applications the benefits of durable execution. See the
+Temporal's Activity retry mechanism gives applications the benefits of Durable Execution. See the
 [Activity retry policy explanation](/activity-definition#activity-retry-policy) for more details.
 
 <SdkTabs hideUnsupportedLanguages>
@@ -185,13 +184,15 @@ plugin = Temporalio::SimplePlugin.new(
 
 ### Workflow-friendly libraries
 
-You can provide a library with functionality for use within a Workflow if you'd like to abstract away some
-Temporal-specific details for your users. Your library will call elements you include in your Plugin such as Activities,
-Child Workflows, Signals, Updates, Queries, Nexus Operations, Interceptors, Data Converters, and any other code as long
-as it follows these requirements:
+You can provide a library for use within a Workflow if you'd like to abstract away some Temporal-specific details for
+your users. Your library will call elements you include in your Plugin such as Activities, Child Workflows, Signals,
+Updates, Queries, Nexus Operations, Interceptors, Data Converters, and any other code as long as it follows these
+requirements:
 
-- It should be [deterministic](/workflow-definition#deterministic-constraints), running the same way every time it’s
-  executed. Non-deterministic code should go in Activities or Nexus Operations.
+- Any code that runs in the Workflow context must be [deterministic](/workflow-definition#deterministic-constraints),
+  meaning it produces the same commands and results when replayed. For example, don't call system time APIs, generate
+  random values, or perform direct network and file I/O from Workflow-context code; move that work to Activities or
+  Nexus Operations.
 - See [observability](/evaluate/development-production-features/observability) to avoid duplicating observation side
   effects when Workflows replay.
 - Put other side effects inside of Activities or [Local Activities](/local-activity). This helps your Workflow handle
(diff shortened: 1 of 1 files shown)


docs/develop/plugins-guide.mdx
```diff
@@ -118,85 +214,391 @@ Consider adding a Child Workflow when one or more of these conditions applies:
...
+[features/snippets/plugins/plugins.rb](https://github.com/temporalio/features/blob/plugin_snippets/features/snippets/plugins/plugins.rb)
+```rb
+custom_converter = Temporalio::Converters::DataConverter.new(
+  payload_converter: Temporalio::Converters::PayloadConverter.default
+)
 
-plugin = SimplePlugin(
-  data_converter = add_converter
+plugin = Temporalio::SimplePlugin.new(
+  name: 'PluginName',
```
drewhoskins-temporal: Orthogonal but perhaps convenient to tackle here.
I have realized that we need scope names to disambiguate these plugins, and I think customers would like that too for their observability.  i.e. "AgentPlugin" and "AiSDKPlugin" are so generic.
WDYT about recommending `my_scope.MyPlugin` for all of these?  We can suggest package/module/gh repo name as scopes.
However, I don't feel a strong need to ask for cross-language naming consistency.

docs/develop/plugins-guide.mdx
```diff
@@ -205,42 +607,191 @@ Interceptors are middleware that can run before and after various calls such as
...
+  client_interceptors: [SomeClientInterceptor.new],
+  worker_interceptors: [SomeWorkerInterceptor.new]
+)
+```
+    {/* SNIPEND */}
+  </SdkTabs.Ruby>
+</SdkTabs>
 
 ### Special considerations for different languages
```
drewhoskins-temporal, line 785: isn't there bundles for TS?

docs/develop/plugins-guide.mdx
chris-olszewski, line 48: If we want a non-Python example we could link to OTEL TS plugin: https://github.com/temporalio/sdk-typescript/blob/main/packages/interceptors-opentelemetry/src/plugin.ts

chris-olszewski (commented):
Added a quick blurb about the Typescript bundler. I omitted configuring the Typescript bundler via a plugin as I think that might be too far in the weeds for 95%+ of developers.

docs/develop/plugins-guide.mdx
tconley1428: Good idea.

docs/develop/plugins-guide.mdx
```diff
@@ -796,6 +797,33 @@ plugin = SimplePlugin("organization.PluginName", workflow_runner=workflow_runner
 ```
 {/* SNIPEND */}
 
+#### TypeScript
+
+The TypeScript SDK bundles all Workflow code from a single module. This prevents TypeScript Plugins from providing
+built-in Workflows the way other SDKs can. Plugin users must import and re-export any Plugin-provided Workflows from
+their own Workflow module so the bundle includes them.
+
+Workflow interceptors are also bundled, but they aren't limited to a single module. `SimplePlugin` appends any
```
drewhoskins-temporal: Is this actionable/do we need this paragraph?  It sounds like it's basically saying "interceptors are just like in other languages"
Should we instead just make sure that the code in the interceptors snippet is correct and bundle-friendly?

docs/develop/plugins-guide.mdx
```diff
@@ -205,42 +610,219 @@ Interceptors are middleware that can run before and after various calls such as
...
+```ts
+class MyWorkflowClientInterceptor implements WorkflowClientInterceptor {}
+
+class MyActivityInboundInterceptor implements ActivityInboundCallsInterceptor {}
+
+class MyActivityOutboundInterceptor implements ActivityOutboundCallsInterceptor {}
+
+const workflowInterceptorsPath = '';
+
+const plugin = new SimplePlugin({
```
drewhoskins-temporal, line 701: @chris-olszewski is this correct given what you mentioned about bundling of interceptors?

tconley1428: It is accurate. This plugin could be used in the bundleWorkflowCode as shown below

docs/develop/plugins-guide.mdx
tconley1428: It is a little more complicated than that because the plugin which defines interceptors would be used differently depending on whether the user is doing bundling explicitly or leaving it up to the worker.

docs/develop/plugins-guide.mdx
```diff
@@ -205,42 +610,219 @@ Interceptors are middleware that can run before and after various calls such as
...
+  include Temporalio::Client::Interceptor
+
+  def intercept_client(next_interceptor)
+    # Your interceptor implementation
+    next_interceptor
+  end
+end
+
+plugin = Temporalio::SimplePlugin.new(
+  name: 'PluginName',
```
drewhoskins-temporal: Now that we've resolved our Slack convo, let's use `my_library.MyPlugin` as the example nomenclature.  (I'll also be explicit about this in the AI Partner Intake Guide.)

tconley1428: I've done `organization.PluginName` which I think we discussed before. Same difference?

docs/develop/plugins-guide.mdx
```diff
@@ -118,85 +215,393 @@ Consider adding a Child Workflow when one or more of these conditions applies:
 
 Any Workflow can be run as a standalone Workflow or as a Child Workflow, so registering a Child Workflow in a `SimplePlugin` is the same as registering any Workflow.
 
-Here is an example with Python:
-
-```python
+<SdkTabs hideUnsupportedLanguages>
```
drewhoskins-temporal, line 250: Looks like this needs a TS sample.  Maybe move the snippet from the special considerations section to here?
The special considerations section can say like "Note how in <link to build in workflows section> the client code is adding the plugin to the bundler as well.

docs/develop/plugins-guide.mdx
drewhoskins-temporal: Okay, I'd start with the prescriptive advice -- plugin users should include the plugin in their bundle, show that code -- and then explan why.

docs/develop/plugins-guide.mdx
```diff
@@ -205,42 +610,219 @@ Interceptors are middleware that can run before and after various calls such as
...
-  return runner
 
-SimplePlugin(..., workflow_runner=workflow_runner)
+plugin = SimplePlugin("organization.PluginName", workflow_runner=workflow_runner)
+```
+{/* SNIPEND */}
+
+#### TypeScript
+
+The TypeScript SDK bundles all Workflow code from a single module. This prevents TypeScript Plugins from providing
```
drewhoskins-temporal: I'd start with "Typescript bundles cannot provide built-in Workflows" and then explain why.

docs/develop/plugins-guide.mdx
chris-olszewski: Yes, this is correct

drewhoskins-temporal (approved):

PR #4331, opened 2026-03-24
Add Standalone Activities Interactive Demo page [experimentation]

jsundai:
## What does this PR do?
preview https://temporal-documentation-git-feat-standalone-activities-in-03cd2c.preview.thundergun.io/develop/standalone-activities-interactive-demo 

## Notes to reviewers
TBD: Going to do a final self review before merging and probably will make some edits on wording. 

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213797817618959">EDU-6095 Add Standalone Activities Interactive Demo page [experimentation]</a>


docs/develop/standalone-activities-interactive-demo.mdx
```diff
@@ -0,0 +1,83 @@
...
+in the Go, Python, and .NET SDKs, and as a
+[Pre-release](/evaluate/development-production-features/release-stages#pre-release) feature in the
+Java and TypeScript SDKs.
+
+Standalone Activities require Temporal CLI v1.7.0 or higher and Temporal Server v1.31.0 or higher.
+
+:::
+
+Standalone Activities let you run a single Activity straight from your application without
+writing a Workflow. Your code calls the Temporal Client, the Server durably enqueues the request
```
dandavison:
```suggestion
writing a Workflow. Your code uses the Temporal Client to send the request to the Server, the Server durably enqueues the request
```

docs/develop/standalone-activities-interactive-demo.mdx
```diff
@@ -0,0 +1,83 @@
...
+   the request survives Worker restarts and network interruptions.
+3. **Poll**: A Worker that is polling that Task Queue picks up the Activity Task and prepares to
+   execute it.
+4. **Execute**: The Worker runs your Activity function with the provided arguments and reports the
+   outcome back to the Server.
+5. **Return**: The Server stores the result and returns it to the original caller through the
+   `ActivityHandle` that was created when the Activity was scheduled.
+
+Because the Server durably persists the Activity, it survives Worker restarts and network
+interruptions. If the Activity fails, the Server automatically retries it according to the Retry
```
dandavison: > "it survives Worker restarts and network interruptions"

I feel like we use similar language for Workflows. Is it worth adding something like (I'm not saying this is the right phrasing)

> However, unlike a Workflow, an Activity is by default re-executed from the beginning on each attempt, so you should typically do one operation with side effects in an Activity. If you need to do two different operations (both with side effects), you should typically use a Workflow with two Activities.

src/components/elements/StandaloneActivityDemo.js
```diff
@@ -0,0 +1,624 @@
...
+  }, [config]);
+
+  const codeLanguage = language === 'dotnet' ? 'csharp' : language;
+
+  const failureNote =
+    config.simulateFailures
+      ? config.failCount > config.maxRetries
+        ? `All ${config.maxRetries + 1} attempt(s) will fail because failCount exceeds maxRetries.`
+        : `Activity will fail ${config.failCount} time(s), then succeed on attempt ${config.failCount + 1}.`
+      : null;
```
dandavison, line 358: It would be nice to style this as an error because it's saying that the user input doesn't make sense.

Image

src/components/elements/StandaloneActivityDemo.js
```diff
@@ -0,0 +1,624 @@
...
+        ['completed', 'active', 'pending', 'pending', 'pending'],
+        `Scheduling activity "${config.activityId}" on task queue "${config.taskQueue}"...`
+      );
+      await sleep(600);
+      if (isCancelled()) return;
+
+      update(['completed', 'completed', 'pending', 'pending', 'pending'], null);
+
+      // ── Retry loop ───────────────────────────────────────────────────────
+      while (true) {
```
dandavison, line 211: Does setting StartToClose timeout actually do anything? Here it succeeds even though StartToClose was zero.

Related, I thint here should be a way to set the duration that each attempt takes, so that user scan understand that a StartToClose timeout failure occurs when the duration of an individual attempt exceeds the specified timeout.

And related, shall we add ScheduleToStart timeout also, so that users can understand that it occurs when something spends too long in the queue (typically due to insufficient worker resources).

Image

docs/develop/standalone-activities-interactive-demo.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+import { StandaloneActivityDemo } from '@site/src/components';
+
+<StandaloneActivityDemo />
+
+---
+
+## How it works
+
+When you call `client.ExecuteActivity()` (or the equivalent in your SDK), the following happens:
```
dandavison: The code written there looks like Go SDK code. Unfortunately, the Go SDK is different from all other SDKs: "ExecuteFoo" in Go means what "StartFoo" means in other SDKs. So, `ExecuteActivity()` in Go does _not_ poll. A simple fix would be to use Python code: `client.execute_activity()`.


To see what I mean, see the Go code from the docs here:

```go
	handle, err := c.ExecuteActivity(context.Background(), activityOptions, helloworld.Activity, "Temporal") // Just sends it to the server
    ...
	err = handle.Get(context.Background(), &result) // this does the polling for result in Go!
```

docs/develop/standalone-activities-interactive-demo.mdx
```diff
@@ -0,0 +1,83 @@
...
+Policy you configure.
+
+### Standalone vs Workflow Activities
+
+| | Workflow Activity | Standalone Activity |
+|---|---|---|
+| Orchestrated by | A Workflow Definition | Your application code (via the Temporal Client) |
+| Started with | SDK-specific (for example, `workflow.ExecuteActivity()` in Go) | SDK-specific (for example, `client.ExecuteActivity()` in Go, `client.StartActivityAsync()` in .NET) |
+| Retry policy | Set in `ActivityOptions` inside your Workflow | Set in `StartActivityOptions` when calling the client |
+| Visibility | Shown in the Workflow's Event History | Shown in the Activity list and count views |
```
dandavison: If this is referring to what they see in the UI then

```suggestion
| Visibility | Shown in the Workflow's Event History | Shown in the Standalone Activity list and count views |
```

src/components/elements/StandaloneActivityDemo.js
```diff
@@ -0,0 +1,624 @@
...
+    activityType = 'Activity';
+    inputFlag = `--input '"${name}"'`;
+  } else if (language === 'python') {
+    activityType = 'compose_greeting';
+    inputFlag = `--input '{"greeting": "${GREETING}", "name": "${name}"}'`;
+  } else {
+    activityType = 'ComposeGreeting';
+    inputFlag = `--input '{"Greeting": "${GREETING}", "Name": "${name}"}'`;
+  }
+
```
dandavison: It's bothering me a bit that in one place the input is a class instance containing two fields, and in the CLI the user sees the JSON representation of that, but in the input form field in the simulation then just enter the "name" bit, and in the log it suggests that the activity only takes one argument. Maybe the whole thing would be simpler   if it just has a single string argument and we don't try to teach them about using classes/structs as input in this demo?

Image

lennessyy (approved):
Looks great! Just make sure to adjust some text contrasts in both dark and light mode

Image

PR #4375, opened 2026-03-31
Interactive Workflow Message Passing demo [experimentation] 

jsundai:
## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213891462636013">EDU-6144 interactive Workflow Message Passing demo [experimentation]</a>


jsundai:
The Build It Section is probably the most important part of this experimentation to come back to that can be used as a reusable component across the docs. 

The core idea in one sentence is also a good idea to use across the docs. The resources card section could also be pulled into a reusable component to be used across the docs.

PR #4379, opened 2026-04-02
docs: update blob size limit error troubleshooting page

lennessyy:
## Summary
- Updates the blob size limit error troubleshooting page to recommend External Storage as the primary solution for large payloads
- Reorders resolution strategies to put External Storage first, followed by compression and batching
- Adds link to the Python SDK large payload storage guide

Split from #4333 (PR 3 of 3).

## Test plan
- [ ] Verify the blob-size-limit-error.mdx page renders correctly
- [ ] Confirm links to External Storage and Python SDK pages resolve properly

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213918690039389">EDU-6149 docs: update blob size limit error troubleshooting page</a>


docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -21,39 +24,48 @@ The `BlobSizeLimitError` is an error that occurs when the size of a blob (payloa
...
 
 ## How do I resolve this error?
 
-To resolve this error, reduce the size of the blob so that it is within the 4 MB limit.
+To resolve this error, reduce the size of the blob so that it is within the limits.
 
 There are multiple strategies you can use to avoid this error:
 
-1. Use compression with a [custom payload codec](/payload-codec) for large payloads.
+1. (Recommended) Use [External Storage](/external-storage) to offload large payloads to an object store like S3.
```
drewhoskins-temporal: "Recommended" is a bit strong given that it's in pre-release. But if you add a pre-release note, that could address gaps in terms of language support and stability

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -2,7 +2,9 @@
 id: blob-size-limit-error
 title: Troubleshoot the blob size limit error
 sidebar_label: Blob size limit error
-description: The BlobSizeLimitError occurs when a Workflow's payload exceeds the 2 MB request limit or the 4 MB Event History transaction limit set by Temporal. Reduce blob size via compression or batching.
+description:
```
drewhoskins-temporal, line 5: Hmm, I know you aren't adding this, so no need to act now, but I wonder if anything about this is still accurate. If it's referring to server side errors, I think we throw GrpcMessageTooLarge in the latter case (but we can check w/ eng to be sure.) and BadAttributes in the former case. At least that's what I discussed with @simvlad and @jmaeagle99 but maybe I'm confused about what this means.

Update: according to https://github.com/search?q=org%3Atemporalio%20BlobSizeLimitError&type=code, I think "BlobSizeLimitError" is the actual size limit rather than the error that's thrown...

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -21,39 +24,48 @@ The `BlobSizeLimitError` is an error that occurs when the size of a blob (payloa
...
+   a storage driver uploads it to your external store and replaces it with a small reference token in the Event History.
+   Your Workflow and Activity code doesn't need to change. Even if your payloads are within the limit today, consider
+   implementing External Storage if their size could grow over time.
 
-   - This addresses the immediate issue of the blob size limit; however, if blob sizes continue to grow this problem can arise again.
+   For SDK-specific guides, see:
+   - [Python: Large payload storage](/develop/python/data-handling/large-payload-storage)
 
-2. Break larger batches of commands into smaller batch sizes:
+2. Use compression with a [custom Payload Codec](/payload-codec) for large payloads. This addresses the immediate issue,
```
drewhoskins-temporal:
```suggestion
2. Use compression with a [custom Payload Codec](/payload-codec) for large payloads.  But even if this addresses the immediate issue,
```

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -21,39 +24,48 @@ The `BlobSizeLimitError` is an error that occurs when the size of a blob (payloa
...
-To resolve this error, reduce the size of the blob so that it is within the 4 MB limit.
+To resolve this error, reduce the size of the blob so that it is within the limits.
 
 There are multiple strategies you can use to avoid this error:
 
-1. Use compression with a [custom payload codec](/payload-codec) for large payloads.
+1. (Recommended) Use [External Storage](/external-storage) to offload large payloads to an object store like S3.
+   Currently available in the [Python SDK](/develop/python/data-handling/large-payload-storage). When a payload exceeds a size threshold,
+   a storage driver uploads it to your external store and replaces it with a small reference token in the Event History.
+   Your Workflow and Activity code doesn't need to change. Even if your payloads are within the limit today, consider
```
drewhoskins-temporal: nit: "Even if your payloads are within the limit today" -- the user probably wouldn't be in this doc if that were the case. =)  But I like the advice in general.

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -21,39 +24,48 @@ The `BlobSizeLimitError` is an error that occurs when the size of a blob (payloa
...
+   implementing External Storage if their size could grow over time.
 
-   - This addresses the immediate issue of the blob size limit; however, if blob sizes continue to grow this problem can arise again.
+   For SDK-specific guides, see:
```
drewhoskins-temporal: I think it's too early to wipe the old advice.  (a) cos pre-release/limited language support, (b) since we don't provide the ability to create payload handles yet, the advice to "Pass references to the stored payloads within the Workflow instead of the actual data." is still valid no matter what.

Maybe we just list these as alternatives?

docs/troubleshooting/blob-size-limit-error.mdx
simvlad: In case if it exceeds 2MB, we terminated the workflow with one of bad** errors, like the following:

`BadScheduleActivityAttributes: ScheduleActivityTaskCommandAttributes.Input exceeds size limit.`

but this is specific to the large payloads in the activity input. For example, we also throw WORKFLOW_TASK_FAILED_CAUSE_BAD_UPDATE_WORKFLOW_EXECUTION_MESSAGE.

As for the GrpcMessageTooLarge, it would be on the client and look like this:

```
rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5767664 vs. 4194304
```

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -21,39 +24,48 @@ The `BlobSizeLimitError` is an error that occurs when the size of a blob (payloa
...
    - Workflow Task-level batching:
      1. Execute Activities in smaller batches within a single Workflow Task.
-     2. Introduce brief pauses or sleeps (for example, 1ms) between batches.
-
-3. Consider offloading large payloads to an object store to reduce the risk of exceeding blob size limits:
-   1. Pass references to the stored payloads within the Workflow instead of the actual data.
-   2. Retrieve the payloads from the object store when needed during execution.
+     2. Introduce brief pauses or sleeps between batches.
 
 ## Workflow termination due to oversized response
```
simvlad: Actually, I tried it again. In the case workflow task response exceeds 4MB gRPC limit, the workflow would keep retrying with:
```
{
  "cause": "WORKFLOW_TASK_FAILED_CAUSE_GRPC_MESSAGE_TOO_LARGE",
  "failure": {
    "message": "rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5243837 vs. 4194304)",
    "applicationFailureInfo": {
      "type": "GrpcMessageTooLargeError"
    }
  }
}
```

So this is not correct

docs/troubleshooting/blob-size-limit-error.mdx
lennessyy: I was thinking they may have multiple Workflows. They ran into this error on one of them, but this could nudge them to refactor others even if those aren't reporting errors now.

docs/troubleshooting/blob-size-limit-error.mdx
lennessyy: I'll keep this as the first item but remove (Recommended) for now. We can add it back either during Public Preview or GA

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -12,48 +12,69 @@ tags:
...
 
-To prevent this, use the batching strategies described above to split work across multiple Workflow Tasks instead of scheduling everything at once.
+A Workflow can hit this limit even when every individual payload is under 2 MB. Scheduling several Activities with moderate-sized inputs, or hundreds of Activities with tiny inputs in the same Workflow Task can push the combined request past 4 MB. Activity results are also subject to this limit.
 
-See the [gRPC Message Too Large error reference](/references/errors#grpc-message-too-large) for more details.
+### Error messages
+
+- `WORKFLOW_TASK_FAILED_CAUSE_GRPC_MESSAGE_TOO_LARGE`: When a Workflow Worker completes a Workflow Task, it sends all the commands the Workflow produced (such as Activity schedules and their inputs) back to the Temporal Service. If that response exceeds 4 MB, the SDK catches the gRPC error and sends a failed Workflow Task response with this cause. Because replay produces the same oversized response, the Workflow gets stuck in a retry loop that isn't visible in the Event History.
```
jmaeagle99: There is a new error message that is analogous to this one called `WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE`.

The SDK will get the payload limits from the server and intentionally fail workflow tasks if it contains payloads that are over the payload size limit. This enables the workflow to be retried instead of failing it so that other solutions (such as external storage) may be applied to alleviate the problem and allow the workflow to continue.

lennessyy: Huh, I'm a little confused by this. What makes this error different from the other payload errors (in the above section)? Why does this allow retries while other payload errors terminate the workflow per Vlad's testing?

Or do you mean this is a new error message we are adding together with the release of external storage?

jmaeagle99: It's a new error code. Prior to adding this code and the new behavior to the SDK, if the SDK submitted payloads that were greater than 2 MB but less than 4 MB (the gRPC limit), the server would fail the workflow (not just the workflow task, but the entire workflow). The SDK does the best effort size check on the worker and intentionally fails the workflow task instead of uploading the large payload that would fail the workflow.

docs/troubleshooting/blob-size-limit-error.mdx
drewhoskins-temporal: One note is that they will get TMPRL1103 warnings in this case.

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -12,48 +14,124 @@ tags:
...
+### How to resolve
+
+1. Offload large payloads to an object store to reduce the risk of exceeding payload size limits:
    1. Pass references to the stored payloads within the Workflow instead of the actual data.
-   2. Retrieve the payloads from the object store when needed during execution.
+   1. Retrieve the payloads from the object store when needed during execution.
+
+   This is called the
+   [claim check pattern](https://dataengineering.wiki/Concepts/Software+Engineering/Claim+Check+Pattern). You can
+   implement your own claim check pattern by using a custom [Payload Codec](/payload-codec), or use
```
jmaeagle99, line 65: I would suggest separating these two solutions. As it is, it feels like for both of them, a customer can implement their own claim check pattern. Maybe something like: "The claim check pattern is built into the SDKs as [External Storage](/external-storage), or you can implement your own claim check pattern by using a custom [Payload Codec](/paload-codec)".

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -1,8 +1,10 @@
 ---
 id: blob-size-limit-error
-title: Troubleshoot the blob size limit error
-sidebar_label: Blob size limit error
-description: The BlobSizeLimitError occurs when a Workflow's payload exceeds the 2 MB request limit or the 4 MB Event History transaction limit set by Temporal. Reduce blob size via compression or batching.
+title: Troubleshoot payload and gRPC message size limit errors
+sidebar_label: Payload and gRPC message size limit errors
```
jmaeagle99, line 4:
```suggestion
sidebar_label: Message size limit errors
```
The text is somewhat long in the size bar such that it wraps. I think you can drop "Payload and gRPC" and it still has connotes the same effect.

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -12,48 +14,124 @@ tags:
...
+request past 4 MB. Activity results are also subject to this limit.
 
-This typically happens when a Workflow schedules too many Activities, Child Workflows, or other commands in a single Workflow Task. The total size of all commands generated by the Workflow Task must fit within the 4 MB limit.
+### Error messages
 
-If your Workflow was terminated for this reason, you'll see a `WorkflowExecutionTerminated` event in the Event History with the cause `WORKFLOW_TASK_FAILED_CAUSE_GRPC_MESSAGE_TOO_LARGE`.
+The error message depends on which operation carried the oversized gRPC message and which SDK version is in use.
 
-To prevent this, use the batching strategies described above to split work across multiple Workflow Tasks instead of scheduling everything at once.
+- `WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE`
```
jmaeagle99, line 99: WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE or TMPRL1103 do not indicate a gRPC limit error, only a payload limit error.

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -12,48 +14,124 @@ tags:
...
+- `[TMPRL1103] Attempted to upload payloads with size that exceeded the error limit.`
+- `WORKFLOW_TASK_FAILED_CAUSE_GRPC_MESSAGE_TOO_LARGE`
+- `ScheduleToCloseTimeout` (Activities only, see [error behavior](#grpc-error-behavior) below)
 
-See the [gRPC Message Too Large error reference](/references/errors#grpc-message-too-large) for more details.
+### Error behavior {#grpc-error-behavior}
+
+The behavior when a gRPC message exceeds the size limit depends on the SDK version.
+
+**Python SDK 1.23.0+:** The SDK fails the Workflow Task with cause `WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE`. The
```
jmaeagle99, line 108: My understanding is that only Go and Java are emitting WORKFLOW_TASK_FAILED_CAUSE_GRPC_MESSAGE_TOO_LARGE at this time. The other SDKs are failing the workflow task but not specifying this exact cause. I can try to locate exact versions for Go and Java for you.

jmaeagle99: I've opened https://github.com/temporalio/sdk-core/issues/1228 to track fixing it for the other SDKs.

brianmacdonald-temporal (approved):
Looks good to me.

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -12,48 +14,122 @@ tags:
...
+
+1. Offload large payloads to an object store to reduce the risk of exceeding payload size limits:
    1. Pass references to the stored payloads within the Workflow instead of the actual data.
-   2. Retrieve the payloads from the object store when needed during execution.
+   1. Retrieve the payloads from the object store when needed during execution.
+
+   This is called the
+   [claim check pattern](https://dataengineering.wiki/Concepts/Software+Engineering/Claim+Check+Pattern). The claim
+   check pattern is built into the SDKs as [External Storage](/external-storage), or you can implement your own claim
+   check pattern by using a custom [Payload Codec](/paload-codec)
```
lennessyy:
```suggestion
   check pattern by using a custom [Payload Codec](/payload-codec)
```

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -12,48 +14,122 @@ tags:
...
+- `WORKFLOW_TASK_FAILED_CAUSE_BAD_UPDATE_WORKFLOW_EXECUTION_MESSAGE`
 
-There are multiple strategies you can use to avoid this error:
+### Error behavior {#payload-error-behavior}
 
-1. Use compression with a [custom payload codec](/payload-codec) for large payloads.
+The behavior when a payload exceeds the size limit depends on the SDK version.
 
-   - This addresses the immediate issue of the blob size limit; however, if blob sizes continue to grow this problem can arise again.
+**Python SDK 1.23.0+:** The SDK fails the Workflow Task with cause `WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE`. The
```
drewhoskins-temporal, line 46: This sentence suddenly assumes that we're talking about workflow tasks, but it seems like this document is trying to be more general and covering other cases such as when clients are uploading?  My suggestion is to make this document cover both client and WFT errors and differentiate those cases.  Second choice is to scope the whole thing to WFTs and circle back with docs for clients.

docs/troubleshooting/blob-size-limit-error.mdx
```diff
@@ -12,48 +14,122 @@ tags:
...
    1. Pass references to the stored payloads within the Workflow instead of the actual data.
-   2. Retrieve the payloads from the object store when needed during execution.
+   1. Retrieve the payloads from the object store when needed during execution.
+
+   This is called the
+   [claim check pattern](https://dataengineering.wiki/Concepts/Software+Engineering/Claim+Check+Pattern). The claim
+   check pattern is built into the SDKs as [External Storage](/external-storage), or you can implement your own claim
+   check pattern by using a custom [Payload Codec](/payload-codec)
+
+   This is the most reliable way to avoid hitting payload size limits. Consider implementing the claim check pattern for
```
drewhoskins-temporal, line 68: nit: Claim check isn't really implemented for specific Ws and As.  It's installed in the temporal client and used for any Ws/As that the client touches.  So maybe "Consider adding External Storage to your clients when you have Ws and As..."

drewhoskins-temporal (commented):
big improvement, thanks!  Sorry for the delay in review.

PR #4392, opened 2026-04-06
Document known accuracy limitations for ApproximateBacklogCount

dustin-temporal:
## Summary
- Adds a "Known accuracy limitations" section to the worker performance page explaining two scenarios where `ApproximateBacklogCount` can be temporarily inaccurate:
  - **Overcount from cancelled Workflows** - pending Tasks from cancelled Workflows inflate the count until they expire
  - **Reset to zero on partition unload** - idle Task Queues (no pollers, no new Tasks for ~5 min) lose their count when the partition is unloaded
- Consolidates the existing sticky queue caveat into the same section
- Adds a cross-reference note to the Cloud metrics reference (`temporal_cloud_v1_approximate_backlog_count`)

## Context
These limitations were identified during internal discussion with the matching team (Shahab, Nish). Both behaviors are explainable and expected, but not currently documented - leading to customer confusion.

## Test plan
- [ ] Verify the anchor link `#backlog-accuracy-limitations` works from the Cloud metrics reference page
- [ ] Review rendering of the `:::note` admonition in the metrics reference

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213963890395073">EDU-6164 Document known accuracy limitations for ApproximateBacklogCount</a>


docs/cloud/metrics/openmetrics/metrics-reference.mdx
```diff
@@ -508,6 +508,11 @@ These metrics could have high cardinality depending on number of task queues pre
 
 The approximate number of tasks pending in a task queue. Started Activities are not included in the count as they have been dequeued from the task queue.
 
+:::note Known accuracy limitations
+This metric may temporarily overcount due to cancelled Workflow Tasks that haven't yet expired, and may reset to zero if no Workers poll and no Tasks are added for approximately 5 minutes (due to partition unload).
```
ShahabT: Nits:
- it's not only canceled/terminated workflows, it's also expired tasks, say for expired workflows (specially in new matcher).
- instead of "partition unload" we can just say "task queue unload" because user should not be exposed to partition concept.

docs/develop/worker-performance.mdx
```diff
@@ -720,9 +720,18 @@ The age is based on the creation time of the Task at the head of the queue.
...
 
-Please note: [Sticky queues](https://docs.temporal.io/sticky-execution) will affect these values, but only for a few seconds.
-That's because Tasks sent to Sticky queues are not included in the returned values for `ApproximateBacklogCount` and `ApproximateBacklogAge`.
-Inaccuracies diminish as the backlog grows.
+#### Known accuracy limitations {#backlog-accuracy-limitations}
+
+These values are approximate and may be temporarily inaccurate in the following scenarios:
+
+- **Overcount from cancelled Workflows**: When a Workflow is cancelled, its pending Tasks may still be counted in the backlog until they expire.
+  Workers process Tasks in order within a partition, so valid Tasks _can_ be blocked behind invalid (expired) Tasks, but invalid Tasks cannot block other invalid Tasks.
```
ShahabT: - again, the partition concept doesn't have much bearing here, so maybe we can omit it.
- "valid Tasks _can_ be blocked behind invalid (expired) Tasks" -> this is the other way around: valid tasks, until dispatched, can block invalid tasks. once an invalid task is in the front of the queue, we remove it quickly so they do not generally block other invalid/valid tasks. A valid task may stay longer in front of the queue if there are not enough workers.

docs/develop/worker-performance.mdx
```diff
@@ -720,9 +720,18 @@ The age is based on the creation time of the Task at the head of the queue.
...
+  The count eventually converges to the correct value as expired Tasks are cleared.
+- **Reset to zero on partition unload**: If no Workers are polling a Task Queue and no new Tasks are added for approximately 5 minutes, the Temporal Service unloads the Task Queue partition from memory.
```
ShahabT: > no Workers are polling a Task Queue and no new Tasks are added

Other task queue activities such as DescribeTaskQueue or UpdateTaskQueueConfig will also keep the task queue loaded.

docs/cloud/metrics/openmetrics/metrics-reference.mdx
dnr: I'm not sure what is meant by "... tasks that haven't yet expired". The idea is that when they expire or become invalid, they are still counted because they haven't been _processed_ and discarded yet

docs/develop/worker-performance.mdx
```diff
@@ -720,9 +720,18 @@ The age is based on the creation time of the Task at the head of the queue.
...
 You can rely on both these counts when making scaling decisions.
 
-Please note: [Sticky queues](https://docs.temporal.io/sticky-execution) will affect these values, but only for a few seconds.
-That's because Tasks sent to Sticky queues are not included in the returned values for `ApproximateBacklogCount` and `ApproximateBacklogAge`.
-Inaccuracies diminish as the backlog grows.
+#### Known accuracy limitations {#backlog-accuracy-limitations}
+
+These values are approximate and may be temporarily inaccurate in the following scenarios:
+
+- **Overcount from cancelled Workflows**: When a Workflow is cancelled, its pending Tasks may still be counted in the backlog until they expire.
```
dnr: same, I think "expire" is the wrong word here

docs/develop/worker-performance.mdx
dnr: also: "invalid" (workflow is closed, or activity is canceled, or various other things) and "expired" (due to timeout) are technically different internally but I think it's good to just bundle them together for this purpose. We should be consistent, though, maybe use "invalid" everywhere, or "invalid or expired"?

docs/develop/worker-performance.mdx
```diff
@@ -720,9 +720,18 @@ The age is based on the creation time of the Task at the head of the queue.
 
 You can rely on both these counts when making scaling decisions.
 
-Please note: [Sticky queues](https://docs.temporal.io/sticky-execution) will affect these values, but only for a few seconds.
-That's because Tasks sent to Sticky queues are not included in the returned values for `ApproximateBacklogCount` and `ApproximateBacklogAge`.
-Inaccuracies diminish as the backlog grows.
+#### Known accuracy limitations {#backlog-accuracy-limitations}
+
+These values are approximate and may be temporarily inaccurate in the following scenarios:
```
dnr: There's also just all the discrepancy due to metadata updates being infrequent, and discrepancy due to database TTLs (cassandra makes rows disappear and we don't see them). I think "may be temporarily inaccurate" is too strong, it suggests inaccuracy is temporary and will go away, but that's not true at all.

docs/develop/worker-performance.mdx
```diff
@@ -720,9 +720,18 @@ The age is based on the creation time of the Task at the head of the queue.
...
+
+- **Overcount from cancelled Workflows**: When a Workflow is cancelled, its pending Tasks may still be counted in the backlog until they expire.
+  Workers process Tasks in order within a partition, so valid Tasks _can_ be blocked behind invalid (expired) Tasks, but invalid Tasks cannot block other invalid Tasks.
+  The count eventually converges to the correct value as expired Tasks are cleared.
```
dnr: This feels like it's promising too much. Invalid tasks will eventually be accounted for, but I don't think we can say the count will converge to the correct value, there are other sources of discrepancy.

docs/develop/worker-performance.mdx
```diff
@@ -720,9 +720,18 @@ The age is based on the creation time of the Task at the head of the queue.
...
+- **Reset to zero on partition unload**: If no Workers are polling a Task Queue and no new Tasks are added for approximately 5 minutes, the Temporal Service unloads the Task Queue partition from memory.
+  When this happens, `ApproximateBacklogCount` resets to zero until the partition is reloaded (by a Worker polling or a new Task being added).
```
carlydf, line 731: should we say that calling `DescribeTaskQueue` also reloads the task queue?

docs/cloud/metrics/openmetrics/metrics-reference.mdx
dustin-temporal: This has been reworded to address the nits and also clarified the language here that we're counting everything that hasn't been processed: valid, expired, or invalid alike. Needs to be processed in before it is removed from the count.

docs/develop/worker-performance.mdx
dustin-temporal: Clarified here as well

docs/develop/worker-performance.mdx
dustin-temporal: Feels a little weird to talk about the database TTLs but I took a shot at noting there are some inaccuracies that just exist and alluded to their source.

docs/develop/worker-performance.mdx
dustin-temporal: Broke this out to talk about transient versus potentially persistent discrepancy sources.

docs/develop/worker-performance.mdx
dustin-temporal: added this scenario for reload

docs/develop/worker-performance.mdx
dustin-temporal: dropped partition mentions and fixed my backward logic!

dustin-temporal:
@dnr @ShahabT @carlydf at your leisure, please take another look. I believe I've now addressed all of the comments. Thanks for the help!

docs/cloud/metrics/openmetrics/metrics-reference.mdx
```diff
@@ -518,6 +518,14 @@ These metrics could have high cardinality depending on number of task queues pre
 
 The approximate number of tasks pending in a task queue. Started Activities are not included in the count as they have been dequeued from the task queue.
 
+:::note Known accuracy limitations
+This metric is approximate.
+It can overcount because invalid or expired Tasks (for example, from cancelled, terminated, completed, or timed out Workflows) remain in the count until they reach the head of the queue and are processed and discarded.
```
flippedcoder:
```suggestion
It can overcount because invalid or expired Tasks, like from cancelled, terminated, completed, or timed out Workflows, remain in the count until they reach the head of the queue and are processed and discarded.
```

docs/cloud/metrics/openmetrics/metrics-reference.mdx
```diff
@@ -518,6 +518,14 @@ These metrics could have high cardinality depending on number of task queues pre
...
+It can also reset to zero on an idle Task Queue. If no Worker polls, no new Tasks are added, and no other Task Queue calls occur (such as `DescribeTaskQueue` or `UpdateTaskQueueConfig`) for approximately 5 minutes, the Task Queue is unloaded from memory.
```
flippedcoder:
```suggestion

It can also reset to zero on an idle Task Queue. If no Worker polls, no new Tasks are added, and no other Task Queue calls occur (such as `DescribeTaskQueue` or `UpdateTaskQueueConfig`) for approximately 5 minutes. The Task Queue is unloaded from memory.
```

docs/develop/worker-performance.mdx
```diff
@@ -729,9 +729,20 @@ The age is based on the creation time of the Task at the head of the queue.
...
 
-Please note: [Sticky queues](https://docs.temporal.io/sticky-execution) will affect these values, but only for a few seconds.
-That's because Tasks sent to Sticky queues are not included in the returned values for `ApproximateBacklogCount` and `ApproximateBacklogAge`.
-Inaccuracies diminish as the backlog grows.
+#### Known accuracy limitations {#backlog-accuracy-limitations}
+
+These values are approximate.
+The most common sources:
+
+- **Overcount from invalid or expired Tasks**: Tasks belonging to cancelled, terminated, completed, or timed out Workflows (and Activities) stay in the count until they reach the head of the queue and are processed and discarded.
```
flippedcoder:
```suggestion
- **Overcount from invalid or expired Tasks**: Tasks belonging to cancelled, terminated, completed, or timed out Workflows and Activities stay in the count until they reach the head of the queue and are processed and discarded.
```

docs/develop/worker-performance.mdx
```diff
@@ -729,9 +729,20 @@ The age is based on the creation time of the Task at the head of the queue.
...
+  An invalid or expired Task at the head is removed quickly, so it rarely holds up the count for long.
+  A valid Task at the head can stay there longer when there aren't enough Workers to dispatch it; while it sits there, invalid or expired Tasks queued behind it cannot be removed.
```
flippedcoder:
```suggestion
  A valid Task at the head can stay there longer when there aren't enough Workers to dispatch it. While it sits there, invalid or expired Tasks queued behind it cannot be removed.
```

docs/develop/worker-performance.mdx
```diff
@@ -729,9 +729,20 @@ The age is based on the creation time of the Task at the head of the queue.
...
+  Invalid and expired Tasks are eventually accounted for, but the count may not return to a fully accurate value because of other sources of discrepancy (such as infrequent metadata updates and database row expirations).
```
flippedcoder:
```suggestion
  Invalid and expired Tasks are eventually accounted for, but the count may not return to a fully accurate value because of other sources of discrepancy like infrequent metadata updates and database row expirations.
```

flippedcoder (approved):

PR #4448, opened 2026-04-17
Update docs to refine explanations for non-determinism

flippedcoder:
## What does this PR do?

Some pages in the docs need to be updated so that determinism is better understood. Right now we say that Temporal is deterministic when really only Workflows are deterministic and Activities handle non-determinism operations. 

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1214124203356022">EDU-6219 Update docs to refine explanations for non-determinism</a>

### DIFF
--- a/docs/develop/dotnet/workflows/basics.mdx
+++ b/docs/develop/dotnet/workflows/basics.mdx
@@ -64,7 +64,7 @@ Temporal requires that the deterministic `TaskScheduler.Current` is used, but ma
 Here are some known gotchas to avoid with .NET tasks inside of Workflows:
 
 -  Use `Workflow.RunTaskAsync` instead of `Task.Run`. `Task.Run` uses the default scheduler and puts work on the thread pool.
-    - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
+    - You can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
 - If you need to use `Task.ConfigureAwait`, use `Task.ConfigureAwait(true)`. `Task.ConfigureAwait(false)` won't use the current context.
   - There is no significant performance benefit to `Task.ConfigureAwait` in workflows because of how the scheduler works.
 - Avoid anything that defaults to the default task scheduler.
--- a/docs/develop/dotnet/workflows/basics.mdx
+++ b/docs/develop/dotnet/workflows/basics.mdx
@@ -66,7 +66,7 @@ Here are some known gotchas to avoid with .NET tasks inside of Workflows:
 -  Use `Workflow.RunTaskAsync` instead of `Task.Run`. `Task.Run` uses the default scheduler and puts work on the thread pool.
     - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
 - If you need to use `Task.ConfigureAwait`, use `Task.ConfigureAwait(true)`. `Task.ConfigureAwait(false)` won't use the current context.
-  - There is no significant performance benefit to `Task.ConfigureAwait` in workflows anyways due to how the scheduler works.
+  - There is no significant performance benefit to `Task.ConfigureAwait` in workflows because of how the scheduler works.
 - Avoid anything that defaults to the default task scheduler.
 - Use `Workflow.DelayAsync`, `Workflow.WaitConditionAsync`, or non-timeout-based cancellation token sources instead of `Task.Delay`, `Task.Wait`, timeout-based `CancellationTokenSource`, or anything that uses .NET built-in timers.
 - Use `Workflow.WhenAnyAsync` instead of `Task.WhenAny`.
--- a/docs/develop/dotnet/workflows/basics.mdx
+++ b/docs/develop/dotnet/workflows/basics.mdx
@@ -44,11 +44,9 @@ All Workflow Definition parameters must be serializable.
 
 ## Workflow logic requirements {#workflow-logic-requirements}
 
-Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints).
-Therefore, each language is limited to the use of certain idiomatic techniques.
-However, each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with external (to the Workflow) application code.
+Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints). Each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with application code outside the Workflow.
 
-This means there are several things Workflows cannot do such as:
+This means there are several things Workflows shouldn't do such as:
 
 - Perform IO (network, disk, stdio, etc)
 - Access/alter external mutable state
@@ -65,27 +63,20 @@ This is especially true with `Task`s.
 Temporal requires that the deterministic `TaskScheduler.Current` is used, but many .NET async calls will use `TaskScheduler.Default` implicitly (and some analyzers even encourage this).
 Here are some known gotchas to avoid with .NET tasks inside of Workflows:
 
-- Do not use `Task.Run` - this uses the default scheduler and puts work on the thread pool.
-  - Use `Workflow.RunTaskAsync` instead.
-  - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
-- Do not use `Task.ConfigureAwait(false)` - this will not use the current context.
-  - If you must use `Task.ConfigureAwait`, use `Task.ConfigureAwait(true)`.
+-  Use `Workflow.RunTaskAsync` instead of `Task.Run`. `Task.Run` uses the default scheduler and puts work on the thread pool.
+    - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
+- If you need to use `Task.ConfigureAwait`, use `Task.ConfigureAwait(true)`. `Task.ConfigureAwait(false)` won't use the current context.
   - There is no significant performance benefit to `Task.ConfigureAwait` in workflows anyways due to how the scheduler works.
-- Do not use anything that defaults to the default task scheduler.
-- Do not use `Task.Delay`, `Task.Wait`, timeout-based `CancellationTokenSource`, or anything that uses .NET built-in timers.
-  - `Workflow.DelayAsync`, `Workflow.WaitConditionAsync`, or non-timeout-based cancellation token source is suggested.
-- Do not use `Task.WhenAny`.
-  - Use `Workflow.WhenAnyAsync` instead.
+- Avoid anything that defaults to the default task scheduler.
+- Use `Workflow.DelayAsync`, `Workflow.WaitConditionAsync`, or non-timeout-based cancellation token sources instead of `Task.Delay`, `Task.Wait`, timeout-based `CancellationTokenSource`, or anything that uses .NET built-in timers.
+- Use `Workflow.WhenAnyAsync` instead of `Task.WhenAny`.
   - Technically this only applies to an enumerable set of tasks with results or more than 2 tasks with results. Other
     uses are safe. See [this issue](https://github.com/dotnet/runtime/issues/87481).
-- Do not use `Task.WhenAll`
-  - Use `Workflow.WhenAllAsync` instead.
+- Use `Workflow.WhenAllAsync` instead of `Task.WhenAll`.
   - Technically `Task.WhenAll` is currently deterministic in .NET and safe, but it is better to use the wrapper to be
     sure.
-- Do not use `CancellationTokenSource.CancelAsync`.
-  - Use `CancellationTokenSource.Cancel` instead.
-- Do not use `System.Threading.Semaphore` or `System.Threading.SemaphoreSlim` or `System.Threading.Mutex`.
-  - Use `Temporalio.Workflows.Semaphore` or `Temporalio.Workflows.Mutex` instead.
+- Use `CancellationTokenSource.Cancel` instead of `CancellationTokenSource.CancelAsync`.
+- Use `Temporalio.Workflows.Semaphore` or `Temporalio.Workflows.Mutex` instead of `System.Threading.Semaphore`, `System.Threading.SemaphoreSlim`, or `System.Threading.Mutex`.
   - _Technically_ `SemaphoreSlim` does work if only the async form of `WaitAsync` is used without no timeouts and
     `Release` is used. But anything else can deadlock the workflow and its use is cumbersome since it must be disposed.
 - Be wary of additional libraries' implicit use of the default scheduler.
@@ -93,6 +84,7 @@ Here are some known gotchas to avoid with .NET tasks inside of Workflows:
 
 In order to help catch wrong scheduler use, by default the Temporal .NET SDK adds an event source listener for info-level task events.
 While this technically receives events from all uses of tasks in the process, we make sure to ignore anything that is not running in a Workflow in a high performant way (basically one thread local check).
+
 For code that does run in a Workflow and accidentally starts a task in another scheduler, an `InvalidWorkflowOperationException` will be thrown which "pauses" the Workflow (fails the Workflow Task which continually retries until the code is fixed).
 This is unfortunately a runtime-only check, but can help catch mistakes early. If this needs to be turned off for any reason, set `DisableWorkflowTracingEventListener` to `true` in Worker options.
 
--- a/docs/develop/java/workflows/basics.mdx
+++ b/docs/develop/java/workflows/basics.mdx
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does 
 
 ## Workflow logic requirements {#workflow-logic-requirements}
 
-Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints).
-Therefore, each language is limited to the use of certain idiomatic techniques.
-However, each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with external (to the Workflow) application code.
+Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints). Therefore, each language is limited to the use of certain idiomatic techniques. However, each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with external (to the Workflow) application code.
 
 When defining Workflows using the Temporal Java SDK, the Workflow code must be written to execute effectively once and to completion.
 
 The following constraints apply when writing Workflow Definitions:
 
-- Do not use mutable global variables in your Workflow implementations.
-  This will ensure that multiple Workflow instances are fully isolated.
-- Your Workflow code must be deterministic.
-  Do not call non-deterministic functions (such as non-seeded random or `UUID.randomUUID()`) directly from the Workflow code.
-  The Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
-- Do not use programming language constructs that rely on system time.
-  For example, only use `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
-- Do not use native Java `Thread` or any other multi-threaded classes like `ThreadPoolExecutor`.
-  Use `Async.function` or `Async.procedure`, provided by the Temporal SDK, to execute code asynchronously.
-- Do not use synchronization, locks, or other standard Java blocking concurrency-related classes besides those provided by the Workflow class.
-  There is no need for explicit synchronization because multi-threaded code inside a Workflow is executed one thread at a time and under a global lock.
+- Do not use mutable global variables in your Workflow implementations. This will ensure that multiple Workflow instances are fully isolated.
+- Your Workflow code must be deterministic. Do not call non-deterministic functions (such as non-seeded random or `UUID.randomUUID()`) directly from the Workflow code. The Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
+  - For operations like calling external APIs, invoking LLMs, querying databases, or performing I/O, use Activities. Activities run outside Workflow replay and are retried reliably.
+- Do not use programming language constructs that rely on system time. For example, only use `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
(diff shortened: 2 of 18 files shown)


cdavisafc:
Hi @flippedcoder - thanks so much for getting this ball rolling. 

I'd like to draw your attention to a few comments in [this doc](https://docs.google.com/document/d/1KziQa9-5V0EABQ2-Qk6YNZXDH12BAizmVeEiqkL_HV4/edit?tab=t.0) - in particular, the ones around how LLMs pattern match rather than understand. I don't think it is enough to say "workflows are deterministic. activities are not"

I know that makes the task tricker - which is what I was poking at a bit with my suggestion that perhaps we take a crack at explaining replay without that word.

flippedcoder:
> Hi @flippedcoder - thanks so much for getting this ball rolling.
> 
> I'd like to draw your attention to a few comments in [this doc](https://docs.google.com/document/d/1KziQa9-5V0EABQ2-Qk6YNZXDH12BAizmVeEiqkL_HV4/edit?tab=t.0) - in particular, the ones around how LLMs pattern match rather than understand. I don't think it is enough to say "workflows are deterministic. activities are not"
> 
> I know that makes the task tricker - which is what I was poking at a bit with my suggestion that perhaps we take a crack at explaining replay without that word.

@cdavisafc Thanks for sharing this! I see what you mean about that little nuance. So I went ahead and took a stab at explaining replay [here](https://github.com/temporalio/documentation/pull/4448/changes#diff-be01e6576084754945791d3dc60cc9e7d7a3581df1c5b32da8b84d9daff2c5eeR40).

It gave me an idea of a direction to go in. Maybe instead of just saying "workflows are deterministic. activities are not", what if we said something like "Workflows are deterministic, but the Activities they execute don't need to be"?

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
 
 ## Workflow logic requirements {#workflow-logic-requirements}
 
-Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints).
-Therefore, each language is limited to the use of certain idiomatic techniques.
-However, each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with external (to the Workflow) application code.
+Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints). Therefore, each language is limited to the use of certain idiomatic techniques. However, each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with external (to the Workflow) application code.
```
Sushisource:
```suggestion
Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints). Each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with external (to the Workflow) application code.
```

Not sure this sentence was doing much for us.

docs/develop/typescript/best-practices/debugging.mdx
```diff
@@ -51,16 +51,14 @@ If something isn't behaving the way you expect, make sure to check both location
...
-You may get errors that make no sense to you because
+If those old Workflows have the same name and are on the same Task Queue, Temporal will try to continue executing them on your new code by design. You may get errors that make no sense to you because:
 
-- Temporal is trying to execute old Workflow code that no longer exists in your codebase, or
-- your new Client code is expecting Temporal to execute old Workflow/Activity code it doesn't yet know about.
+- Temporal is trying to execute old Workflow code that no longer exists in your codebase
+- Your new Client code is expecting Temporal to execute old Workflow/Activity code it doesn't yet know about.
 
-The biggest sign that this is happening is if you notice Temporal is acting non-deterministically: running the same Workflow twice gets different results.
+The biggest sign that this is happening is if you notice Workflow executions are producing different results on replay. If running the same Workflow twice gets different results, that points to a non-determinism error.
```
Sushisource: This is a bit confusing. If you're _replaying_ a workflow, it's not really even _possible_ for it to emit different results. You'd already have run into an NDE. 

However running the same workflow twice (IE: two invocations of the same workflow type) can perfectly legitimately produce different results.

docs/encyclopedia/workflow/workflow-overview.mdx
```diff
@@ -25,29 +23,59 @@ This guide provides a comprehensive overview of Temporal Workflows and covers th
...
+You'll develop those Workflows by writing code in a general-purpose programming language such as Go, Java, TypeScript, or Python. The code you write is the same code that will be executed at runtime, so you can use your favorite tools and libraries to develop Temporal Workflows.
 
-Temporal Workflows are resilient.
-They can run—and keep running—for years, even if the underlying infrastructure fails.
-If the application itself crashes, Temporal will automatically recreate its pre-failure state so it can continue right where it left off.
+Temporal Workflows are resilient. They can run—and keep running—for years, even if the underlying infrastructure fails. If the application itself crashes, Temporal will automatically recreate its pre-failure state so it can continue right where it left off.
 
-Each Workflow Execution progresses through a series of **Commands** and **Events**, which are recorded in an **Event
-History**.
+Each Workflow Execution progresses through a series of **Commands** and **Events**, which are recorded in an **Event History**.
```
Sushisource:
```suggestion
Each Workflow Execution emits a series of **Commands** and processes a sequence of **Events**, which are recorded in an **Event History**.
```

Not sure if we want to work this in (might be too much) but commands are not directly recorded in event history, however they _do_ map 1:1 to some corresponding event for that command, which is.

docs/encyclopedia/workflow/workflow-overview.mdx
```diff
@@ -25,29 +23,59 @@ This guide provides a comprehensive overview of Temporal Workflows and covers th
...
 
-Workflows must follow deterministic constraints to ensure consistent replay behavior.
+### How Workflow replay works
+
+When a Workflow pauses or encounters an error, the goal of Temporal is to bring the Workflow back to the exact same state it was in before the pause occurred. To make that possible, Temporal keeps the Event History. This is a complete, ordered log of everything that has already happened in a Workflow.
```
Sushisource: "Pauses" here I think overloads with our concept of "workflow pause", but what we really mean here is "yields". I don't know if we feel "yields" is too technical of a term but it would be more accurate than "pauses" here.

Even more pedantically, we don't replay a workflow unless it yielded _and_ was evicted from the cache of a worker, but we can probably skip over that detail.

docs/encyclopedia/workflow/workflow-overview.mdx
```diff
@@ -25,29 +23,59 @@ This guide provides a comprehensive overview of Temporal Workflows and covers th
...
+- Started Timer for 5 minutes
+- Scheduled Activity X
+- Activity X completed with result Y
+- Received Signal Z
+
+This history is the source of truth for everything that happens in the Workflow.
+
+#### Resuming a Workflow
+
+When it's time to continue the Workflow, Temporal doesn't restore memory from a snapshot. It starts the Workflow code from the beginning, replays the Event History step by step, and uses that history to guide the code back to the exact state as before. So the Workflow code is re-run, but uses the recorded events instead of redoing work.
```
Sushisource: Maybe here we could mention it doesn't always have to start from the beginning if it's cached

docs/encyclopedia/workflow/workflow-overview.mdx
```diff
@@ -25,29 +23,59 @@ This guide provides a comprehensive overview of Temporal Workflows and covers th
...
+
+Because the Workflow is re-executed to rebuild its state:
+
+- It has to make the same decisions when given the same history, which makes a Workflow deterministic.
+- It shouldn't depend on values that can change between runs.
```
Sushisource: This isn't strictly speaking accurate - it's perfectly fine to depend on values that are different between runs because they are different in the workflow input, or different results of activities. 

Most accurately, they can't depend on any values _not recorded in history_ which would be different between runs.

docs/encyclopedia/workflow/workflow-overview.mdx
```diff
@@ -25,29 +23,59 @@ This guide provides a comprehensive overview of Temporal Workflows and covers th
...
+
+For example:
+
+- A direct call to `Date.now()` could return a different value on replay.
+- A random number could change.
+- A network call could return something new.
```
Sushisource:
```suggestion
- A network call (which wasn't performed inside an activity) could return something new.
```

Sushisource (commented):
Overall I think this helps a lot. Thank you!

I find it a bit odd that we call out LLM calls separately from other API calls (they're not any meaningfully different for the purposes of these docs). I kinda understand why, maybe not everyone in the audience understands that, but it also kinda makes it seem like *we* don't understand that, heh.

docs/develop/java/workflows/basics.mdx
brianmacdonald-temporal:
```suggestion
Workflow logic is constrained by [deterministic execution requirements](/workflow-definition#deterministic-constraints). Each Temporal SDK provides a set of APIs that can be used inside your Workflow to interact with application code outside the Workflow.
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
-  The Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
-- Do not use programming language constructs that rely on system time.
-  For example, only use `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
-- Do not use native Java `Thread` or any other multi-threaded classes like `ThreadPoolExecutor`.
-  Use `Async.function` or `Async.procedure`, provided by the Temporal SDK, to execute code asynchronously.
-- Do not use synchronization, locks, or other standard Java blocking concurrency-related classes besides those provided by the Workflow class.
-  There is no need for explicit synchronization because multi-threaded code inside a Workflow is executed one thread at a time and under a global lock.
+- Do not use mutable global variables in your Workflow implementations. This will ensure that multiple Workflow instances are fully isolated.
+- Your Workflow code must be deterministic. Do not call non-deterministic functions (such as non-seeded random or `UUID.randomUUID()`) directly from the Workflow code. The Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
```
brianmacdonald-temporal:
```suggestion
-Workflow code must be deterministic. If you need to call non-deterministic functions (such as non-seeded random or `UUID.randomUUID()`) directly from the Workflow code, the Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
+  - For operations like calling external APIs, invoking LLMs, querying databases, or performing I/O, use Activities. Activities run outside Workflow replay and are retried reliably.
+- Do not use programming language constructs that rely on system time. For example, only use `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
```
brianmacdonald-temporal:
```suggestion
- Use Temporal-provided functions instead of programming language constructs that rely on system time. For example, only use `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
+- Do not use native Java `Thread` or any other multi-threaded classes like `ThreadPoolExecutor`. Use `Async.function` or `Async.procedure`, provided by the Temporal SDK, to execute code asynchronously.
```
brianmacdonald-temporal:
```suggestion
- Use `Async.function` or `Async.procedure`, provided by the Temporal SDK, to execute code asynchronously instead of native Java `Thread` or any other multi-threaded classes like `ThreadPoolExecutor`. 
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
+- Do not use synchronization, locks, or other standard Java blocking concurrency-related classes besides those provided by the Workflow class. There is no need for explicit synchronization because multi-threaded code inside a Workflow is executed one thread at a time and under a global lock.
```
brianmacdonald-temporal:
```suggestion
- Use only the concurrency features provided by the Workflow class. Multi-threaded code inside a Workflow is executed one thread at a time and under a global lock, so there is no need for explicit synchronization.
```

docs/encyclopedia/workflow/workflow-overview.mdx
brianmacdonald-temporal: I agree on the overloading of "pause" here, so we probably want to avoid that term. "Yield" might be too technical for this purpose. I think what we're trying to convey is that the Workflow execution ends unexpectedly, right? Could we say "ends" or "stops" without overloading?

docs/encyclopedia/retry-policies.mdx
```diff
@@ -3,8 +3,7 @@ id: retry-policies
 title: What is a Temporal Retry Policy?
 sidebar_label: Retry Policies
 description:
-  Optimize your Workflow and Activity Task Executions with a custom Retry Policy on Temporal. Understand default
-  retries, intervals, backoff, and maximum attempts for error handling.
+  Optimize your Workflow and Activity Task Executions with a custom Retry Policy on Temporal Understand default retries, intervals, backoff, and maximum attempts for error handling.
```
brianmacdonald-temporal: This is metadata, but I think it should still be two sentences.

brianmacdonald-temporal (commented):
Primarily did some wordsmithing -- I particularly wanted to lighten that "do not" bullet list on the workflow basics page. I was able to flip most of them from "don't do X; do Y instead" to "do Y instead of X."

Not all the SDKs have a "don't do this" list on the workflow basics page, but Python and .NET both do, and those could probably use a polish.

docs/encyclopedia/workflow/workflow-overview.mdx
flippedcoder: I'll go with "stops" here. Thanks y'all!

docs/develop/dotnet/workflows/basics.mdx
```diff
@@ -65,34 +63,28 @@ This is especially true with `Task`s.
...
 
-- Do not use `Task.Run` - this uses the default scheduler and puts work on the thread pool.
-  - Use `Workflow.RunTaskAsync` instead.
-  - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
-- Do not use `Task.ConfigureAwait(false)` - this will not use the current context.
-  - If you must use `Task.ConfigureAwait`, use `Task.ConfigureAwait(true)`.
+-  Use `Workflow.RunTaskAsync` instead of `Task.Run`. `Task.Run` uses the default scheduler and puts work on the thread pool.
+    - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
+- If you need to use `Task.ConfigureAwait`, use `Task.ConfigureAwait(true)`. `Task.ConfigureAwait(false)` won't use the current context.
   - There is no significant performance benefit to `Task.ConfigureAwait` in workflows anyways due to how the scheduler works.
```
brianmacdonald-temporal:
```suggestion
  - There is no significant performance benefit to `Task.ConfigureAwait` in workflows because of how the scheduler works.
```

docs/develop/dotnet/workflows/basics.mdx
```diff
@@ -65,34 +63,28 @@ This is especially true with `Task`s.
 Temporal requires that the deterministic `TaskScheduler.Current` is used, but many .NET async calls will use `TaskScheduler.Default` implicitly (and some analyzers even encourage this).
 Here are some known gotchas to avoid with .NET tasks inside of Workflows:
 
-- Do not use `Task.Run` - this uses the default scheduler and puts work on the thread pool.
-  - Use `Workflow.RunTaskAsync` instead.
-  - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
-- Do not use `Task.ConfigureAwait(false)` - this will not use the current context.
-  - If you must use `Task.ConfigureAwait`, use `Task.ConfigureAwait(true)`.
+-  Use `Workflow.RunTaskAsync` instead of `Task.Run`. `Task.Run` uses the default scheduler and puts work on the thread pool.
+    - Can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
```
brianmacdonald-temporal:
```suggestion
    - You can also use `Task.Factory.StartNew` with current scheduler or instantiate the `Task` and run `Task.Start` on it.
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
-  The Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
-- Do not use programming language constructs that rely on system time.
-  For example, only use `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
-- Do not use native Java `Thread` or any other multi-threaded classes like `ThreadPoolExecutor`.
-  Use `Async.function` or `Async.procedure`, provided by the Temporal SDK, to execute code asynchronously.
-- Do not use synchronization, locks, or other standard Java blocking concurrency-related classes besides those provided by the Workflow class.
-  There is no need for explicit synchronization because multi-threaded code inside a Workflow is executed one thread at a time and under a global lock.
+- Do not use mutable global variables in your Workflow implementations. This will ensure that multiple Workflow instances are fully isolated.
+- Workflow code must be deterministic. If you need to call non-deterministic functions (such as non-seeded random or `UUID.randomUUID()`) directly from the Workflow code. The Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
```
brianmacdonald-temporal:
```suggestion
- Workflow code must be deterministic. If you need to call non-deterministic functions (such as non-seeded random or `UUID.randomUUID()`) directly from the Workflow code, the Temporal SDK provides specific API for calling non-deterministic code in your Workflows.
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
+  - For operations like calling external APIs, invoking LLMs, querying databases, or performing I/O, use Activities. Activities run outside Workflow replay and are retried reliably.
+- Use Temporal-provided functions instead of that rely on system time. For example, only use `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
```
brianmacdonald-temporal:
```suggestion
- Use Temporal-provided functions instead of that rely on system time. For example, use only `Workflow.currentTimeMillis()` to get the current time inside a Workflow.
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
   - Use `Promise` and `CompletablePromise` instead of `Future` and `CompletableFuture`.
   - Use `WorkflowQueue` instead of `BlockingQueue`.
-- Use `Workflow.getVersion` when making any changes to the Workflow code.
-  Without this, any deployment of updated Workflow code might break already running Workflows.
-- Do not access configuration APIs directly from a Workflow because changes in the configuration might affect a Workflow Execution path.
-  Pass it as an argument to a Workflow function or use an Activity to load it.
-- Use `DynamicWorkflow` when you need a default Workflow that can handle all Workflow Types that are not registered with a Worker.
-  A single implementation can implement a Workflow Type which by definition is dynamically loaded from some external source.
-  All standard `WorkflowOptions` and determinism rules apply to Dynamic Workflow implementations.
+- Use `Workflow.getVersion` when making any changes to the Workflow code. Without this, any deployment of updated Workflow code might break already running Workflows.
```
brianmacdonald-temporal:
```suggestion
- Use `Workflow.getVersion` when making any changes to the Workflow code to avoid deployment of updated Workflow code interfering with already-running Workflows.
```

docs/develop/java/workflows/basics.mdx
```diff
@@ -211,34 +211,23 @@ When you set the Workflow Type this way, the value of the `name` parameter does
...
+- Do not access configuration APIs directly from a Workflow because changes in the configuration might affect a Workflow Execution path. Pass it as an argument to a Workflow function or use an Activity to load it.
```
brianmacdonald-temporal:
```suggestion
- Do not access configuration APIs directly from a Workflow because changes in the configuration might affect a Workflow Execution path. Instead, pass it as an argument to a Workflow function or use an Activity to load it.
```

docs/develop/typescript/best-practices/debugging.mdx
```diff
@@ -51,16 +51,12 @@ If something isn't behaving the way you expect, make sure to check both location
 
 If you are developing Workflows and finding that code isn't executing as expected, the first place to look is whether old Workflows are still running.
 
-If those old Workflows have the same name and are on the same task queue, Temporal will try to continue executing them on your new code by design.
-You may get errors that make no sense to you because
+If those old Workflows have the same name and are on the same Task Queue, Temporal will try to continue executing them on your new code by design. You may get errors that make no sense to you because:
 
-- Temporal is trying to execute old Workflow code that no longer exists in your codebase, or
-- your new Client code is expecting Temporal to execute old Workflow/Activity code it doesn't yet know about.
+- Temporal is trying to execute old Workflow code that no longer exists in your codebase
```
brianmacdonald-temporal:
```suggestion
- Temporal is trying to execute old Workflow code that no longer exists in your codebase.
```

brianmacdonald-temporal (approved):
Just a little copyedit polishing.

docs/encyclopedia/workflow/workflow-overview.mdx
Sushisource: Hmm, I think we want to avoid stops too. Stops implies the workflow is completed. Honestly yield is technical but really is the most accurate thing to say here. Maybe we can link from the word to a glossary entry for those who don't know it?

flippedcoder: Changed that to "yields" and linked to a wiki definition!

PR #4469, opened 2026-04-22
Fix preview links for long branch names

lennessyy:
## Summary
- When a branch name produces a subdomain over 63 characters, Vercel truncates it with a hash suffix, causing our preview link comment to point to the wrong URL.
- The workflow now checks the subdomain length first. Short names (≤63 chars) use the constructed URL directly with no added delay.
- Long names poll for the `vercel[bot]` comment (up to 3 minutes) and extract the actual preview URL from its base64 payload, falling back to the constructed URL if the comment doesn't appear.

## Test plan
- [ ] Verify short branch names still produce correct preview links with no delay
- [ ] Create a test branch with a name long enough to exceed 63 chars in the subdomain (e.g., `fix/this-is-a-very-long-branch-name-that-will-exceed-the-dns-label-limit`) and confirm the workflow extracts the correct URL from the Vercel comment

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1214162252315316">EDU-6246 Fix preview links for long branch names</a>


.github/workflows/docs-preview-links.yml
```diff
@@ -28,24 +28,63 @@ jobs:
...
-        env:
-          REPO_PREVIEW_BASE_URL: ${{ vars.DOCS_PREVIEW_BASE_URL }}
-          REPO_PREVIEW_TEMPLATE: ${{ vars.DOCS_PREVIEW_BASE_URL_TEMPLATE }}
-        run: |
-          BRANCH_NAME="${GITHUB_HEAD_REF:-${GITHUB_REF_NAME}}"
-          BRANCH_SLUG=$(echo "$BRANCH_NAME" | tr '[:upper:]' '[:lower:]' | tr -d '_' | sed 's/[^a-z0-9]/-/g' | sed 's/-\+/-/g' | sed 's/^-//' | sed 's/-$//')
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
```
flippedcoder, line 34: Just a nit, but could this script be put in a file like `preview-base-url.js`? It's usually easier to write and maintain these scripts when they're not embedded in the workflow.

Then the `package.json > "scripts"` section could be updated with something like `preview-base-url` and then called here like: `yarn run preview-base-url`.

flippedcoder (approved):

PR #4472, opened 2026-04-22
Document OpenMetrics scrape window behavior

austonli:
## Summary

- Clarifies that the OpenMetrics endpoint exposes only the most recently completed one-minute aggregation window
- Makes explicit that there is no query interface for historical data via this endpoint
- Advises users to configure their own monitoring system to retain scraped metrics

## Why

Users could reasonably assume a scrape endpoint also provides access to historical data. Without this documentation, they would discover the limitation only after building a monitoring setup that does not retain anything.

## Checklist
- [x] Follows STYLE.md guidelines
- [x] Frontmatter unchanged (edits to existing pages only)
- [x] No new pages; sidebars.js not modified
- [x] No broken links introduced


brianmacdonald-temporal (approved):
Looks good!

PR #4493, opened 2026-04-28
Document Worker Status UI in worker performance and worker health

dustin-temporal:
## Summary

Documents the Worker Status UI feature, which surfaces Worker details (task slots, CPU usage, configuration) and the Workers assigned to a Task Queue in the Temporal UI.

- **`docs/develop/worker-performance.mdx`** — adds a new section, "Visualize Workers in the UI," after the "Worker performance concepts" section. Describes how the SDK heartbeat feeds Server APIs that the UI consumes, calls out version requirements (Server 1.30+, API 1.62+), notes Cloud availability, and links to Manage Worker Heartbeating for minimum SDK versions.
- **`docs/cloud/worker-health.mdx`** — adds a brief tip near the top pointing to the new section in worker-performance.

## Why

There is no current docs coverage for visualizing Workers in the UI. This gives PMs, SAs, and customers a single place to point users when they ask how to see Workers and Workers-per-Task-Queue in the UI for troubleshooting delayed Workflow processing.

## Checklist
- [x] Follows STYLE.md (sentence-case, infinitive-form heading)
- [x] No new pages, so no sidebars.js change
- [x] Internal link uses slug-based path
- [x] Cross-links between Cloud and Develop pages

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1214383141585556">EDU-6275 Document Worker Status UI in worker performance and worker health</a>

### DIFF
--- a/docs/develop/worker-performance.mdx
+++ b/docs/develop/worker-performance.mdx
@@ -178,7 +178,7 @@ To recover from errors, Eager Workflow Start falls back to the non-eager path. F
 
 ## Visualize Workers in the UI {#visualize-workers}
 
-The Temporal SDK includes a heartbeat for Worker processes that is sent to the Server every 60 seconds, carrying information such as available task slots, CPU usage, and Worker configuration.
+The Temporal SDK includes a heartbeat for Worker processes that is sent to the Server every 60 seconds by default, carrying information such as available task slots, CPU usage, and Worker configuration.
 The Server exposes this data through APIs that surface Worker details in the Temporal UI.
 Use the UI to view the list of Workers assigned to a Task Queue and inspect Worker details to troubleshoot Workflows with delayed processing.
 
--- a/docs/develop/worker-performance.mdx
+++ b/docs/develop/worker-performance.mdx
@@ -176,6 +176,15 @@ The above figure shows Eager Workflow Start in action:
 
 To recover from errors, Eager Workflow Start falls back to the non-eager path. For example, when the first Task is returned eagerly, but the local Worker fails or times out while processing the task, the server retries this task non-eagerly after WorkflowTaskTimeout.
 
+## Visualize Workers in the UI {#visualize-workers}
+
+The Temporal SDK includes a heartbeat for Worker processes that is sent to the Server every 60 seconds, carrying information such as available task slots, CPU usage, and Worker configuration.
+The Server exposes this data through APIs that surface Worker details in the Temporal UI.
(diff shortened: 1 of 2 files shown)


docs/develop/worker-performance.mdx
```diff
@@ -176,6 +176,15 @@ The above figure shows Eager Workflow Start in action:
 
 To recover from errors, Eager Workflow Start falls back to the non-eager path. For example, when the first Task is returned eagerly, but the local Worker fails or times out while processing the task, the server retries this task non-eagerly after WorkflowTaskTimeout.
 
+## Visualize Workers in the UI {#visualize-workers}
+
+The Temporal SDK includes a heartbeat for Worker processes that is sent to the Server every 60 seconds by default, carrying information such as available task slots, CPU usage, and Worker configuration.
+The Server exposes this data through APIs that surface Worker details in the Temporal UI.
+Use the UI to view the list of Workers assigned to a Task Queue and inspect Worker details to troubleshoot Workflows with delayed processing.
```
rkannan82: > to view the list of workers assigned to a task queue.

It seems like user has to do some filtering to actually get to a task queue specific worker. IIUC, there are 2 entry points.
1. The new workers side nav. This includes all task queues (fleet view).
2. From the workers tab within a workflow view. If a worfklow has multiple task queues, then what do we show up by default?

dustin-temporal: The existing Workflow view actually doesn't support multiple task queues today. I am not sure how it selects which one to show but it only shows one. It does correctly filter down the Worker list for the task queue that was selected. I think multiple task queue support will have to be future work.

dustin-temporal: This is not entirely correct, it shows the Workflow Task's task queue. If you drill into the history of the Activities you can see individual task queues that had activities scheduled on them. It does seem we have a gap in the Worker list from this perspective. I think we can do a follow up there as part of public preview. I'll note that.

rkannan82: Thanks for the context. But this PR does not mention the new Workers nav bar (case 1). I think we should call out that we now have a fleet wide view (nav bar), and then the contextualized view (within a workflow). What do you think?

dustin-temporal: Ah yes, I missed the forest through the trees in your comment! I will update to specify the entry paths now

brianmacdonald-temporal (approved):
Looks good!

PR #4447, opened 2026-04-17
feat(streaming): added docs for streaming feature

flippedcoder:
## What does this PR do?

This adds the docs for Streaming and includes the AI agent use case.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1214127682858398">EDU-6218 feat(streaming): added docs for streaming feature</a>

PR #4513, opened 2026-04-30
docs: Workflow Streams Python guide and encyclopedia concept page

jssmith:
## Summary

- Adds the Python SDK guide for **Workflow Streams** (`temporalio.contrib.workflow_streams`) under a new **Libraries** category, sibling to Workflows / Activities / etc. Workflow Streams is a contrib library, not a core Workflow primitive, so it does not belong alongside Schedules, Timers, Versioning, etc.
- Adds an encyclopedia concept page at `/workflow-streams` covering the streaming problem, with/without-Temporal architecture, publish/subscribe transport pattern, concurrency model, durability/resumability tradeoffs, and an AI agent use case.
- Encyclopedia gets its own **Libraries** category (sibling to Extensibility) so future SDK contrib-library overviews have a natural home.

Supersedes #4447 — that PR predates the `pubsub → workflow_streams` rename and points at a personal-fork sample. The conceptual content from #4447 was adapted into the encyclopedia page here, restated for the post-rename API surface and pointing at the canonical sample in [`temporalio/samples-python`](https://github.com/temporalio/samples-python/tree/main/workflow_streams).

## Test plan

- [x] `yarn build` succeeds
- [x] `yarn start` renders the new pages, sidebar entries, and cross-links
- [ ] Vercel preview review

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1214412026837491">EDU-6295 docs: Workflow Streams Python guide and encyclopedia concept page</a>


docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,387 @@
...
+  the [Temporal 101](https://learn.temporal.io/courses/temporal_101/) course.
+- Ensure you have set up your local development environment by following the
+  [Set up your local development environment](/develop/python/set-up-your-local-python) guide.
+- Familiarity with [Workflow message passing](/develop/python/workflows/message-passing) — Workflow Streams is built on Signals and Updates.
+
+## Concepts
+
+Three roles cover everything you need to use Workflow Streams:
+
+- **A streaming Workflow** hosts the event stream. Constructing a `WorkflowStream` instance from `@workflow.init` registers the handlers that accept publishes and serve subscribers.
```
brianstrauch: Is this a good place to mention that you can only have one WorkflowStream per workflow?

docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,387 @@
...
+- **Publishers** append events. The Workflow itself can publish via the in-memory `WorkflowStream`. Activities and external clients publish via `WorkflowStreamClient`, which batches and ships events to the Workflow.
+- **Subscribers** consume events. They call `WorkflowStreamClient.subscribe(...)` and iterate the returned async iterator.
+
+Events are organized by **topic** — a string label set when publishing. Subscribers can filter to one or more topics or consume all of them. Topics are implicit; publishing to a topic creates it.
+
+The deeper picture — where Signals and Updates fit, how batching works, how state is carried across continue-as-new — appears in [Architecture](#architecture) at the end of this guide. You can skip ahead to it once the basics are wired up.
+
+## Usage patterns
+
+Workflow Streams is built for the case where an observer can show up late — after a page refresh, server restart, or closed laptop — and still see what happened while they were away. Three patterns recur when applying the library; the surrounding rationale is published separately (see [TODO: blog post link]).
```
brianstrauch: Need a review on the blog post?

docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,387 @@
...
+
+**Build consumer state from individual events.** Publishers emit facts in their own vocabulary (`STATUS_CHANGE`, `TOOL_CALL`, `TEXT_DELTA`, and so on) and consumers project those facts into whatever they want to render. Re-running the projection from offset 0 reproduces consumer state, which is what makes reconnection trivially correct. Two event shapes show up in practice:
+
+- *Set-semantic* events are idempotent and last-writer-wins (`STATUS_CHANGE`, lifecycle markers). Safe to replay or deduplicate by key.
+- *Append-semantic* events are cumulative and order-dependent (`TEXT_DELTA` from a token stream). Carry an explicit application-level sequence number on the payload — distinct from the stream's global offset — so the consumer can detect or recover from out-of-order delivery within a topic.
```
brianstrauch: If order is not guaranteed, should the same suggestion of using a sequence number apply for the set-semantic events, too?

brianstrauch: Also, I didn't realize this until now. I guess all of our plugins will need to pass a sequence number while streaming LLM output?

docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,387 @@
...
+        self.status.publish(StatusEvent(state="completed", progress=100))
+```
+
+`publish()` runs the payload converter to encode each value. The codec chain (encryption, compression, and so on) runs once on the Signal or Update envelope that carries the batch — never per item — so encryption and compression are applied exactly once each direction.
+
+The `type=` argument is optional and defaults to `Any`. Pass it when you want the binding recorded so re-binding the same name to an unequal type raises, or so subscribers can pick up the type from the same handle.
+
+## Publish from an Activity
+
+Use `WorkflowStreamClient.from_within_activity()` inside an Activity. The Temporal client and target Workflow Id are taken from the Activity context. The async context manager batches publishes and flushes on exit. Bind a topic handle on the client and publish through it, the same way you would on the Workflow side.
```
brianstrauch:
```suggestion
Use `WorkflowStreamClient.from_within_activity()` inside an Activity. The Temporal client and target Workflow ID are taken from the Activity context. The async context manager batches publishes and flushes on exit. Bind a topic handle on the client and publish through it, the same way you would on the Workflow side.
```

docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,402 @@
...
+[Public Preview](/evaluate/development-production-features/release-stages#public-preview). Refer to the
+[Temporal product release stages guide](/evaluate/development-production-features/release-stages) for more information.
+
+Cross-language client support is on the roadmap. Only the Python client is covered for now.
+
+The API may change before general availability.
+
+:::
+
+For runnable end-to-end examples, see the [Workflow Streams samples](https://github.com/temporalio/samples-python/tree/main/workflow_streams). The directory has four scenarios: basic publish/subscribe with heterogeneous topics, a reconnecting subscriber that demonstrates offset-based resume, an external (non-Activity) publisher pattern, and a long-running ticker that bounds its log via `truncate()`.
```
flippedcoder: This link goes to a 404.

docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,402 @@
...
+- **Publishers** append events. The Workflow itself can publish via the in-memory `WorkflowStream`. Activities and external clients publish via `WorkflowStreamClient`, which batches and ships events to the Workflow.
+- **Subscribers** consume events. They call `WorkflowStreamClient.subscribe(...)` and iterate the returned async iterator.
+
+Events are organized by **topic** — a string label set when publishing. Subscribers can filter to one or more topics or consume all of them. Topics are implicit; publishing to a topic creates it.
+
+The deeper picture — where Signals and Updates fit, how batching works, how state is carried across continue-as-new — appears in [Architecture](#architecture) at the end of this guide. You can skip ahead to it once the basics are wired up.
+
+## Usage patterns
+
+Workflow Streams is built for the case where an observer can show up late — after a page refresh, server restart, or closed laptop — and still see what happened while they were away. Three patterns recur when applying the library. You can find the surrounding rationale in [this blog post](see [TODO: blog post link]).
```
flippedcoder: Do we have the blog post link for here? 😅

docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,402 @@
...
+
+    status = stream.topic("status", type=StatusEvent)
+    async for item in status.subscribe():
+        evt = item.data
+        print(f"[{evt.progress:3d}%] {evt.state}: {evt.detail}")
+        if evt.state == "completed":
+            break
+```
+
+The iterator handles re-polling, server-imposed range limits, and Workflow-side log truncation transparently. There are no extra signals or exceptions for normal callers to handle. See the [API reference](https://python.temporal.io/temporalio.contrib.workflow_streams.html) if you need the underlying details.
```
flippedcoder: This link goes to a 404.

docs/develop/python/libraries/workflow-streams.mdx
```diff
@@ -0,0 +1,402 @@
...
+
+    The fix is to make the handler `async` and `await asyncio.sleep(0)` before reading state. That `yield` lets pending publishes apply with no Temporal timer or history events. Don't substitute `workflow.sleep(0)`, which schedules a timer. Most handlers are already safe like, anything that `await`s before reading state, the module's own poll handler, and Workflow-internal publishes.
+- **Response size cap.** Poll responses are capped at roughly 1 MB. Subscribers repoll immediately when the server caps a response. No special handling is required, but high-throughput producers should expect a steady stream of small batches rather than occasional large ones.
+
+## See also
+
+- [Workflow Streams samples (samples-python)](https://github.com/temporalio/samples-python/tree/main/workflow_streams) — four runnable scenarios covering basic publish/subscribe, reconnecting subscribers, external publishers, and bounded logs via `truncate()`.
```
flippedcoder: This link goes to a 404.

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,402 @@
...
+
+    The fix is to make the handler `async` and `await asyncio.sleep(0)` before reading state. That `yield` lets pending publishes apply with no Temporal timer or history events. Don't substitute `workflow.sleep(0)`, which schedules a timer. Most handlers are already safe like, anything that `await`s before reading state, the module's own poll handler, and Workflow-internal publishes.
+- **Response size cap.** Poll responses are capped at roughly 1 MB. Subscribers repoll immediately when the server caps a response. No special handling is required, but high-throughput producers should expect a steady stream of small batches rather than occasional large ones.
+
+## See also
+
+- [Workflow Streams samples (samples-python)](https://github.com/temporalio/samples-python/tree/main/workflow_streams) — four runnable scenarios covering basic publish/subscribe, reconnecting subscribers, external publishers, and bounded logs via `truncate()`.
+- [`temporalio.contrib.workflow_streams` API reference](https://python.temporal.io/temporalio.contrib.workflow_streams.html).
```
flippedcoder, line 565: This link goes to a 404.

flippedcoder (commented):
I made some docs edits on a commit, but it looks great overall! There are a few 404s which I'm guessing are probably waiting on other releases. Once those links are live and you're good to go I can get this approved and merged.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: Let's aim to merge the samples at the same time as the docs. This will then resolve.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: Yes, we can mention that here.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: The blog is unlikely to land this week. Let's add the link later.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: The convention in the docs is to use Id.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: Good catch. This is live now.

docs/develop/python/workflows/workflow-streams.mdx
jssmith: This is live now—following the release.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: Yes, after some revisions.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: Good catch — the framing was muddled. Reworked in 9a0d48e18.

The library actually does deliver in a stable total order (each event lands at a fixed offset; replay from 0 reproduces the sequence) with exactly-once dedup on `(publisher_id, sequence)`. So there is no "out-of-order delivery within a topic" to worry about. What the prior bullet was really gesturing at was Activity retries: a retried Activity uses a fresh `WorkflowStreamClient` (new `publisher_id`), so its events are appended to the log alongside the failed attempt's. That's an application-layer concern, not a transport-layer one.

So no, set-semantic events don't need an app-level sequence number, and neither do append-semantic ones — both are reconciled by an idempotent consumer reducer. For LLM token streams in the plugins specifically, the conventional pattern is an `AGENT_START` (or equivalent) sentinel that resets the accumulator before the new attempt's deltas resume — so the plugins don't need to add a per-token sequence; they just need to publish the start sentinel on each attempt.

Also added an explicit Ordering paragraph to the Delivery semantics section covering total order, per-publisher order, and cross-publisher interleaving.

docs/develop/python/libraries/workflow-streams.mdx
flippedcoder: I'll go ahead and remove this line for now and we can add it back once the blog post is ready.

docs/develop/python/libraries/workflow-streams.mdx
jssmith: Same situation as the link on line 45 — the samples-python rename PR is open (temporalio/samples-python#300) and the URL will resolve once it merges. We aim to land the samples alongside this docs PR.

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+
+**Looking for...**
+
+- Runnable end-to-end samples (basic publish/subscribe, reconnecting subscriber, external publisher, bounded log): [Workflow Streams samples](https://github.com/temporalio/samples-python/tree/main/workflow_streams).
+- A complete LLM streaming example on this page (Activity publishes deltas, terminal consumer that resets on retry): [Stream LLM output](#stream-llm-output).
+- Delivery guarantees, ordering, and retry semantics: [Delivery semantics](#delivery-semantics).
+- History-size cost and tuning: [Architecture](#architecture).
+- Long-running streams that need Continue-As-New: [Continue-As-New](#continue-as-new).
+
+## Concepts
```
lennessyy: We isolate conceptual content to encyclopedia pages. SDK guides need to stay task focused.

docs/develop/python/libraries/index.mdx
```diff
@@ -0,0 +1,21 @@
+---
+id: index
+title: Python SDK libraries
+sidebar_label: Libraries
```
lennessyy: We should not create a category called Libraries. Libraries doesn't map to a task a user is trying to accomplish. And this would be inconsistent with how we talk about content from `contrib` packages in general.  

Instead, I think we should split the content here under the different primitives we already have. I see streaming on Workflows and Clients on this page. Perhaps create a page called Streaming under Workflows, which covers streaming and publishing on a workflow, and then another page titled Streaming under Clients to talk about how to do the same on Clients.

lennessyy (changes requested):
There is a lot of good content here, but this page is way too long and we need to break it down into more a few more focused pages in better labeled sections in the docs. I offered some suggestions here.

docs/develop/python/libraries/index.mdx
jssmith: Putting it under Workflows is where we started, but this is not a core Temporal feature and we want to be sure the positioning is clear.

Maybe we call the page "Workflow Streams Library"

I understand the reluctance to add a Libraries tab, particularly as it isn't an established pattern and we have only one right now. I expect to see more Libraries, but moving + redirecting if we do so isn't that hard. 

We should definitely not split the content. That will make it difficult to wrap your head around the feature (for humans + agents).

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+  - Workflows
+  - Messages
+  - Streaming
+  - Python SDK
+  - Temporal SDKs
+description: Stream events from a Workflow to subscribers using the Temporal Python SDK Workflow Streams contrib module.
+---
+
+**Workflow Streams** is a Temporal Python SDK `contrib` library that gives a Workflow a durable, offset-addressed event channel built on Temporal's basic message primitives: Signals, Updates, and Queries.
+It batches publishes to amortize per-Signal cost, deduplicates batches for exactly-once delivery to the log, supports topic filtering, and carries stream state across Continue-As-New for long-running runs.
```
brianstrauch: Awkward wording
```suggestion
It batch-publishes events to amortize per-Signal cost, deduplicates batches for exactly-once delivery to the log, supports topic filtering, and carries state across Continue-As-New for long-running streams.
```

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+
+Use Workflow Streams when you want outside observers to follow the progress of a Workflow and its Activities: updating a UI as an AI agent works, surfacing status from a payment or order pipeline, reporting intermediate results from a data job. It is not suited to ultra-low-latency cases like real-time voice, and it targets modest fan-out: tens of publishers and subscribers per Workflow, not thousands.
```
brianstrauch:
```suggestion
Use Workflow Streams when you want outside observers to follow the progress of a Workflow and its Activities: updating a UI as an AI agent works, surfacing status from a payment or order pipeline, or reporting intermediate results from a data job. It is not suited to ultra-low-latency cases like real-time voice, and it targets modest fan-out: tens of publishers and subscribers per Workflow, not thousands.
```

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+
+## Where the stream lives
+
+A `WorkflowStream` is hosted inside a Workflow, so the first design choice is whether one Workflow handles both the work and the stream, or whether a separate Workflow exists only to host the stream. The choice is mostly about lifecycle.
+
+**Host the stream on the Workflow that does the work** when the events come from what that Workflow is already orchestrating: an agent run, an order pipeline, a chat session. The stream's lifecycle aligns with the run, starting when the run starts and ending when it returns. The Workflow Id you use to start the work is the same one subscribers attach to. This is the common shape for AI agents and most progress-streaming cases, where streaming is just one more thing the Workflow does as part of its job.
```
brianstrauch:
```suggestion
**Host the stream on the Workflow that does the work** when the events come from what that Workflow is already orchestrating: an agent run, an order pipeline, a chat session. The stream's lifecycle aligns with the run, starting when the run starts and ending when it returns. The Workflow ID you use to start the work is the same one subscribers attach to. This is the common shape for AI agents and most progress-streaming cases, where streaming is just one more thing the Workflow does as part of its job.
```

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+
+**Use a dedicated Workflow for the stream alone** when the stream should outlive any single producer, accept fan-in from multiple unrelated sources, or be subscribable before any work has started. Producers publish from outside the stream Workflow (Activities of other Workflows, or external `WorkflowStreamClient` instances). The trade-off is explicit lifecycle management: a dedicated stream Workflow does not terminate on its own, so you need a signal-driven shutdown or a Continue-As-New strategy.
+
+Whichever shape you pick, the Workflow Id is the address subscribers use to attach. Multiple subscribers can attach to the same Id concurrently, which is the normal case for a UI with multiple browser tabs. Use distinct Workflow Ids for unrelated streams rather than packing them into one Workflow.
```
brianstrauch:
```suggestion
Whichever shape you pick, the Workflow ID is the address subscribers use to attach. Multiple subscribers can attach to the same ID concurrently, which is the normal case for a UI with multiple browser tabs. Use distinct Workflow IDs for unrelated streams rather than packing them into one Workflow.
```

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+        self.status.publish(StatusEvent(state="completed", progress=100))
+```
+
+`publish()` runs the payload converter to encode each value. The codec chain (encryption, compression, and so on) runs once on the Signal or Update envelope that carries the batch, never per item, so encryption and compression are applied exactly once each direction.
+
+The `type=` argument is optional and defaults to `Any`. Pass it when you want the binding recorded so re-binding the same name to an unequal type raises, or so subscribers can pick up the type from the same handle.
+
+## Publish from a client
+
+Any process that has a Temporal `Client` and the target Workflow Id can publish to that Workflow's stream by constructing a `WorkflowStreamClient`. This is the general pattern and covers HTTP backends, starters, one-off scripts, other Workflows' Activities, and standalone Activities. Construct one with `WorkflowStreamClient.create(client, workflow_id)`, then use it the same way you would the Workflow-side handle: bind a topic, publish through it, and let the async context manager flush on exit.
```
brianstrauch:
```suggestion
Any process that has a Temporal `Client` and the target Workflow ID can publish to that Workflow's stream by constructing a `WorkflowStreamClient`. This is the general pattern and covers HTTP backends, starters, one-off scripts, other Workflows' Activities, and standalone Activities. Construct one with `WorkflowStreamClient.create(client, workflow_id)`, then use it the same way you would the Workflow-side handle: bind a topic, publish through it, and let the async context manager flush on exit.
```

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+        batch_interval=timedelta(milliseconds=200),
+    )
+    async with stream_client:
+        status = stream_client.topic("status", type=StatusEvent)
+        status.publish(StatusEvent(state="started"))
+        ...
+        # Buffer is flushed on context manager exit.
+```
+
+Inside an Activity scheduled by a Workflow, `WorkflowStreamClient.from_within_activity()` is a convenience that infers the Temporal `Client` and the parent Workflow Id from the Activity context, so you don't have to thread them through the Activity's input:
```
brianstrauch:
```suggestion
Inside an Activity scheduled by a Workflow, `WorkflowStreamClient.from_within_activity()` is a convenience that infers the Temporal `Client` and the parent Workflow ID from the Activity context, so you don't have to thread them through the Activity's input:
```

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,589 @@
...
+            text = chunk.choices[0].delta.content
+            if not text:
+                continue
+            # force_flush only on the first delta so the user sees something
+            # immediately; subsequent deltas batch at the 200 ms interval.
+            deltas.publish(TextDelta(text=text), force_flush=first)
+            first = False
+            full.append(text)
+        complete.publish(TextComplete(full_text="".join(full)))
+    return "".join(full)
```
brianstrauch: Why is it useful to publish the full text if the Activity also returns it?

jssmith: Let's delete this - activity can return.

docs/develop/python/libraries/workflow-streams.mdx
brianstrauch: Resolved: https://github.com/temporalio/samples-python/tree/main/workflow_streams

docs/develop/python/libraries/workflow-streams.mdx
brianstrauch: Resolved: https://github.com/temporalio/samples-python/tree/main/workflow_streams

docs/develop/python/workflows/workflow-streams.mdx
```diff
@@ -0,0 +1,585 @@
...
+
+**Looking for...**
+
+- Runnable end-to-end samples (basic publish/subscribe, reconnecting subscriber, external publisher, bounded log): [Workflow Streams samples](https://github.com/temporalio/samples-python/tree/main/workflow_streams).
+- A complete LLM streaming example on this page (Activity publishes deltas, terminal consumer that resets on retry): [Stream LLM output](#stream-llm-output).
+- Delivery guarantees, ordering, and retry semantics: [Delivery semantics](#delivery-semantics).
+- History-size cost and tuning: [Architecture](#architecture).
+- Long-running streams that need Continue-As-New: [Continue-As-New](#continue-as-new).
+
+## Concepts
```
lennessyy: Wanted to flag this concern one more time. I am assuming when the other languages come in we'll want to move this content out even if you keep it here today. 

The way it is now, we'll end up with Python specific links--all external inks used in marketing materials--that we will have a hard time redirecting because this page will still be live. If you are good with that, go head and merge. Otherwise if it's not too much work, I'd still do the work now to write the conceptual docs in a language agnostic way to avoid headaches later.

lennessyy (approved):
I don't want to block this because it is replay time and i understand the crunch. But just wanted to flag the conceptual content location one more time.

docs/develop/python/workflows/workflow-streams.mdx
brianstrauch: Will follow up with an encyclopedia page in a separate PR!

PR #4519, opened 2026-05-01
Remove outdated Workflow Update notes

drewhoskins-temporal:
## What does this PR do?
These instructions are for very old Temporal Server versions released in 2024.  Taking up a lot of space for the low value they provide.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1214441932867320">EDU-6301 Remove outdated Workflow Update notes</a>


brianmacdonald-temporal:
Hey, Drew, I'd rather keep these unless they're actively causing confusion. It's valid info for legacy users.

drewhoskins-temporal:
I think it could cause confusion if people copy commandlines without reading carefully, but perhaps more importantly, we no longer support using workflow update on the versions that these address.

Duncanma:
Hey folks, I was looking into the % of people that use versions < 1.25.0, but I see this from @drewhoskins-temporal "but perhaps more importantly, we no longer support using workflow update on the versions that these address."

So it sounds to me like this isn't a matter of "people on these old versions need these commands", but either remove it completely (this PR) or modify to say "not supported until 1.25.0" ? (fyi @brianmacdonald-temporal )

Duncanma:
A large % of users are still < 1.25, so closing this for now.

PR #4595, opened 2026-05-19
Clarify Text CSA tokenization behavior and add type selection guidance

austonli:
## Summary
- Adds a **"Choose a string type"** section to the Search Attributes page with a comparison table for Keyword vs KeywordList vs Text, including indexing behavior, supported operators, and best-use guidance
- Expands the **Text matching** section in the List Filter page with concrete tokenization examples showing how common value shapes (dotted class names, hyphenated IDs, file paths) are split into tokens
- Adds caution admonitions on both pages warning that Text type splits structured identifiers at punctuation, with cross-links between the two pages
- Adds a forward-looking note that broader tokenizer improvements are being evaluated as part of future Visibility enhancements

## Context
Customers frequently create Text CSAs for structured identifiers (workflow type names, UUIDs, hyphenated IDs) expecting exact-match or substring behavior, then hit confusing search failures because the standard tokenizer silently splits at punctuation. The existing docs mention the distinction briefly but don't explain the implications clearly enough to prevent this mistake.

## Test plan
- [x] `npm run build` succeeds with no errors
- [x] Built HTML verified: tables, admonitions, and cross-links render correctly on both pages
- [ ] Visual review of `/search-attribute#choose-a-string-type`
- [ ] Visual review of `/list-filter#text`

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1214950351015754">EDU-6394 Clarify Text CSA tokenization behavior and add type selection guidance</a>

### DIFF
--- a/docs/encyclopedia/visibility/list-filter.mdx
+++ b/docs/encyclopedia/visibility/list-filter.mdx
@@ -57,7 +57,8 @@ There are different options for partial string matching when the type of the Sea
 
 Search Attributes of type `Text` are [broken up into words](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html) that match with the `=` operator.
 
-The standard tokenizer splits values on Unicode word boundaries — punctuation characters like `.`, `-`, `/`, `_`, and `:` act as delimiters.
+The standard tokenizer splits values on [Unicode word boundaries](https://unicode.org/reports/tr29/) — punctuation characters like `.`, `-`, `/`, and `:` act as delimiters.
+Underscores do not cause splits (`payment_retry` stays as one token).
 All tokens are lowercased.
 The following table shows how common value shapes are tokenized:
 
--- a/docs/encyclopedia/visibility/list-filter.mdx
+++ b/docs/encyclopedia/visibility/list-filter.mdx
@@ -56,15 +56,15 @@ There are different options for partial string matching when the type of the Sea
 #### Text
 
 `Text` Search Attributes support word-level search.
-Both stored values and query strings are broken into tokens by the [Elasticsearch standard tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html), and the `=` operator returns a result if any query token matches any indexed token.
+Both stored values and query strings are analyzed by the [Elasticsearch standard analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html), and the `=` operator returns a result if any query token matches any indexed token.
 
 If you need exact matching on identifiers, UUIDs, or other structured strings, use `Keyword` instead.
 See [Choose a string type](/search-attribute#choose-a-string-type).
 
 ##### Tokenization {#text-tokenization}
 
-The standard tokenizer applies [Unicode word boundary rules](https://unicode.org/reports/tr29/) to split values into tokens.
-All tokens are lowercased.
+The standard analyzer applies [Unicode word boundary rules](https://unicode.org/reports/tr29/) to split values into tokens.
+All tokens are lowercased by the analyzer's lowercase filter.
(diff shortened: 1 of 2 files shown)


jsundai (approved):

docs/encyclopedia/visibility/list-filter.mdx
```diff
@@ -55,7 +55,29 @@ There are different options for partial string matching when the type of the Sea
...
+`Text` Search Attributes support word-level search.
+Both stored values and query strings are broken into tokens by the [Elasticsearch standard tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html), and the `=` operator returns a result if any query token matches any indexed token.
+
+If you need exact matching on identifiers, UUIDs, or other structured strings, use `Keyword` instead.
+See [Choose a string type](/search-attribute#choose-a-string-type).
+
+##### Tokenization {#text-tokenization}
+
+The standard tokenizer applies [Unicode word boundary rules](https://unicode.org/reports/tr29/) to split values into tokens.
+All tokens are lowercased.
```
jsundai: This section is going to save so many people from headaches. One tiny thing on this line: the lowercase behavior seems to come from the standard analyzer's lowercase token filter rather than the tokenizer itself (the tokenizer alone would leave `QUICK` as `QUICK`). Since Temporal uses the standard analyzer for `Text` fields, everything written is true in practice but it might just be worth swapping "standard tokenizer" for "standard analyzer" so it holds up if someone goes digging in the ES docs. Super minor, great work on this!

jsundai:
```suggestion
All tokens are lowercased by the analyzer's lowercase filter.
```

docs/encyclopedia/visibility/list-filter.mdx
```diff
@@ -55,7 +55,29 @@ There are different options for partial string matching when the type of the Sea
 
 #### Text
 
-Search Attributes of type `Text` are [broken up into words](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html) that match with the `=` operator.
+`Text` Search Attributes support word-level search.
+Both stored values and query strings are broken into tokens by the [Elasticsearch standard tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html), and the `=` operator returns a result if any query token matches any indexed token.
```
jsundai:
```suggestion
Both stored values and query strings are analyzed by the [Elasticsearch standard analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html), and the `=` operator returns a result if any query token matches any indexed token.
```

docs/encyclopedia/visibility/list-filter.mdx
```diff
@@ -55,7 +55,29 @@ There are different options for partial string matching when the type of the Sea
...
+If you need exact matching on identifiers, UUIDs, or other structured strings, use `Keyword` instead.
+See [Choose a string type](/search-attribute#choose-a-string-type).
+
+##### Tokenization {#text-tokenization}
+
+The standard tokenizer applies [Unicode word boundary rules](https://unicode.org/reports/tr29/) to split values into tokens.
+All tokens are lowercased.
+The splitting behavior depends on the delimiter:
+
+| Delimiter | Splits? | Example |
```
jsundai: Love the tokenization table! Heads up on an edge case that might be worth double checking, a dot between a digit and a letter may split, so something like v1.ProcessOrder might tokenize into v1 and ProcessOrder rather than staying as one token. Could be worth a small note since that pattern shows up a lot in workflow names such as "Note: A dot between a digit and a letter (for example, v1.ProcessOrder) may split into separate tokens." Either way this is a great addition!

jsundai:
Great, let's get this merged!

Issue #3875, opened 2025-09-29
.NET SDK missing info on interceptors


GSmithApps:
### Temporal Affiliation

- [x] I am a Temporal employee

### Page URL

https://docs.temporal.io/develop/dotnet/interceptors

### Content Issue

Some users noted that there is no interceptor documentation in the .NET guide -- can we add it? 🙏 

For example:
- TS has https://docs.temporal.io/develop/typescript/interceptors
- Python has https://docs.temporal.io/develop/python/interceptors
- but .NET doesn't have https://docs.temporal.io/develop/dotnet/interceptors

Examples of users asking about this include, but are not limited to, the following: [#1](https://temporalio.slack.com/archives/CTRCR8RBP/p1683934148834849?thread_ts=1683934148.834849&cid=CTRCR8RBP) and [#2](https://temporalio.slack.com/archives/C012SHMPDDZ/p1716361765798549?thread_ts=1716361765.798549&cid=C012SHMPDDZ)

In the meantime, perhaps we can direct users to the .NET samples that use interceptors: [ContextPropagation sample](https://github.com/temporalio/samples-dotnet/tree/main/src/ContextPropagation) and [CounterInterceptor sample](https://github.com/temporalio/samples-dotnet/tree/main/src/CounterInterceptor).

Akash29g:
hi @GSmithApps , i would like to work on this, could u assign this to me

PR #4603, opened 2026-05-21
feat(interceptors): added Interceptor page for .NET

flippedcoder:
## What does this PR do?

Addresses #3875. .NET didn't have a page for interceptors and it's a request that's been brought up in the community a few times.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215008297828339">EDU-6406 feat(interceptors): added Interceptor page for .NET</a>

### DIFF
--- a/docs/develop/dotnet/workers/interceptors.mdx
+++ b/docs/develop/dotnet/workers/interceptors.mdx
@@ -50,7 +50,7 @@ request and response data.
 
 ### Register on the Client 
 
-Pass interceptors in the `Interceptors` argument of [`TemporalClientConnectOptions`](https://dotnet.temporal.io/api/Temporalio.Client.TemporalClientConnectOptions.html). Client interceptors modify outbound calls such
+Pass interceptors in the `Interceptors` property of [`TemporalClientConnectOptions`](https://dotnet.temporal.io/api/Temporalio.Client.TemporalClientConnectOptions.html). Client interceptors modify outbound calls such
 as starting and signaling Workflows. One example is [setting up tracing](/develop/dotnet/platform/observability) to see your call graph of a Workflow.
 
 ```csharp
@@ -106,6 +106,17 @@ using Temporalio.Api.Common.V1;
 using Temporalio.Client;
 using Temporalio.Client.Interceptors;
 
+public static class UserContext
+{
+    private static readonly AsyncLocal<string?> CurrentUser = new();
+
+    public static string? UserId
+    {
+        get => CurrentUser.Value;
+        set => CurrentUser.Value = value;
+    }
+}
+
 public class ContextPropagationInterceptor : IClientInterceptor
 {
     public ClientOutboundInterceptor InterceptClient(
@@ -120,13 +131,14 @@ public class ContextPropagationClientOutboundInterceptor(
     public override Task<WorkflowHandle<TWorkflow, TResult>>
         StartWorkflowAsync<TWorkflow, TResult>(StartWorkflowInput input)
     {
-        input.Headers["user-id"] = new Payload
+        var headers = input.Headers ?? new Dictionary<string, Payload>();
+        headers["user-id"] = new Payload
         {
             Metadata = { ["encoding"] = ByteString.CopyFromUtf8("plain/text") },
             Data = ByteString.CopyFromUtf8(UserContext.UserId),
         };
 
-        return base.StartWorkflowAsync<TWorkflow, TResult>(input);
+        return base.StartWorkflowAsync<TWorkflow, TResult>(input with { Headers = headers });
     }
 }
 ```
--- a/docs/develop/dotnet/workers/interceptors.mdx
+++ b/docs/develop/dotnet/workers/interceptors.mdx
@@ -16,19 +16,27 @@ tags:
 
 Interceptors are SDK hooks that let you intercept inbound and outbound Temporal calls. You use them to apply shared
 behavior across many calls, such as tracing and authorization, before calls reach the application code and after they return.
-This is similar to middleware in other frameworks.
+This is similar to middleware in other frameworks, like [ASP.NET Core](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware).
 
 There are two main types of interceptors: inbound and outbound.
 
 * Outbound interceptors wrap network calls, running before they reach the network and after they return.
 * Inbound interceptors run after the network hop, wrapping application code and running before it starts and after it returns.
 
-Those further break down into different interceptor types.
+Concretely, there are five categories of inbound and outbound calls that you can modify in this way:
+
+| | [Outbound Client](https://dotnet.temporal.io/api/Temporalio.Client.Interceptors.ClientOutboundInterceptor.html) | [Inbound Workflow](https://dotnet.temporal.io/api/Temporalio.Worker.Interceptors.WorkflowInboundInterceptor.html) | [Outbound Workflow](https://dotnet.temporal.io/api/Temporalio.Worker.Interceptors.WorkflowOutboundInterceptor.html) | [Inbound Activity](https://dotnet.temporal.io/api/Temporalio.Worker.Interceptors.ActivityInboundInterceptor.html) | [Outbound Activity](https://dotnet.temporal.io/api/Temporalio.Worker.Interceptors.ActivityOutboundInterceptor.html) |
+| --- | --- | --- | --- | --- | --- |
+| **Description** | Wraps calls from your application to the Temporal Client to start a Workflow or send [Messages](/encyclopedia/workflow-message-passing/) to it | Wraps calls arriving into a [Workflow Execution](/workflow-execution), such as executing the Workflow, handling [Messages](/encyclopedia/workflow-message-passing/) | Wraps calls a [Workflow](/workflow-definition) makes to the SDK, such as scheduling [Activities](/activities), starting [Child Workflows](/child-workflows), and invoking [Nexus Operations](/nexus) | Wraps calls arriving into an [Activity Execution](/activity-execution) | Wraps calls an [Activity](/activities) makes to the SDK, such as sending [Heartbeats](/encyclopedia/detecting-activity-failures#activity-heartbeat) and reading Activity info |
+| **Runs on** | Client | Worker (Workflow sandbox) | Worker (Workflow sandbox) | Worker (Activity context) | Worker (Activity context) |
+| **Example methods** | `StartWorkflowAsync()`, `WorkflowHandle.SignalAsync()`, `ListWorkflowsAsync()` | `ExecuteWorkflowAsync()`, `WorkflowHandle.QueryAsync()`, `WorkflowHandle.SignalAsync ()`, `WorkflowHandle.ExecuteUpdateAsync()` | `StartActivityAsync()`, `StartChildWorkflowAsync()`, `ChildWorkflowHandle.SignalAsync()`, `StartNexusOperationAsync()` | `ExecuteActivityAsync()` | `Info()`, `Heartbeat()` |
 
 :::warning Workflow interceptors and replay
 
 Workflow inbound and outbound interceptor methods also execute during [replay](/develop/dotnet/best-practices/testing-suite#replay). Use replay-safe APIs for logging, randomness, and time in these interceptors.
 See [Develop Workflow logic](/develop/dotnet/workflows/basics#workflow-logic-requirements) for details.
+
+If you want to write generic code shared by all inbound Workflow call handlers but want to skip read-only operations, check [`Workflow.Unsafe.IsReplaying`](https://dotnet.temporal.io/api/Temporalio.Workflows.Workflow.Unsafe.html#Temporalio_Workflows_Workflow_Unsafe_IsReplaying).
 
 Activity and Client interceptors are not affected by replay.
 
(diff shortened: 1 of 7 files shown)


docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+var client = await TemporalClient.ConnectAsync(new()
+{
+    TargetHost = "localhost:7233",
+    Interceptors = [interceptor],
+});
+```
+
+The `Interceptors` list can contain multiple interceptors.
+In this case they form a chain: a method implemented on an interceptor instance in the list can perform side effects, and modify the data, before passing it on to the corresponding method on the next interceptor in the list.
```
brianmacdonald-temporal: I'm a little lost by "in this case they form a chain." That makes it sound like this code is a special case. But that's the default behavior for interceptors?

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+### Register via a Plugin
+
+If you're building a reusable library or want to bundle interceptors with other primitives, you can register them through a [Plugin](/develop/plugins-guide#interceptors).
+
+### Register on the Worker only 
+
+If your interceptor doesn't affect the Client, you can pass interceptors in the `interceptors` argument of `Worker()`.
```
brianmacdonald-temporal: Looking at the code below, I think this sentence should read "in the `Interceptors` argument of `TemporalWorker()`. (note the capitalized I)

brianmacdonald-temporal (commented):
Just a couple of small notes

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+  - interceptors
+tags:
+  - Interceptors
+  - .NET SDK
+  - Temporal SDKs
+---
+
+Interceptors are SDK hooks that let you intercept inbound and outbound Temporal calls. You use them to apply shared
+behavior across many calls, such as tracing and authorization, before calls reach the application code and after they return.
+This is similar to middleware in other frameworks.
```
cconstable: I like this line. I think it helps people immediately reify the idea.

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+Registering an interceptor means supplying an interceptor instance to the SDK so Temporal can invoke it when matching
+Client or Worker calls occur. Once registered, the interceptor runs as part of the call path and can observe or modify
+request and response data.
+
+### Register on the Client 
+
+Pass interceptors in the `Interceptors` argument of `TemporalClient.ConnectAsync`. Client interceptors modify outbound calls such
+as starting and signaling Workflows.
+
+```dotnet
```
jmaeagle99: These should be csharp

docs/develop/dotnet/workers/interceptors.mdx
jmaeagle99: Python calls out some specific middleware examples. Analogous example for .NET would be https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+Interceptors are SDK hooks that let you intercept inbound and outbound Temporal calls. You use them to apply shared
+behavior across many calls, such as tracing and authorization, before calls reach the application code and after they return.
+This is similar to middleware in other frameworks.
+
+There are two main types of interceptors: inbound and outbound.
+
+* Outbound interceptors wrap network calls, running before they reach the network and after they return.
+* Inbound interceptors run after the network hop, wrapping application code and running before it starts and after it returns.
+
+Those further break down into different interceptor types.
```
jmaeagle99: Should we include a table or list of the interceptor types, what they are used for, where they run, and examples of things that call them?

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+## Register an Interceptor {#register}
+
+Registering an interceptor means supplying an interceptor instance to the SDK so Temporal can invoke it when matching
+Client or Worker calls occur. Once registered, the interceptor runs as part of the call path and can observe or modify
+request and response data.
+
+### Register on the Client 
+
+Pass interceptors in the `Interceptors` argument of `TemporalClient.ConnectAsync`. Client interceptors modify outbound calls such
```
jmaeagle99: _Technically_ not correct. `Interceptors` is a property on `TemporalClientConnectOptions` (which is what the `new()` is creating in the following example).

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+The `Interceptors` list can contain multiple interceptors.
+The default behavior for interceptors is to form a chain. A method implemented on an interceptor instance in the list can perform side effects, and modify the data, before passing it on to the corresponding method on the next interceptor in the list.
+
+### Register via a Plugin
+
+If you're building a reusable library or want to bundle interceptors with other primitives, you can register them through a [Plugin](/develop/plugins-guide#interceptors).
+
+### Register on the Worker only 
+
+If your interceptor doesn't affect the Client, you can pass interceptors in the `Interceptors` argument of `TemporalWorker()`.
```
jmaeagle99: Same as https://github.com/temporalio/documentation/pull/4603/changes#r3284706388, but the options type is `TemporalWorkerOptions`. The sample below shows it correctly.

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+Client or Worker calls occur. Once registered, the interceptor runs as part of the call path and can observe or modify
+request and response data.
+
+### Register on the Client 
+
+Pass interceptors in the `Interceptors` argument of `TemporalClient.ConnectAsync`. Client interceptors modify outbound calls such
+as starting and signaling Workflows.
+
+```dotnet
+var interceptor = new MyCounterInterceptor();
```
jmaeagle99: We'll want all of these code snippets to be in the temporalio/features repo and use [snipsync](https://github.com/temporalio/documentation/blob/main/CONTRIBUTING.md?plain=1#L126) to synchronize them.

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+You can then [register](#register) this interceptor in your client/starter code.
+
+Your interceptor classes don't need to implement every method. The default implementation is always to pass the data on to the next method in the interceptor chain.
+During execution, when the SDK encounters an Inbound Activity call, it will look to the first Interceptor instance, get hold of the appropriate intercepted method, and call it.
+The intercepted method will perform its function then call the same method on the next Interceptor in the chain.
+At the end of the chain the SDK will call the "real" SDK method.
+
+### Implementing Worker call Interceptors
+
+To modify inbound Workflow and Activity calls, define a class implementing [`IWorkerInterceptor`](https://dotnet.temporal.io/api/Temporalio.Worker.Interceptors.IWorkerInterceptor.html). It provides `InterceptActivity()`, `InterceptWorkflow()`, and `InterceptNexusOperation()` methods for Activity, Workflow, and Nexus interception.
```
jmaeagle99, line 155: `InterceptNexusOperation` should be added into the sample below, even if it just returns the next interceptor, just like `InterceptWorkflow`. Using snipsync would reveal this issue because the code would actually be built in the temporalio/features repo.

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+Pass interceptors in the `Interceptors` argument of `TemporalClient.ConnectAsync`. Client interceptors modify outbound calls such
+as starting and signaling Workflows.
+
+```dotnet
+var interceptor = new MyCounterInterceptor();
+
+var client = await TemporalClient.ConnectAsync(new()
+{
+    TargetHost = "localhost:7233",
+    Interceptors = [interceptor],
```
jmaeagle99, line 64: Python sample uses `TracingInterceptor` as the example here, and this is available in the .NET SDK as well. Consider doing the same and maybe refer to the observability doc.

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+If your interceptor doesn't affect the Client, you can pass interceptors in the `Interceptors` argument of `TemporalWorker()`.
+Worker interceptors modify inbound and outbound Workflow and Activity calls.
+
+```dotnet
+using var worker = new TemporalWorker(
+    client,
+    new TemporalWorkerOptions("my-task-queue")
+    {
+        Interceptors = new IWorkerInterceptor[]
```
jmaeagle99: Can we use collection syntax (e.g. `[interceptor]`) instead of the explicit array syntax, just like the above example?

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
+---
```
jmaeagle99, line 1: Is there a page in the encyclopedia that needs to be updated to link to this new page?

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,204 @@
...
+* Outbound interceptors wrap network calls, running before they reach the network and after they return.
+* Inbound interceptors run after the network hop, wrapping application code and running before it starts and after it returns.
+
+Those further break down into different interceptor types.
+
+:::warning Workflow interceptors and replay
+
+Workflow inbound and outbound interceptor methods also execute during [replay](/develop/dotnet/best-practices/testing-suite#replay). Use replay-safe APIs for logging, randomness, and time in these interceptors.
+See [Develop Workflow logic](/develop/dotnet/workflows/basics#workflow-logic-requirements) for details.
+
```
jmaeagle99, line 38: The Python interceptors doc has some more information about writing generic code... The same can be written here and _I think_ the check would be `Workflow.Unsafe.IsReplaying`

docs/develop/dotnet/workers/interceptors.mdx
flippedcoder: Great catch! Thanks for the reminder.

docs/develop/dotnet/workers/interceptors.mdx
flippedcoder: Do you mean essentially make the same table we have for Python? 😅

docs/develop/dotnet/workers/interceptors.mdx
flippedcoder: I'll have to come back and address this comment about moving the code snippets later. 😅 This is an ongoing discussion for where docs snippets should live.

docs/develop/dotnet/workers/interceptors.mdx
flippedcoder: Went ahead and added that table!

jmaeagle99: Yeah, table works. TypeScript describes similar information in a bulleted list. So there's already inconsistency there, but didn't don't feel strong about necessarily unifying everything into the same exact content. But it could yield less ambiguity for readers as well as content authors.

docs/develop/dotnet/workers/interceptors.mdx
jmaeagle99: I'm going to pick on this just a little bit more. `Interceptors` is a _property_, not an _argument_ and not a _field_ (like it is in Python). Properties are first class constructs in C# and this is how C# developers would expect them to be described: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties

docs/develop/dotnet/workers/interceptors.mdx
jmaeagle99: This was recently enabled [here](https://github.com/temporalio/features/tree/main/features/snippets) and you can see a handful of snippets of newer features being added there.

docs/develop/dotnet/workers/interceptors.mdx
jmaeagle99: `input.Headers` is nullable, so if the Headers property is not set, this will throw a `NullReferenceException`. You'll want to probably defend against it with something like:
```csharp
var headers = input.Headers ?? new Dictionary<string, Payload>();
headers["user-id"] = new Payload
{
    Metadata = { ["encoding"] = ByteString.CopyFromUtf8("plain/text") },
    Data = ByteString.CopyFromUtf8(UserContext.UserId),
};

return base.StartWorkflowAsync<TWorkflow, TResult>(input with { Headers = headers });
```

jmaeagle99 (approved):

docs/develop/dotnet/workers/interceptors.mdx
```diff
@@ -0,0 +1,215 @@
...
+    ClientOutboundInterceptor next)
+    : ClientOutboundInterceptor(next)
+{
+    public override Task<WorkflowHandle<TWorkflow, TResult>>
+        StartWorkflowAsync<TWorkflow, TResult>(StartWorkflowInput input)
+    {
+        input.Headers["user-id"] = new Payload
+        {
+            Metadata = { ["encoding"] = ByteString.CopyFromUtf8("plain/text") },
+            Data = ByteString.CopyFromUtf8(UserContext.UserId),
```
jmaeagle99, line 138: Where is `UserContext` coming from? Do we need to import something or change it to something else?

docs/develop/dotnet/workers/interceptors.mdx
flippedcoder: Thanks for the nit! I'm not the most familiar with .NET so this is super helpful.

docs/develop/dotnet/workers/interceptors.mdx
flippedcoder: It was a placeholder example. I updated the snippet to show what it can look like.

brianmacdonald-temporal (approved):

PR #4626, opened 2026-05-28
nexus security clarification - add limitation warning to Option 3 encryption pattern

jsundai:
## What does this PR do?
warn about wrapper type incompatibility with typed operations 

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215206638776751">EDU-6438 nexus security clarification</a>


docs/encyclopedia/nexus/nexus-security.mdx
```diff
@@ -89,6 +89,20 @@ See the [encryption sample](https://github.com/temporalio/samples-go/blob/main/e
 Use wrapper types (for example, `EndpointValue`) so the Data Converter selects an Endpoint-specific encryption key.
 This encrypts only Nexus traffic with a dedicated key, without sharing Namespace keys across teams.
 
+:::warning Limited to string-based operation names
+
+When calling `ExecuteOperation` with a non-string operation (for example, an operation definition or reference), the wrapper changes the input type and `ExecuteOperation` fails type-checking:
+
+```
+cannot assign argument of type "converter.NexusOperation" to type "…EchoInput"
```
codemonkeycxy:
```suggestion
cannot assign argument of type "<WrapperType>" to type "<ActualInputType>"
```

docs/encyclopedia/nexus/nexus-security.mdx
```diff
@@ -89,6 +89,20 @@ See the [encryption sample](https://github.com/temporalio/samples-go/blob/main/e
...
+```
+
+This approach only works when the operation is passed as a string name.
+For typed operations, pass the Nexus endpoint info through the workflow context instead of the wrapped input.
+This requires a Nexus workflow outbound interceptor and a context-aware codec converter, and does not work with synchronous operations.
```
codemonkeycxy: >does not work with synchronous operations

Is this true? I don't have enough knowledge to know for sure, will need to get someone from the Nexus team to confirm

Quinn-With-Two-Ns: I don't think we have a concept of a `context-aware codec converter` so I don't think this section is correct? Where did this information come from?

codemonkeycxy: i think it's referring to a custom implementation like this https://github.com/temporalio/saas-control-plane/blob/94dd6279177060599fef797d4fe647012b7e6f2d/services/codec/converter.go#L73-L86

Quinn-With-Two-Ns: AH but that requires a whole customer `DataConverter`, not just a context-aware codec converter.

jsundai:
Getting back to this sometime today.

jsundai:
@Quinn-With-Two-Ns @codemonkeycxy Coming back to old PRs! How does this look now?

docs/encyclopedia/nexus/nexus-security.mdx
```diff
@@ -83,6 +83,19 @@ See the [encryption sample](https://github.com/temporalio/samples-go/blob/main/e
...
+:::warning Limited to string-based operation names
+
+When calling `ExecuteOperation` with a non-string operation (for example, an operation definition or reference), the wrapper changes the input type and `ExecuteOperation` fails type-checking:
+
+```
+cannot assign argument of type "<WrapperType>" to type "<ActualInputType>"
+```
+
+This approach only works when the operation is passed as a string name.
+For typed operations, pass the Nexus endpoint info through the workflow context instead of the wrapped input.
```
brianmacdonald-temporal:
```suggestion
For typed operations, pass the Nexus endpoint info through the Workflow context instead of the wrapped input.
```

brianmacdonald-temporal (approved):

codemonkeycxy (approved):

PR #4276, opened 2026-03-11
docs: standardize release-stage callouts

bechols:
## What this PR does
- adds a new `ReleaseStage` component to standardize Pre-release and Public Preview callouts
- removes GA callouts where GA is the default state and no callout is needed
- converts release-stage callouts to use the component where the base phrasing fits
- keeps custom wording for special mixed-stage cases like High Availability and Nexus overview sections
- leaves non-release-stage support callouts as `:::tip`

## Notes
- the component supports custom children for exceptions and an optional `verb` prop for plural subjects
- local docs build succeeded after the final changes

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1213624703555124">EDU-6032 docs: standardize release-stage callouts</a>

PR #4644, opened 2026-06-02
feat(components): added new ReleaseNoteHeader component

flippedcoder:
## What does this PR do?

This closes #4276 and implements a new component for release stage announcements on feature pages. The new component allows us to:
- update the stage for features with the same `featureName` at the same time
- show which stage the feature is in
- show which languages are supported
- put custom messages for special conditions

## Notes to reviewers

See [COMPONENTS.md](https://github.com/temporalio/documentation/pull/4644/changes#diff-eb975a80998ce43dadc636ae432d176e7dd2a5bf4259db1cf588678a813c173e) for more detailed instructions.

Examples of new ReleaseNoteHeader:

Screenshot 2026-06-10 at 14 59 22

Screenshot 2026-06-10 at 14 56 37

Screenshot 2026-06-10 at 14 59 47

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215339000262954">EDU-6459 feat(components): added new ReleaseNoteHeader component</a>


bechols:
General suggestion I missed in my original PR: guidance/nudge for pre-release when users need to actively request to participate (i.e. the pre-release feature is invite-only). Doesn't apply for all pre-releases, but when it's the case, really important to avoid confusion.

src/components/info/ReleaseNoteHeader/ReleaseNoteHeader.js
```diff
@@ -0,0 +1,125 @@
...
+    ),
+  };
+}
+
+export default function ReleaseNoteHeader({
+  // type can be "prerelease", "publicPreview", or "generalAvailability"
+  type = "generalAvailability",
+  // name of the feature being released
+  featureName,
+  // If there is anything specific to say about the release, it can be passed as a child to the component. It can also be a link if href is provided.
```
lennessyy, line 56: Can we add some default verbiage that gets shown for each release stage unless we overwrite it? 
Something like we used to say "APIs and configuration may change before the stable release."

src/components/info/ReleaseNoteHeader/ReleaseNoteHeader.js
```diff
@@ -0,0 +1,125 @@
...
+    ...(BASE_RELEASE_STAGES[type] || BASE_RELEASE_STAGES.generalAvailability),
+    ...Object.fromEntries(
+      Object.entries(overrides).filter(([, value]) => value !== undefined)
+    ),
+  };
+}
+
+export default function ReleaseNoteHeader({
+  // type can be "prerelease", "publicPreview", or "generalAvailability"
+  type = "generalAvailability",
```
lennessyy: Do we need to marker for general availability? I think it's better to not have GA release note. If something doesn't have a release note, it's GA by default. Otherwise we might need to add the GA header to everything, because people might wonder why do some GA features have it and some not

flippedcoder: I was a little conflicted on that, but you're right. We don't need a marker for GA since that's the default. I'll just get rid of the references to it.

src/components/info/ReleaseNoteHeader/ReleaseNoteHeader.js
flippedcoder: What do you think about this? It's overwritten if there's anything we put as the child. It'll show on all the notes.

Screenshot 2026-06-10 at 15 46 47

lennessyy (approved):

PR #4647, opened 2026-06-02
ci: add CI check that requires redirects when docs pages are moved or deleted

lennessyy:
## What does this PR do?

- Adds a GitHub Actions workflow that detects renamed or deleted pages in docs/ and verifies that vercel.json contains a matching redirect for each old URL path. Supports both exact and wildcard redirect patterns.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215346083738958">EDU-6462 ci: add CI check that requires redirects when docs pages are moved or deleted</a>


bin/check-redirects-for-moved-pages.test.js
```diff
@@ -0,0 +1,120 @@
+const { describe, it } = require('node:test');
```
flippedcoder, line 1: Just wondering, but when does this test file get run?

flippedcoder (approved):

bin/check-redirects-for-moved-pages.test.js
lennessyy: It doesn't run automatically. It's there when you make changes to the script and want to make sure nothing gets broken. Since these are just unit tests and most of the changes that happen in this repo won't touch these files, I don't think running these tests in CI adds much value

PR #4658, opened 2026-06-04
Add Open in ChatGPT and Open in Claude page actions

bilal-karim:
## Summary

Adds two new external-link actions alongside the existing **Copy for LLM** and **View as Markdown** CTAs under the docs page title:

- **Open in ChatGPT** — opens `chatgpt.com` in a new tab with a prompt prefilled to read the current page
- **Open in Claude** — opens `claude.ai` in a new tab with the same prefilled prompt

The prompt sent to each assistant is: `Read <page-url> and answer questions about the content.`

## Changes

- `src/components/LLMActions/LLMActions.tsx`
  - Capture the page URL client-side (`window` is unavailable during SSR) and build encoded ChatGPT/Claude deep links
  - Render the two new actions as anchors with the OpenAI/Claude brand icons (`react-icons/si`) plus a trailing external-link icon
  - Carry the same `data-analytics-*` attributes as the existing actions (`open-in-chatgpt` / `open-in-claude`) so usage is tracked in Amplitude
- `src/components/LLMActions/LLMActions.module.css`
  - Normalize height between `<button>` and `<a>` elements (`line-height`, `box-sizing`, `font-family`, `text-decoration`)
  - Add a muted, smaller `externalIcon` style

## Testing

- Verified locally via `npm run start`; all four actions render at a consistent height under the page title and the external links open the respective assistants with the page URL prefilled.


alfredsmchan (approved):
LGTM

jsundai (approved):
Nice!! Will take a full look sometime today.

alfredsmchan:
if it fails to get the Page URL, we should hide the buttons unless there is a valid URL.

bilal-karim:
Good call — fixed in 3e09dfc40, though I went a slightly different route than disabling the buttons.

Root cause: the prompt URL was built from `window.location.href` captured in a client-side `useEffect`, so it was empty during SSR and the first render — that's the window where a click would produce `Read  and answer questions...`. (Note: the "failed to fetch" you saw is actually a separate path — that's **Copy for LLM** fetching the raw markdown, not the page-URL resolution. `window.location.href` itself is synchronous and doesn't throw.)

Rather than guard/disable on an empty value, I removed the failure mode entirely: the URL is now derived synchronously from Docusaurus's own data —

```ts
const { siteConfig } = useDocusaurusContext();
const pageUrl = `${siteConfig.url}${metadata.permalink}`;
```

So it's a valid canonical URL during SSR, first paint, and every render — no `useEffect`, no state, no empty-URL window to defend against. Bonus: it's the clean canonical URL (e.g. `https://docs.temporal.io/evaluate/`) rather than whatever query/hash happened to be on `window.location`.

Verified in the browser — the rendered links resolve to:
`Read https://docs.temporal.io/evaluate/ and answer questions about the content.`

alfredsmchan (approved):
Thanks for update!

bilal-karim:
Added dropdown.

https://github.com/user-attachments/assets/59d4fa07-78f7-40f3-bcb2-ba7dcfd66db7

PR #4669, opened 2026-06-06
Clarify AWS PL DNS instructions for single region vs. HA

lukeknep:
## What does this PR do?

 - Clarify the PHZ pattern choice for AWS PrivateLink. The Choose-the-override-domain table now distinguishes the single-region pattern (PHZ on the Namespace Endpoint) from the HA pattern (PHZ on regional records), and warns that the single-region pattern short-circuits HA failover.

  - New section: "How to enable HA on a Namespace using Private Connectivity" — step-by-step for moving a previously single-region private Namespace to the multi-region PHZ layout without dropping DNS resolution along the way.

  - New section: "How to migrate to another Temporal Cloud Region when using Private Connectivity" — the recommended per-region PHZ approach (one PHZ per region, each overriding the Namespace Endpoint locally), with linkage from cloud/migrate/migrate-within-cloud.

## Notes to reviewers

* Some users were confused and were applying both overrides on one Namespace (both Namespace Endpoint and the regional records). Docs need to be clearer that the approach is different for HA.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215469331810271">EDU-6487 Clarify AWS PL DNS instructions for single region vs. HA</a>


jsundai (approved):

jsundai:
LGTM - fixed the build error. With the new [docusaurus core update](https://github.com/temporalio/documentation/commit/746bf4d7b0837375fa02d6af4f6e0bd4ffc937be), we have a new way of doing header IDs (i.e - from {#id} to {/* #id */}).

PR #4683, opened 2026-06-08
Logo a11y update

Duncanma:
## What does this PR do?

Couple of updates. In some cases, like /develop, there was alt text on the language logos, but since they are *links* to the content about that language it is better to not have the alt text say "logo" ... in reality this is link text at that point. So this fixes that by updating the text in the SDK lists. On the homepage, using the SdkLogosAsBlocks component, no alt-text was being provided at all ... and the SVGs didn't have role="img", so screenreaders would try to find inner text to read, it can be messy. So this fixes that too. 

No visual change to the site at all. 

I did consider centralizing that SDK list, but it has a mix of functions, css classes, keys, etc so I am leaving that for a followup discussion.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215519949327215">EDU-6501 Logo a11y update</a>


jsundai (approved):

jsundai:
Nice!!

PR #4691, opened 2026-06-10
feat(codec server): added note about the audience claim value

flippedcoder:
## What does this PR do?

This adds a note for the `aud` value that's in the `Pass access token` so users can validate against it.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215592845343388">EDU-6510 feat(codec server): added note about the audience claim value</a>


brianmacdonald-temporal (approved):
Addition looks good. Confirming that Yogesh verified technical accuracy on Slack.

PR #4712, opened 2026-06-12
Note on sustained payloads at 2MB limit

jpvotta:
## What

Adds a note to the **Transaction Payload size limit** section of the Temporal Cloud limits page clarifying that the 2 MB payload limit is a per-request maximum, not a sustained-throughput target, and that Temporal reserves the right to rate limit high-volume large-payload traffic to protect the service. Points readers to External Storage for workloads that regularly produce large payloads.

## Why

Customers have read the documented 2 MB payload limit as a guarantee that sustained, high-volume 2 MB payloads will be served at scale. In practice, sustained large-payload traffic can degrade Namespace performance even though any single request is within limits.

## Changes

- `docs/evaluate/temporal-cloud/limits.mdx`: added a `:::note` after the 2 MB limit statement.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1215678444422723">EDU-6531 Note on sustained payloads at 2MB limit</a>


lennessyy (approved):

docs/evaluate/temporal-cloud/limits.mdx
```diff
@@ -237,6 +237,14 @@ This limit is non-configurable for Temporal Cloud.
 
 The [BlobSizeLimitError guide](/troubleshooting/blob-size-limit-error) provides solutions for handling large payloads.
 
+:::note
+
+The 2 MB limit is the maximum size for a single request, not a sustained-throughput target.
+Sustained, high-volume use of large payloads can degrade Namespace performance, and Temporal reserves the right to rate limit such traffic to protect the service.
```
lennessyy, line 243:
```suggestion
Sustained, high-volume use of large payloads can degrade Namespace performance, and we may rate limit such traffic to protect the service.
```
I personally feel like "Temporal reserves the right to" sounds a bit aggressive, kind of like we are about to sue them. Making this a little more direct actually sounds softer?

bechols (approved):

PR #4806, opened 2026-07-01
docs(contributing): added section about how we handle page routing

flippedcoder:
## What does this PR do?

We were inconsistently using the `slug` field in the frontmatter and it causes unexpected issues with redirects. This aims to clarify when and why to use the `slug` field.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1216219901598237">EDU-6635 docs(contributing): added section about how we handle page routing</a>


CONTRIBUTING.md
```diff
@@ -134,6 +134,33 @@ The location of the source code is written just inside the wrappers.
...
+   - If no `slug` exists, URL comes from `docs/<path>/<file>.mdx`.
+
+5. Sidebar hierarchy
+   - Controls nav placement only; it does not determine the canonical URL.
+
+We do not encourage the use of the `slug` field in the frontmatter for a page. This field will override the URL and could potentially cause issues with redirects and the sidebar nav.
+
+If you are considering using a custom slug, answer these questions first:
+
+- Does the page _need_ a custom URL? If so, why?
```
lennessyy: I think it would be more helpful for us to define what, if any, valid reasons there are to have slugs?

brianmacdonald-temporal: I'm with you on this, but I haven't been able to figure out a good reason for the slugs, other than "convenience." And because it'll break a lot of links if we remove them.

flippedcoder: Yeah... I wasn't sure if we even want to recommend using slugs at all. I did make an update to this so let me know what y'all think. But honestly, I'm leaning towards removing it and just saying we don't recommend using slugs at all.

Duncanma: I think the update is good for now. We should think through if our current slugs are needed in *any* case (or if we could just move things around) and if so, we can update the doc

Duncanma (approved):

PR #4836, opened 2026-07-07
ci: preserve hand-maintained CLI command-reference index pages on regen

chaptersix:
## Summary

The gen-docs change in temporalio/cli#1112 stops emitting the CLI command-reference **index landing pages** (`command-reference/index.mdx` and `cloud/index.mdx`), because those are hand-maintained on the docs site (custom ordering, the `ReleaseNoteHeader` component). This workflow does `rm -rf docs/cli/command-reference` before regenerating, so without a change here those curated pages would be deleted — breaking the `sidebars.js` references (and the build).

This restores those two files from `HEAD` after regeneration. (The `cp -r` that preserves the `cloud/` subdirectory is already in place.)

## What changed since first review

This PR originally also added a `yarn build` validation step here plus a standalone `validate-cli-docs-build.yml` workflow. Per review (thanks @lennessyy) those were dropped:

- The Vercel preview build is the required check on every PR and already compiles the generated MDX (failing on broken links/anchors), so a second build was redundant.
- Worse, the build ran *before* the PR was opened — on failure, no PR would be created and the error would only surface in Actions run history that nobody watches.

So the PR is now just the index-preservation fix.

## Companion PR

temporalio/cli#1112 — the gen-docs change this pairs with. They should land together (this workflow relies on that PR's `-subdir` flag, and the index restore pairs with it no longer emitting those pages).


.github/workflows/update-cli-docs.yml
```diff
@@ -77,6 +77,43 @@ jobs:
...
+        with:
+          node-version: '24'
+          cache: 'yarn'
+          cache-dependency-path: docs/yarn.lock
+
+      - name: Install dependencies
+        working-directory: docs
+        run: yarn install --frozen-lockfile
+
+      - name: Validate docs build (Docusaurus)
```
chaptersix: is this validated elsewhere already?

lennessyy: I think of the PR preview check as what validates build. If this check fails and a PR is never opened, how are we notified of the failure? Would it be in the Actions run history? I feel like we don't check that as much

chaptersix: Good call — yes. The Vercel preview build is the required check on every PR and already compiles the generated MDX, so this step was redundant. Removed it (and the standalone validate workflow).

chaptersix: You're right, and this is the deciding point — thanks. Since the pre-PR build fails silently (no PR opens, error only in Actions history), it's strictly worse than letting the PR open and having the required Vercel preview build flag it visibly. I've dropped this step, the Node/install steps, and the separate validate-cli-docs-build.yml. The PR is now just the index-page restore, and the Vercel preview build remains the single source of build validation.

.github/workflows/update-cli-docs.yml
```diff
@@ -77,6 +77,24 @@ jobs:
...
+      - name: Copy generated command reference into the docs repo
+        working-directory: docs
+        run: |
+          set -ex
+
+          # Regenerate the command reference pages. cp -r preserves the cloud/
+          # subdirectory produced by -subdir.
+          rm -rf docs/cli/command-reference
+          mkdir -p docs/cli/command-reference
+          cp -r ../cli/dist/docs/* docs/cli/command-reference/
```
ks-temporal: Suggestion: consider using -R instead of -r

ks-temporal (approved):

lennessyy (approved):

.github/workflows/update-cli-docs.yml
chaptersix: Done — switched to `cp -R`. Identical on the GNU coreutils runner, but `-R` is the POSIX-standard flag (`-r` is deprecated on BSD/macOS), so it's the better convention.

PR #4845, opened 2026-07-08
fix(timeouts): added clarification about non-retryable nature of schedule to start timeouts

flippedcoder:
## What does this PR do?

This adds the caveat that `Schedule-To-Start` timeouts are non-retryable by design. The info came from a [community conversation](https://temporalio.slack.com/archives/C04S80QKB2Q/p1733205250777109) with Maxim about it.

## Notes to reviewers

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1216394150007656">EDU-6676 fix(timeouts): added clarification about non-retryable nature of schedule to start timeouts</a>

### DIFF
--- a/docs/develop/java/activities/timeouts.mdx
+++ b/docs/develop/java/activities/timeouts.mdx
@@ -28,9 +28,9 @@ Set your Activity Timeout from the [`ActivityOptions.Builder`](https://www.javad
 
 Available timeouts are:
 
-- ScheduleToCloseTimeout()
-- StartToCloseTimeout()
-- ScheduleToStartTimeout()
+- ScheduleToCloseTimeout
+- StartToCloseTimeout
+- ScheduleToStartTimeout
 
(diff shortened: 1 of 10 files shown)


docs/develop/java/activities/timeouts.mdx
```diff
@@ -29,8 +29,8 @@ Set your Activity Timeout from the [`ActivityOptions.Builder`](https://www.javad
 Available timeouts are:
 
 - ScheduleToCloseTimeout()
```
Duncanma: You didn't add this in this PR, but why would we have ( and ) at the end of these? If we were talking about `.setScheduleToCloseTimeout()` that makes sense, but these terms on their own aren't methods, right?

Duncanma (approved):

docs/develop/java/activities/timeouts.mdx
flippedcoder: Good point. I went ahead and removed those.

PR #4849, opened 2026-07-09
Updates for code showing that run must be public

timjbruce:
## What does this PR do?
Updates the basics page for Rust workflows to show run as a public method (required) instead of a private method.

## Notes to reviewers
Rust workflow cannot be compiled, due to breaking requirements for our SDK.


flippedcoder (approved):
Thanks for the clarification here!

PR #4860, opened 2026-07-10
Add custom og:images

### LINKED ISSUE #4858: Update og:image to be custom per page
See how the Temporal Blog works

Image

And Stripe Docs

Image

And then all docs.temporal.io pages currently use a static single image (a massive one by the way, 5.6MB!!)

Image

This feature will move to a generated image, like this

Image

Duncanma:
### Problem

Every page on docs.temporal.io currently shares a single static `og:image` (`themeConfig.image` in `docusaurus.config.js`) — so a link to any page, whether the homepage or a deep SDK reference page, unfurls in Slack/X/LinkedIn with the exact same generic card. There's no way to give a page a more relevant preview without hand-making an image.

That image, below, is *also* huge (5.6MB)
Image

### What this does

Adds a build-time Docusaurus plugin (`plugins/og-image/`) that generates a unique, on-brand social share card for every doc page automatically — no authoring work required. Each page will get an image like this

Image

These images will show up on Slack, Bluesky, X, etc. 

**Before in slack (note the 6MB in the preview, that's Slack pointing out how large the preview image is)

Screenshot 2026-07-10 at 1 34 11 PM

**After**

Screenshot 2026-07-10 at 1 34 05 PM

### How a card is built

Each card is 1200×630 and includes:
- A section badge derived from the page's location (e.g. `develop/go/...` → **Go SDK**, `cloud/...` → **Cloud**)
- The page title (front matter `title`, falling back to the first `#` heading, falling back to a humanized filename)
- The page description, if present, clamped to 3 lines based on actual rendered width (not a fixed character count, so it always fills the available space without ever overflowing)
- The Temporal logo and `docs.temporal.io`

Rendered with [Satori](https://github.com/vercel/satori) + [`@resvg/resvg-js`](https://github.com/yisibl/resvg-js), then re-encoded through `sharp` as a palette-quantized PNG — cuts output size by ~88% (232.7KB → 28.9KB per card) with no visible quality loss, since the raw resvg PNG output has no compression options and the gradient background otherwise compresses poorly.

### Manual overrides

Any page can opt out of the generated card:
- Front matter `image:` field (existing Docusaurus mechanism), or
- An inline `<Head>` component with its own `og:image`/`twitter:image` meta tags, for cases that need more control

The plugin detects either and leaves the page untouched. See the new "Social share images (og:image)" section in `CONTRIBUTING.md` for authoring instructions.

### Caching

Generated images are cached by content hash (`node_modules/.cache/og-images/`) — nested under `node_modules` so it's gitignored for free and rides along with whatever already caches `node_modules` in CI/Vercel between builds. A page's card only regenerates when its title, description, section, or the template itself changes.

### Performance

Measured across all ~625 doc pages:
- **Warm build** (typical — cache hit): +0s, cards are just copied
- **Cold build** (first build, or after a `yarn.lock` change): ~86s to render every card fresh (~106ms/card)
- **Output size**: 17.7MB total (28.9KB/card average)

`bin/check-og-build-budget.js` (`yarn check:og-budget`) asserts these stay within budget on every build, so a future regression (e.g. a design change that balloons per-card size) gets caught rather than quietly shipped.

### Testing

- `plugins/og-image/index.test.js` — unit tests (`yarn test`) for the override-detection logic against synthetic input, no real build required
- `bin/validate-og-images.js` (`yarn validate:og-images`) — confirms after a real build that every doc page has a correct generated image, override pages (if any) kept their own image, and every page the plugin doesn't touch (404, search, changelog, tag pages, etc.) still falls back to the site default
- `bin/generate-og-gallery.js` — dev-only tool that renders a browsable grid of every generated card for visual review

Closes #4858

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1216464383276668">EDU-6690 Add custom og:images</a>


plugins/og-image/index.js
```diff
@@ -0,0 +1,248 @@
...
+}
+
+function extractTitle(content, frontmatter, id) {
+  if (frontmatter.title) return frontmatter.title;
+  const heading = content.match(/^#\s+(.+)$/m);
+  if (heading) return heading[1].trim();
+  return humanize(id);
+}
+
+const SDK_LABELS = {
```
flippedcoder: In an effort to try and DRY things out, could this be used instead? https://github.com/temporalio/documentation/blob/main/src/constants/sdkLanguages.js#L6

plugins/og-image/render.js
```diff
@@ -0,0 +1,173 @@
...
+
+const CARD_WIDTH = 1200;
+const CARD_HEIGHT = 630;
+
+// Bump this whenever the card layout/design changes (including the encoding
+// step below) so cached images invalidate even though the underlying page
+// content didn't change.
+const TEMPLATE_VERSION = 4;
+
+const GRADIENT_FROM = '#3730a3';
```
flippedcoder: Just wondering, did these colors come from design? And do you think we'll use them elsewhere? They could probably go in the `/src/constants/` folder if we'll use them again.

plugins/og-image/index.js
Duncanma: YES, I would do this, but I ended up removing all that code because the design feedback dropped the 'section' label completely.

plugins/og-image/render.js
Duncanma: modified the code to pull colors from a shared set of constants

flippedcoder (approved):

Issue #4891, opened 2026-07-15
Font flash on first load (FOUT) on docs.temporal.io


jsundai:
### Temporal Affiliation

- [x] I am a Temporal employee

### Page URL

docs.temporal.io

### Technical Issue

On first visit to docs site, text briefly renders in a system fallback font, then snaps to correct fonts. Noticeable on headings and body copy. It's most visible on homepage but it's a sitewide issue.

Duncanma:
@jsundai I believe I've fixed this... can you try to repro?

jsundai:
It's fixed!! 🥳 Closing this out!

PR #4938, opened 2026-07-21
Attempting to solve our font-fallback issue

Duncanma:
## What does this PR do?

Fixes the flash of unstyled/wrong font face on docs.temporal.io (FOUT) and cleans up the font-loading pipeline behind it.

**Root cause:** all self-hosted fonts used `font-display: swap` with no metric-matched fallback, so every page painted in the browser's default font and then visibly snapped to Aeonik/Poppins once each downloaded — worse than necessary because two of the three font families being shipped to every visitor weren't used anywhere.

**Removed entirely unused fonts** (confirmed via full-repo search, including generated `ai-cookbook/` content, and a `git blame` showing they were replaced by the [Nov 2023 IA redesign](https://github.com/temporalio/documentation/commit/adab4d6b1)):
- Poppins — all 18 files (~2.9MB). Its only two CSS consumers (`.badge-purple`, `.badge-blue`) were applied nowhere, and its only other reference (a Mermaid font-stack fallback) was unreachable behind Aeonik.
- The Noto Sans Mono **variable font** (1.7MB). Its only CSS consumers (`.cloud-button`, `.core-concepts-*`) were dead. The static `NotoSansMono-Regular.ttf` is kept — `plugins/og-image/render.js` reads it directly off disk for OG-card generation, unrelated to the browser bundle.

**Aeonik (the only font actually used — headings and sidebar nav on every page):**
- Converted from WOFF to WOFF2 (~10% smaller).
- Added a metric-matched `Aeonik Fallback` `@font-face` (`local('Arial')` + computed `size-adjust`/`ascent-override`/`descent-override`), applied alongside a generic `sans-serif` at every usage site, so the fallback-to-Aeonik swap no longer shifts layout.
- Switched `font-display` from `swap` to `optional` — no more disruptive mid-read font swaps; once cached (i.e. any SPA navigation after the first), there's zero flash at all.
- Preload the two above-the-fold weights (Light, Regular) two ways: an HTML `<link rel=preload>` (`docusaurus.config.js` `headTags`) for any environment, and an HTTP `Link` header in `vercel.json` so the browser can start fetching before it even has HTML to parse — verified this actually reaches the exact hashed file the CSS requests (not a wasted preload).

**Guardrail:** `src/constants/preloadFonts.js` is the single source of truth for those two hashed filenames. `bin/check-font-preload-hash.js` runs as a new `postbuild` step and fails the build if the real webpack output ever drifts from that constant (e.g. someone replaces the Aeonik font file), printing the exact fix for both `preloadFonts.js` and `vercel.json`.

**Dead CSS removed:** `.badge-purple`, `.badge-blue`, `.badge-header`, both `.cloud-button` rules, and the whole unused `core-concepts` carousel family (`.core-concepts*`, `.arrow`, `.slide-dot*`) — all confirmed unreferenced by any current page.

## Notes to reviewers

- `yarn build` passes clean, including the new postbuild check.
- Verified no visual regression: built `main` and this branch in parallel (isolated git worktree) as real production builds, and pixel-diffed four representative pages (home, a long docs page, the Mermaid-diagram page at `/local-activity`, and troubleshooting) — all four are byte-for-byte identical.
- Trade-off worth knowing: the preloaded font paths in `preloadFonts.js`/`vercel.json` are hardcoded to today's content hash. That's necessary because Docusaurus doesn't expose a way to preload a webpack-hashed asset without knowing the hash in advance — the postbuild check exists specifically to catch drift instead of letting it fail silently.
- `AeonikTRIAL-*.ttf` files in `src/fonts/Aeonik/` are untouched — they're unreferenced but look like trial-license assets, so left for a separate decision rather than deleted here.


lennessyy (approved):

jsundai:
Closed issue: https://github.com/temporalio/documentation/issues/4891

PR #4993, opened 2026-07-29
docs: Projects pre-release

brianmacdonald-temporal:
## What does this PR do?

Creates Cloud/Projects page for pre-release for new Projects feature.

## Notes to reviewers

A matrix for permission interactions still needs to be added.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217003251669593">EDU-6847 docs: Projects pre-release</a>


docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+  - temporal cloud
+  - namespaces
+tags:
+  - Projects
+  - Namespaces
+  - Nexus
+  - Nexus Endpoints
+---
+
+Projects are an organizational container within a Temporal Cloud account that allows you to group related resources into a single operational unit. With Projects, you can organize Namespaces, Nexus endpoints, and Connectivity Rules around how your organization operates: by team, environment, service, business unit, or for usage attribution purposes. Projects allow you to define ownership boundaries for teams, business units, and services within a single Temporal account. This allows for more detailed usage information in billing, and offers greater control as your usage scales.
```
LutaoX: >  or for usage attribution purposes.
> This allows for more detailed usage information in billing, and offers greater control as your usage scales.

Suggest to delete the part mentioning usage attribution or billing. This may mislead users to use Projects for usage reporting and attribution. Our goal is to double billing from resource hierarchy. We have a different initiative to enable billing group that can aggregate projects, which is not available yet. 
https://app.notion.com/p/temporalio/1-pager-Billing-Group-Resource-Management-3508fc56773880aaa2fcd9e4d4243a1f?source=copy_link

LutaoX:
```suggestion
Projects are an organizational layer within a Temporal Cloud Account that groups related resources into a single operational boundary. With Projects, you can organize Namespaces, Nexus endpoints, and Connectivity Rules around how your organization operates: by team, environment, service, or business unit.

Projects make it easier to delegate ownership, manage access at scale, and organize resources without creating additional Cloud Accounts. As your organization grows, Projects help align your Temporal Cloud resources with your organizational structure while reducing operational overhead.

Projects are for organization and authorization. They don't change how Workflows execute inside a Namespace.
```

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+Every Temporal Cloud Account includes a Default Project. Existing Accounts receive a Default Project when Projects are enabled, and new Accounts are created with one automatically. The Default Project is not a special type of Project. It behaves like any other Project for managing resources. The only difference between the Default Project and any Projects you create is that the Default Project is created for you.
+
+For existing Accounts, all existing resources are automatically placed in the Default Project. All permission resources remain unchanged. For new Accounts, the Default Project is created automatically. New resources are automatically placed in the Default Project unless you specify otherwise.
+
+If you don't need the organizational boundaries offered by Projects, continue to use the Default Project for all your resources. You can always create a new Project later if you need it.
+
+All Projects are identified by immutable IDs, and by display names, which can be edited.
+
+## Limitations
```
LutaoX: we should probably move this section to system limits page? https://docs.temporal.io/cloud/limits
suggest to add a `Projects level` between 'Accounts level` and `namespace level` on the system limit page

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+The total number of Projects in an Account is limited by an adjustable quota, similar to Namespace quotas.
+
+By default, you can create up to 10 Projects per Account. You can increase this number up to 100 Projects by request.
+
+The total number of each Project-level resource is limited within each Project. Projects inherit the current Account-level limits. A Project cannot be set to have higher limits than the Account that contains it.
+
+| Resource | Default | Maximum |
```
LutaoX: I would remove the Maximum column, because this can be misleading, especially for Namespaces, which we don't have a hard upper bond limit. We just need to clarify that Projects have their own defaults; account limits still apply as a ceiling.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+All Projects are identified by immutable IDs, and by display names, which can be edited.
+
+## Limitations
+
+The total number of Projects in an Account is limited by an adjustable quota, similar to Namespace quotas.
+
+By default, you can create up to 10 Projects per Account. You can increase this number up to 100 Projects by request.
+
+The total number of each Project-level resource is limited within each Project. Projects inherit the current Account-level limits. A Project cannot be set to have higher limits than the Account that contains it.
```
LutaoX:
```suggestion
Each Project has configurable resource limits. New Projects start with the default limits shown below. These limits can be adjusted, but cannot exceed the corresponding Account-level limits.
```

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+| Resource | Default | Maximum | 
+| --- | --- | --- |
+| Namespaces | 10 | 5000 |
+| Nexus Endpoints | 10 | 100  |
+| Connectivity Rules | 5 | 25 |
+| Project-scoped Service Accounts | 10 | 1000 |
```
LutaoX: Rather than a table, we can follow the format in this page as we move to this page: https://docs.temporal.io/cloud/limits#account-level

**Projects**
- Scope: Account
- Default limit: 5 projects
- How to increase: Contact support

**Project Level**
The following limits apply at the Project level for any new Project created. The Default Project uses the Account-level resource limits rather than the per-Project defaults shown above. For example, the Default Project inherits the Account's Namespace quota and Account-level limits for Service Accounts, Nexus Endpoints, and Connectivity Rules.

Namespaces & Project-scoped Service Accounts
Scope: Project 
Default: 10
How to increase: Contact support

Nexus Endpoints
Scope: Project 
Default: 5
How to increase: Contact support

Connectivity Rules
Scope: Project. Each non-default Project supports up to one Public Connectivity Rule.
Default: 6
How to increase: Contact support

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+## Create a new Project
+
+Create a Project when a set of resources has a distinct owner or access model, separate from the Default Project. A good Project represents a stable ownership boundary, not just a folder.
+
+Start with one Project for a team or application with clear ownership. Add more only when needed.
+
+In the Temporal UI, click **Projects** at the top of the left navigation bar, under **Your Account**. You'll see a list of existing Projects associated with your Account. The first time you use Projects, you'll only see the Default Project in this list.
+
+Click the **Create Project** button to create a new Project. On the **Create Project** page that appears, give the Project a name (limited to 39 characters) and an optional description. Click **Create** and you'll be taken to the page for your new Project. You cannot currently migrate resources from existing Projects to a new Project. Instead, from this page click **Create Namespace** to add a new Namespace to this Project. To add a Nexus Endpoint instead, click **Nexus Endpoints** in the left navigation and then click the **Create Nexus Endpoint** button.
```
LutaoX: Click the **Create Project** button to create a new Project. On the **Create Project** page that appears, give the Project a name (limited to 64 characters) and an optional description(limited to 255 characters). Click **Create** and you'll be taken to the page for your new Project. You cannot currently migrate resources from existing Projects to a new Project yet in Pre-release. Instead, from this page click **Create Namespace** to add a new Namespace to this Project. To add a Nexus Endpoint instead, click **Nexus Endpoints** in the left navigation and then click the **Create Nexus Endpoint** button.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+
+## Add Resources to a Project
+
+To add a new Namespace to this Project, click **Namespaces** in the left navigation, and then click **Create Namespace** to add a new Namespace.
+
+To add a new Nexus Endpoint to this Project, click **Nexus Endpoints** in the left navigation, and then click **Create Nexus Endpoint** to add a new Nexus Endpoint.
+
```
LutaoX, line 113: We should also introduce `Project-scoped service accounts` here and add the specifics in this service account page: https://docs.temporal.io/cloud/manage-access/service-accounts#create-a-namespace-scoped-service-account

e.g. 
Project-scoped Service Accounts can be created and managed within a Project. Unlike Account-scoped Service Accounts, they can only access resources within their Project, making them useful for workloads that should not have Account-wide permissions. More see [create a Project-scoped-service-account].

Raw Content for create a project scoped service account: 
- To create a Project-scoped service account  to this Project, Go to **project identities** in the left navigation, and then click `Service Accounts` tab to create Project-scoped service account. 

Project-scoped Service Accounts are not created automatically with Project creation. Create them only when needed.
Unlike an Account-scoped Service Account, a Project-scoped Service Account cannot access resources outside its Project. 
Use a Project-scoped Service Account when automation only needs access to one team's resources.
If automation needs to work across multiple Projects, use an Account-scoped Service Account instead.
A Project can contain multiple Project-scoped Service Accounts. Each Project-scoped Service Account belongs to exactly one Project.
A Project-scoped Service Account cannot be moved to another Project after it is created.
Project-scoped Service Accounts count toward the Project's Service Account limit.
A Project cannot be deleted while it contains Project-scoped Service Accounts.
API keys created for the Service Account inherit only that Service Account's permissions. Deleting a Project-scoped Service Account also deletes its API keys.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+## Project roles
+
+The following six roles are available at the Project level for users, groups, and service accounts:
+
+- **Project Admin:** Complete access to the Project resource itself and all resources in it.
+- **Project Write:** Write access to all the Namespaces and other resources in a Project, but cannot add other users.
+- **Project Read:** Read-only access to all the resources in a Project, including viewing the actual contents of a Workflow.
+- **Project Contribute:** Can create new Namespaces, and list all Project resources, but without access to Workflow contents.
+- **Project List:** List-only access to metadata for all the resources in a Project, but without access to Workflow contents.
+- **Project Developer:** Can creating Namespaces and edit other resources in a Project, but without access to see all existing Namespaces in the Project.
```
LutaoX:
```suggestion
```

LutaoX: suggest to delete this , with this role, the total is 7. moving description about this later in Project Developer section

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+- **Project Member:** Access to some specific resource in the Project, without granting any other Project-level access.
```
LutaoX:
```suggestion
**Project Member:** Minimal Project-level access; used when a user already has explicit Namespace-level access and only needs to be represented at the Project level
```

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+:::warning Project Developer
+
+The Project Developer role exists for backward compatibility with the Account Developer role. Assigning this role is not recommended. Assign Project Contribute or Project Write instead.
```
LutaoX:
```suggestion
Project Developer is a compatibility role that is automatically inherited by users with the Account Developer role. It allows them to create Namespaces (and manage them) without automatically granting access to all existing Namespaces in the Project. This role cannot be assigned directly. For new role assignments, assign Project Write or Project Contribute as appropriate.
```

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+The drop-down list on the left lists the available Project Roles. To add a user to a Role, select the Role from the drop-down, and then type the user's name or email in the search box. The user will be assigned to the Role and their details will appear in the list below. Click **Save** to save your changes, or **Cancel** to discard them.
+
+To modify an existing user's role, select the edit button in the list next to their current permission level. You can then update the user's permissions in this Project. Click the trash-can icon to remove Project permissions from that user.
+
+The Service Account and Groups tabs work the same way. 
+
+## Project roles
+
+The following six roles are available at the Project level for users, groups, and service accounts:
```
LutaoX: can we add a placeholder to refer to the permission reference with details? 
https://docs.temporal.io/cloud/manage-access/permissions-reference

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+The Project Developer role exists for backward compatibility with the Account Developer role. Assigning this role is not recommended. Assign Project Contribute or Project Write instead.
+
+:::
+
+Similar to Account-level roles, only a single effective Project role applies to a principal, determined by the highest-precedence role granted.
+
+A principal's Project-level roles inherit from their Account-level roles, but explicit Project grants can supersede inherited roles. For example, an Account Developer could be explicitly granted Project Admin to a particular Project, which would supersede their inherited Project Developer role.
+
+Project-level permissions are additive. Addition of Project level roles do not narrow user permissions gained via Account-level or Namespace-level permissions.
```
LutaoX:
```suggestion
A principal automatically inherits a Project-level access from their Account-level role. For example, a Global Admin automatically inherits the Project Admin access for every Project in the Account. 

Project-level permissions are additive. Granting a Project-level role does not reduce permissions inherited from Account-level or Namespace-level roles. A principal's effective permissions are the union of all applicable Account-, Project-, and Namespace-level role assignments.
```

LutaoX (changes requested):

docs/cloud/projects.mdx
brianmacdonald-temporal: Done. /cloud/limits updated.

docs/cloud/projects.mdx
brianmacdonald-temporal: Removed the whole table and moved content to Limits, as discussed.

docs/cloud/projects.mdx
brianmacdonald-temporal: Added to Limits.

docs/cloud/projects.mdx
brianmacdonald-temporal: I'm trying not to refer to a release stage or make forward-looking statements. The doc describes the functionality as it currently exists.
I realize that we're going to add the ability to migrate resources shortly, which is why I used "currently" in the original text. Once that feature is added, we'll update this text.

docs/cloud/projects.mdx
brianmacdonald-temporal: Added.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+Projects are an organizational layer within a Temporal Cloud Account that groups related resources into a single operational boundary. With Projects, you can organize Namespaces, Nexus endpoints, and Connectivity Rules around how your organization operates: by team, environment, service, or business unit.
+
+Projects make it easier to delegate ownership, manage access at scale, and organize resources without creating additional Cloud Accounts. As your organization grows, Projects help align your Temporal Cloud resources with your organizational structure while reducing operational overhead.
+
+Projects are for organization and authorization. They don't change how Workflows execute inside a Namespace.
+
+## Default Project and Project names
+
+Every Temporal Cloud Account includes a Default Project. Existing Accounts receive a Default Project when Projects are enabled, and new Accounts are created with one automatically. The Default Project is not a special type of Project. It behaves like any other Project for managing resources. The only difference between the Default Project and any Projects you create is that the Default Project is created for you.
```
LutaoX:
```suggestion
Every Temporal Cloud Account includes a Default Project. Existing Accounts receive a Default Project when Projects are enabled, and new Accounts are created with one automatically. The Default Project is not a special type of Project. It behaves like any other Project for managing resources. The main difference between the Default Project and any Projects you create is that the Default Project is automatically created for you as a starting point and provides backward-compatible behavior for existing resources and operations that do not specify a Project. Otherwise, it supports the same resource-management capabilities as other Projects.
```

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+For existing Accounts, all existing resources are automatically placed in the Default Project. All permission resources remain unchanged. For new Accounts, the Default Project is created automatically. New resources are automatically placed in the Default Project unless you specify otherwise.
```
LutaoX:
```suggestion

- For existing Accounts, all existing resources are automatically placed in the Default Project. Existing role assignments and effective permissions remain unchanged. 

- For new Accounts, the Default Project is created automatically. New resources are automatically placed in the Default Project unless you specify otherwise.

File: docs/cloud/projects.mdx [hunk 15]
```diff
@@ -0,0 +1,123 @@
...
+
+If you don't need the organizational boundaries offered by Projects, continue to use the Default Project for all your resources. You can always create a new Project later if you need it.
+
+All Projects are identified by immutable IDs, and by display names, which can be edited.
+
+Namespaces cannot be moved between Projects currently, meaning all existing Namespaces will exist in the Default Project. You can create a new Project and place a new Namespace in that Project.
+
+For limitations on the number of Projects per Account and number of resources per Project, see [Project-level limitations](/cloud/limits#project-level).
+
+## Create a new Project
```
LutaoX, line 59: Can we also have the separate tab to show instructions for Web UI and Cloud Ops API? similar to what we have for custom roles? 
https://docs.temporal.io/cloud/manage-access/custom-roles#create-custom-roles
would be great to also include UI screenshot here example:

Image
Image

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+Create a Project when a set of resources has a distinct owner or access model, separate from the Default Project. A good Project represents a stable ownership boundary, not just a folder.
+
+Start with one Project for a team or application with clear ownership. Add more only when needed.
+
+In the Temporal UI, click **Projects** at the top of the left navigation bar, under **Your Account**. You'll see a list of existing Projects associated with your Account. The first time you use Projects, you'll only see the Default Project in this list.
+
+Click the **Create Project** button to create a new Project. On the **Create Project** page that appears, give the Project a name (limited to 64 characters) and an optional description (limited to 255 characters). Click **Create** and you'll be taken to the page for your new Project. You cannot currently migrate resources from existing Projects to a new Project. Instead, from this page click **Create Namespace** to add a new Namespace to this Project. To add a Nexus Endpoint instead, click **Nexus Endpoints** in the left navigation and then click the **Create Nexus Endpoint** button.
```
LutaoX:
```suggestion
To create a Project using the Temporal Cloud UI:

1. Select **Projects** at the top of the left navigation bar. 
2. Click the **Create Project** button to create a new Project.
3. On the **Create Project** page that appears, give the Project a name (limited to 64 characters) and an optional description (limited to 255 characters).
4. Click **Create**.

After creating the Project, add new Namespaces and other supported resources within it. You cannot yet migrate resources from existing Projects to a new Project.

File: docs/cloud/projects.mdx [hunk 17]
```diff
@@ -0,0 +1,123 @@
...
+
+
+## Add Resources to a Project {/* #add-resources-to-a-project */}
+
+To add a new Namespace to this Project, click **Namespaces** in the left navigation, and then click **Create Namespace** to add a new Namespace.
+
+To add a new Nexus Endpoint to this Project, click **Nexus Endpoints** in the left navigation, and then click **Create Nexus Endpoint** to add a new Nexus Endpoint.
```
LutaoX:
```suggestion
After creating a Project, you can create Project-scoped resources such as Namespaces, Nexus Endpoints, Connectivity Rules within it.

In Temporal Cloud UI:
1. Open the Project. 
2. Select **Namespaces** (or **Nexus Endpoints**) in the left navigation. 
3. Click **Create Namespace** (or **Nexus Endpoints**) to add to the project. 

**Connectivity rules** cannot currently be created via the UI. Connectivity rules and the Namespace they attach to must reside in the same Project.

File: docs/cloud/projects.mdx [hunk 18]
```diff
@@ -0,0 +1,123 @@
...
+
+To add users, groups, and Service Accounts to the Project, click the **Project Identities** tab in the left navigation. The Project Identities page has three tabs, for Users, Service Accounts, and Groups. Click the **Manage Identities** button to add or remove users.
+
+The drop-down list on the left lists the available Project Roles. To add a user to a Role, select the Role from the drop-down, and then type the user's name or email in the search box. The user will be assigned to the Role and their details will appear in the list below. Click **Save** to save your changes, or **Cancel** to discard them.
+
+To modify an existing user's role, select the edit button in the list next to their current permission level. You can then update the user's permissions in this Project. Click the trash-can icon to remove Project permissions from that user.
+
+The Groups tab works the same way as the Users tab.
```
LutaoX:
```suggestion
## Manage Project Access
Users, groups, and Account-scoped Service Accounts can be granted Project-level roles. Each [Project role](#Project-roles) applies to the Project and, depending on the role, may grant inherited access to resources within it.

To manage Project access in the Temporal Cloud UI:

1. Open the Project.
2. Click the **Project Identities** tab in the left navigation. The Project Identities page has three tabs, for Users, Service Accounts, and Groups.
3. Click the **Manage Identities** button to add (or remove) users, groups or service accounts, and assign desired Project Roles from the drop-down list. You can also modify or remove an existing principal's role in the same view.

File: docs/cloud/projects.mdx [hunk 19]
```diff
@@ -0,0 +1,123 @@
...
+To add a new Nexus Endpoint to this Project, click **Nexus Endpoints** in the left navigation, and then click **Create Nexus Endpoint** to add a new Nexus Endpoint.
+
+To add users, groups, and Service Accounts to the Project, click the **Project Identities** tab in the left navigation. The Project Identities page has three tabs, for Users, Service Accounts, and Groups. Click the **Manage Identities** button to add or remove users.
+
+The drop-down list on the left lists the available Project Roles. To add a user to a Role, select the Role from the drop-down, and then type the user's name or email in the search box. The user will be assigned to the Role and their details will appear in the list below. Click **Save** to save your changes, or **Cancel** to discard them.
+
+To modify an existing user's role, select the edit button in the list next to their current permission level. You can then update the user's permissions in this Project. Click the trash-can icon to remove Project permissions from that user.
+
+The Groups tab works the same way as the Users tab. 
+
```
LutaoX, line 216: ### Create a Project-scoped Service Account

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,123 @@
...
+To create a Project-scoped service account in this Project, click **Project Identities** in the left navigation. Then click the  **Service Accounts**  tab to create Project-scoped service account.
+
+Project-scoped Service Accounts are not created automatically with Project creation. Create them only when needed. Unlike an Account-scoped Service Account, a Project-scoped Service Account cannot access resources outside its Project.
+
+Use a Project-scoped Service Account when automation only needs access to one team's resources. If automation needs to work across multiple Projects, use an Account-scoped Service Account instead.
```
LutaoX:
```suggestion
Similar to [Namespace-scoped Service Accounts](/cloud/manage-access/service-accounts#scoped), a Project-scoped Service Account is an identity bound (or scoped) to a single Project. Use it for workers, CI/CD pipelines, Terraform, or operational automation that should not access resources outside that Project. Project-scoped Service Accounts are not created automatically with Project creation. Create them only when needed. 

To create a Project-scoped service account in this Project, click **Project Identities** in the left navigation. Then click the  **Service Accounts**  tab to create Project-scoped service account.

After creating the Service Account, you can optionally create one or more API keys for it. Each API key inherits only the permissions of its associated Service Account. If the Service Account is deleted, all of its API keys are deleted automatically.

File: docs/cloud/projects.mdx [hunk 21]
```diff
@@ -0,0 +1,123 @@
...
+
+A Project can contain multiple Project-scoped Service Accounts. Each Project-scoped Service Account belongs to exactly one Project. 
+
+A Project-scoped Service Account cannot be moved to another Project after it is created.
+
+Project-scoped Service Accounts count toward the Project's Service Account limit.
+
+A Project cannot be deleted while it contains Project-scoped Service Accounts.
+
+API keys created for the Service Account inherit only that Service Account's permissions. Deleting a Project-scoped Service Account also deletes its API keys.
```
LutaoX:
```suggestion
A Project-scope Service Account:

- Can only bound (or scope to) exactly one Project. Each project can contain multiple Project-scoped Service Accounts. 
- Cannot be moved to another Project after it is created.
- Counts toward the Project's Service Account limit.
- Must be deleted before the Project can be deleted

Unlike an [Account-scoped Service Account](/cloud/manage-access/service-accounts), a Project-scoped Service Account cannot access resources outside its Project. For automation that needs access across multiple Projects, use an Account-scoped Service Account and grant it the appropriate Project roles.

File: docs/cloud/projects.mdx [hunk 22]
```diff
@@ -0,0 +1,123 @@
...
+
+To modify an existing Project, click **Project Settings** in the left navigation. The Project Settings page appears, with the General tab selected by default. Here, you can change the name or description of your Project. You can also access your Project's unique ID. You can copy the ID, but you can't change it.
+
+## Delete a Project
+
+To delete a Project, click **Project Settings** in the left navigation and then click the **Delete Project** tab. Click the **Delete** button to permanently delete the Project.
+
+Use the **Deletion Prevention** slider to set the Deletion Protection flag for this Project. Once set, the Project can't be deleted unless the flag is deliberately disabled.
+
+A Project must be empty, meaning it contains zero Namespaces, to be deleted.
```
LutaoX: A Project must be empty, meaning it contains zero Namespaces, Nexus Endpoints, Connectivity Rules or Project-scoped Service Accounts, to be deleted.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,123 @@
...
+  - Nexus Endpoints
+---
+
+import { ReleaseNoteHeader } from '@site/src/components';
+
+<ReleaseNoteHeader
+  featureName="projects"
+/>
+
+Projects are an organizational layer within a Temporal Cloud Account that groups related resources into a single operational boundary. With Projects, you can organize Namespaces, Nexus endpoints, and Connectivity Rules around how your organization operates: by team, environment, service, or business unit.
```
LutaoX:
```suggestion
Projects are an organizational layer within a Temporal Cloud Account that groups related resources into a single operational boundary. With Projects, you can organize [Namespaces](/cloud/namespaces), [Nexus endpoints](/nexus/endpoints), and [Connectivity Rules](/cloud/connectivity#connectivity-rules) around how your organization operates: by team, environment, service, or business unit.

File: docs/cloud/manage-access/permissions-reference.mdx
Comment (LutaoX) [inline docs/cloud/manage-access/permissions-reference.mdx:377]: Not sure if we should directly change here. 
Need to add similar condition to Projects e.g. 

Requires Namespace (or Project) Admin permission on the target Namespace (or Project). Any role can manage Namespace-scoped service accounts (or Project-scoped service accounts) if
they hold Namespace (or Project) Admin on that Namespace (or Project).

File: docs/cloud/projects.mdx (hunk 15 above)
Comment (brianmacdonald-temporal) [reply docs/cloud/projects.mdx:59]: Added.

File: docs/cloud/projects.mdx (hunk 22 above)
Comment (brianmacdonald-temporal) [reply docs/cloud/projects.mdx]: fixed.

File: docs/cloud/manage-access/permissions-reference.mdx
Comment (brianmacdonald-temporal) [reply docs/cloud/manage-access/permissions-reference.mdx:377]: Added a new row to the table for Projects, and a new footnote like the Namespace one, but for Projects. Does that correct the issue?

**Comment by lennessyy** [comment]
I think these pages will need updating as well. These pages cover Temporal Cloud's access model and core concepts are most users primary entry points to learn how access is governed in Cloud. If they don't mention projects users may be surprised to land on an isolated projects page:

- [Account access](https://temporal-documentation-git-projects-feature-docs.preview.thundergun.io/cloud/manage-access/)
- [Roles and permissions](https://temporal-documentation-git-projects-feature-docs.preview.thundergun.io/cloud/manage-access/roles-and-permissions)

In addition, given this is a primarily and IAM feature, I think this page belongs in the cloud/manage-access/ folder. It's worth doing a pass over every page in this folder to check if any content needs updating

File: docs/cloud/projects.mdx [hunk 24]
```diff
@@ -0,0 +1,368 @@
...
+
+Projects are for organization and authorization. They don't change how Workflows execute inside a Namespace.
+
+## Default Project and Project names
+
+Every Temporal Cloud Account includes a Default Project. Existing Accounts receive a Default Project when Projects are enabled, and new Accounts are created with one automatically. The Default Project is not a special type of Project. It behaves like any other Project for managing resources. The main difference between the Default Project and any Projects you create is that the Default Project is automatically created for you as a starting point and provides backward-compatible behavior for existing resources and operations that do not specify a Project. Otherwise, it supports the same resource-management capabilities as other Projects.
+
+
+- For existing Accounts, all existing resources are automatically placed in the Default Project. Existing role assignments and effective permissions remain unchanged.
```
lennessyy: When we say new accounts vs existing accounts, what are these relative to? I am assuming we mean relative to when the account enables the Projects feature? 

If that's the case, I think it's better to dedicate a section "Enabling Projects on your account"?  This way, you can more freely describe how Projects work in it's ideal, already enabled state. And when the feature moves into GA and becomes enabled by default, it's a clean edit to just remove that section. 

Otherwise this may get confusing as this feature moves into release stages, and cleaning up the docs might also become more work.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,368 @@
...
+  }
+}
+```
+
+In this call, `description` is a string description of the Project, `displayName` is the Project name to be displayed in the UI, and `enableDeleteProtection` is a Boolean that indicates whether the Project can be deleted.
+
+  </TabItem>
+</Tabs>
+
+After creating the Project, add new Namespaces and other supported resources within it. You cannot yet migrate resources from existing Projects to a new Project.
```
Duncanma: would it be simpler to say "You cannot migration resources between projects." ? (because it's not just that you can't move to a new project, you can't move them at all)

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,368 @@
...
+          "allowedCloudNamespacePolicySpec": {
+            "namespaceId": "checkout-prod.a1b2c3"
+          }
+        }
+      ]
+    }
+  }'
+```
+
+This call is the same as the standard process for [creating a Nexus Endpoint](/nexus/registry#view-and-manage-nexus-endpoints), with the addition of the `projectId` for the Project in which you're creating the Nexus Endpoint.
```
Duncanma: I get that all 3 of these are the same basic idea, but the repeated "This call is the same as the standard process" sounds awkward to me

docs/cloud/projects.mdx
brianmacdonald-temporal: That's a good idea; thanks.

docs/cloud/projects.mdx
brianmacdonald-temporal: Good call; thanks. (Although migrating resources is on the list for Public Preview.)

docs/cloud/projects.mdx
brianmacdonald-temporal: DRY; that makes sense. Refactored.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,367 @@
...
+
+
+## Default Project and Project names
+
+Every Temporal Cloud Account includes a Default Project. The Default Project is not a special type of Project. It behaves like any other Project for managing resources. The main difference between the Default Project and any Projects you create is that the Default Project is automatically created for you as a starting point and provides backward-compatible behavior for existing resources and operations that do not specify a Project. Otherwise, it supports the same resource-management capabilities as other Projects.
+
+If you don't need the organizational boundaries offered by Projects, continue to use the Default Project for all your resources. You can always create a new Project later if you need it.
+
+All Projects are identified by immutable IDs, and by display names, which can be edited.
```
LutaoX:
```suggestion
Every Project has a unique, immutable `Project ID` and an editable `display name`. The Project ID is used by the Cloud Ops API, CLI, and Terraform to identify the Project when creating and managing Project-scoped resources.
```
wanted to emphasize the use case for `Project ID` here.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,367 @@
...
+Projects are for organization and authorization. They don't change how Workflows execute inside a Namespace.
+
+## Enable Projects on your Account
+
+While Projects is in pre-release, it is not enabled by default. [Contact Support](/cloud/support#support-ticket) to request enabling Projects for your account.
+
+Once Projects are enabled, existing Accounts receive a Default Project, and new Accounts are created with one automatically. For existing Accounts, all existing resources are automatically placed in the Default Project. Existing role assignments and effective permissions remain unchanged. For new Accounts, the Default Project is created automatically. New resources are automatically placed in the Default Project unless you specify otherwise.
+
+
+## Default Project and Project names
```
LutaoX: Should we say `Default Project and Project ID` instead? Project ID is something net new we are introducing.

docs/cloud/manage-access/permissions-reference.mdx
```diff
@@ -270,6 +273,65 @@ permissions. These permissions are configured per Namespace per user.
 | Initiate Namespace failover                                   |      |       |        ✔        |
 | Convert single-region Namespace to multi-region               |      |       |        ✔        | */}
 
+## Project-level permissions
+
+Project-level permissions are granted to users, groups, and service accounts by assigning them a Project-level role.
```
LutaoX:
```suggestion
[Project-level permissions](/cloud/manage-access/permissions-reference#project-level-permissions) are granted to users, groups, and service accounts by assigning them a Project-level role.
```
adding hyperlink to permissions references so users can learn more details.

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,367 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import { ReleaseNoteHeader } from '@site/src/components';
+
+<ReleaseNoteHeader featureName="projects">
+Features in pre-release are experimental and may be subject to backwards-incompatible changes.
+
+Projects currently support the Temporal UI and Temporal Cloud Ops API.
```
bechols: This sentence sounds strange. 

1. You can see the object (Projects) via different interfaces (UI, API). Current ordering feels backwards.
2. What does the user do with this info? Is this implying Projects aren't visible in the CLI?

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,367 @@
...
+
+Projects are an organizational layer within a Temporal Cloud Account that groups related resources into a single operational boundary. With Projects, you can organize [Namespaces](/cloud/namespaces), [Nexus endpoints](/nexus/endpoints), and [Connectivity Rules](/cloud/connectivity#connectivity-rules) around how your organization operates: by team, environment, service, or business unit.
+
+Projects make it easier to delegate ownership, manage access at scale, and organize resources without creating additional Cloud Accounts. As your organization grows, Projects help align your Temporal Cloud resources with your organizational structure while reducing operational overhead.
+
+Projects are for organization and authorization. They don't change how Workflows execute inside a Namespace.
+
+## Enable Projects on your Account
+
+While Projects is in pre-release, it is not enabled by default. [Contact Support](/cloud/support#support-ticket) to request enabling Projects for your account.
```
bechols: Should be consistent if we're treating Projects as singular or plural. Plural seems more natural overall.

docs/cloud/projects.mdx
bechols: +1

bechols: Alternatively, have one section that explains default projects, and a separate section explaining ID usage vs name

docs/cloud/projects.mdx
```diff
@@ -0,0 +1,367 @@
...
+
+Projects currently support the Temporal UI and Temporal Cloud Ops API.
+</ReleaseNoteHeader>
+
+Projects are an organizational layer within a Temporal Cloud Account that groups related resources into a single operational boundary. With Projects, you can organize [Namespaces](/cloud/namespaces), [Nexus endpoints](/nexus/endpoints), and [Connectivity Rules](/cloud/connectivity#connectivity-rules) around how your organization operates: by team, environment, service, or business unit.
+
+Projects make it easier to delegate ownership, manage access at scale, and organize resources without creating additional Cloud Accounts. As your organization grows, Projects help align your Temporal Cloud resources with your organizational structure while reducing operational overhead.
+
+Projects are for organization and authorization. They don't change how Workflows execute inside a Namespace.
+
```
LutaoX, line 31: can we include this diagram here? 

Image

LutaoX (changes requested):

docs/cloud/projects.mdx
Duncanma: added

docs/cloud/projects.mdx
Duncanma: I dropped the second line for now, I don't think it is critical and the meaning/intent is unclear

docs/cloud/projects.mdx
Duncanma: Switched this to plural... it is somewhat questionable, since the feature name has a plural, so it's a singular entity as is... but we are inconsistent about this across the docs and while I think treating it as a singular is technically correct, I also agree it sounds more natural to treat it like a plural :) 

2 inconsistent examples in the docs I found in 10 seconds 

"Temporal Schedules is a feature that allows you to "schedule" Temporal Workflows at specified times or intervals, adjusting for peak use."

"Serverless Workers are a good fit for many workloads. They offer several advantages compared to long-lived Workers on dedicated compute."

Duncanma (approved):

PR #504, opened 2021-07-05
refactor swizzled DocItem component as per latest docusaurus version

covalentbond:
### What does this PR do?
Fixes the [banner issue](https://discord.com/channels/398180168688074762/567414046073159701) while updating the Docusaurus version to `v2.0.0-beta.3` mainly causing because of [facebook/docusaurus#5052](https://github.com/facebook/docusaurus/pull/5052)

### Notes to reviewers
Banner does not get displayed now, while updating the Docusaurus version to `v2.0.0-beta-3`
Completes #502.

@flossypurse kindly review it :)

PR #5024, opened 2026-08-03
docs(typescript): snipsync hand-copied code blocks; drop unpublished GCP OTel content

lennessyy:
Three independent changes, one commit each.

## 1. Snipsync twelve hand-copied blocks (`c6bc8f8`)

The TypeScript SDK guide has **217 inline TS/JS code blocks and none of them were snipsynced**, so nothing catches drift from the samples they were copied from. This wraps the twelve whose content is already byte-identical to a marker in `samples-typescript`:

- `versioning.mdx` — the four patching blocks (`patching-api`)
- `set-up.mdx` — the greet Activity and example Workflow (`hello-world`)
- `standalone-activities.mdx` — six blocks (`standalone-activity`)

**Depends on [temporalio/samples-typescript#504](https://github.com/temporalio/samples-typescript/pull/504)**, which adds the six `standalone-activity` markers. No strict merge order: snipsync only splices IDs it can resolve, so until #504 lands those blocks stay as-is rather than emptying. The other six markers already exist on `main`.

Rather than hand-checking the output format, each block was verified against snipsync's own output, reproduced by driving its `deindentByCommonPrefix` over the real sample files. All twelve match byte-for-byte, so the next `yarn snipsync` is a no-op here.

Two incidental fixes on `standalone-activities.mdx`: it carried hand-written source-link lines that snipsync generates itself, and its fences said `typescript` where snipsync emits `ts`.

## 2. Drop unpublished GCP OpenTelemetry content (`8791a44`)

The Cloud Run pages documented a GCP OpenTelemetry plugin that does not exist. `sdk-go/contrib/gcp` and sdk-python's `temporalio/contrib/gcp` both 404 on `main`, and `gcp_open_telemetry/` is absent from `samples-python` — so four snipsync IDs on these pages resolve nowhere and the code shown cannot be run. The samples are weeks out.

Removes the plugin sections from the Go and Python Cloud Run pages, including the shutdown-flush subsection and the 70-line collector config. In their place, the generic `## Add observability` section the TypeScript page already carries — a Cloud Run Worker emits the same telemetry as any other Worker, which is true independent of the plugin. Keeps the three pages parallel and preserves the `#add-observability` anchor.

Also drops two stale "add the OpenTelemetry plugin to the Worker" lead-ins from the deployment guide's Python and Go tabs. Those sentences carried the tabs' only links to their SDK pages, so the links are kept and reworded to match the TypeScript tab. Python's serverless index and page keywords advertised the plugin where its Go and TypeScript siblings did not; both now match.

## 3. Link the first Temporal Client reference (`235b1596`)

Four pages in this branch mentioned a Temporal Client without linking anywhere. Each now links its first reference to that SDK's Temporal Client page: the Go and Python Cloud Run pages (in "Configure the Temporal connection"), `standalone-activities.mdx` (opening paragraph), and `set-up.mdx` (step 4, where the reader creates the Client). The other three pages in the branch don't mention a Temporal Client.

## Verification

`yarn build` passes. CI-scoped Vale is clean on all changed files (3 suggestions, all pre-existing headings). All ten new source-link URLs return 200. `yarn check-links` reports failures spread evenly across every sample directory in the repo including untouched ones — GitHub 429 rate limiting, not attributable to this change.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217133809143483">EDU-6877 docs(typescript): snipsync hand-copied code blocks; drop unpublished GCP OTel content</a>


lennessyy:
I will do this together with #5001

PR #5037, opened 2026-08-05
Worker performance pg refactor

jsundai:
## What does this PR do?
- Breaks up super long worker performance page into subpages. 
- Moves existing Worker tuning reference (/develop/worker-tuning-reference) under this category. 
- The original pg has a high bounce rate so we can monitor if the bounce rate changes and only 11% get to the performance tuning coding examples on the current page which is about halfway through the page. 

## Notes to reviewers
Scroll length: 
Original page: ~20–21 scrolls (1032 lines / ~1023 body lines)
New pages: 1-6 scrolls max 

TODO: Links needs to be checked. Double check all content is still the same and just reorganized. 

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217205738857392">EDU-6890 Worker performance pg refactor</a>


docs/develop/worker-performance/index.mdx
flippedcoder, line 1: Should this be a `index.mdx` file under the new worker-performance folder? That way it'll follow the pattern we already have.

flippedcoder (approved):

PR #5040, opened 2026-08-05
refactor temporal proxy

jsundai:
## What does this PR do?
- Breaks up the Temporal Proxy page into a hub + subpages. URL stays the same (`/production-deployment/temporal-proxy`), so no redirect needed. 

Scroll length:
Original page: ~17 scrolls (863 lines)
New pages: ~4–6 scrolls each

Subpages:
- Configure
- Encrypt payloads
- Deploy to Kubernetes

## Notes to reviewers

TODO: Links need to be checked. Double check all content is still the same and just reorganized. Transition sentences need to be added. 

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217217039089707">EDU-6893 refactor temporal proxy</a>


lennessyy (approved):
We should probably put the prelease header on each page? Otherwise looks great!

PR #5046, opened 2026-08-06
Label each troubleshooting section as a recommended alert

austonli:
Follow-up to #5029, which merged before this change was pushed.

## Problem

Reading a section on its own, nothing indicated it corresponded to an alert. Each opened with a bare `**Metric:**` line, so the threshold, `for` duration, and severity existed only in the tables on [Worker alerting](https://docs.temporal.io/best-practices/worker-alerting).

Anyone arriving from a firing alert's `runbook_url` — which is the main path onto these pages — had no way to see what condition was supposed to have triggered it.

## Change

All 20 sections across the three troubleshooting pages now open with the full alert definition:

> **Recommended alert:** `temporal_workflow_task_execution_failed` with `failure_reason=NonDeterminismError`
> Fires on any occurrence, held for 1m. Routed Critical by default. [Tune the threshold](https://docs.temporal.io/best-practices/worker-alerting#execution-failures).

Two sections cover more than one alert and say so: UNIMPLEMENTED and INTERNAL are separate alerts on the same metric, and Workflow Task schedule-to-start has two thresholds.

Each page also now opens by stating that every section on it is one of the recommended alerts, and that if you arrived from a firing alert, the section you landed on is its runbook.

Prose only. No thresholds, metric names, tags, anchors, or links changed.

## Known trade-off

This duplicates the alert definitions: once in the tables on the alerting page, once per section here. That will drift.

It's the right trade for now, because a runbook that doesn't state its own trigger condition isn't much of a runbook. But when the generated PromQL reference lands with Tiho's ruleset contribution, both should be sourced from the YAML rather than hand-maintained in two places.

## Verification

`yarn build` exits 0 with `onBrokenLinks` and `onBrokenAnchors` both set to `throw`. `check:metrics` reports 50 metrics consistent, `check:orphans` clean.

Left as a draft pending review of the approach.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217242737848754">EDU-6898 Label each troubleshooting section as a recommended alert</a>


Duncanma:
Hey @austonli, just want to make sure we do this second step (the docs team can figure out how to automate from the yaml, but if you could provide that PR or thread on the other side of this work it would be great!)

> But when the generated PromQL reference lands with Tiho's ruleset contribution, both should be sourced from the YAML rather than hand-maintained in two places.

Duncanma (approved):

PR #5054, opened 2026-08-06
Document replay-safe Workflow APIs per SDK

lennessyy:
Adds a heading per topic under the Workflow logic requirements section of each SDK's Workflow basics page, so readers can find the replay-safe alternative for what they need. Python already had this structure; PHP is unchanged.

## What changed

| SDK | Change |
| --- | --- |
| Ruby | Adds Logging, Random numbers and UUIDs, Current time, and Detecting replay. The page previously listed only what Workflows can't do, with no replacement APIs. |
| .NET | Adds the same four sections. `Workflow.NewGuid()` was undocumented and `Workflow.Random` appeared only in an analyzer-suppression note. |
| Go | Adds the same four sections above the existing construct-mapping list. Go has no built-in seeded random or UUID helper, so that section shows the Side Effect approach. |
| Java | Adds the same four sections, and points the existing "non-deterministic functions" bullet at the new one. The page named `UUID.randomUUID()` as forbidden without naming `Workflow.randomUUID()`. |
| TypeScript | Rewrites Random numbers and UUIDs to lead with the SDK's own `uuid4()` rather than the `uuid` npm package. |

## Verification

Each documented API was exercised by a Workflow per language run against `temporal server start-dev`, then the recorded history was replayed through that SDK's replayer. All five passed, replay included.

Java methods link to the `Workflow` class page rather than per-method anchors: `javadoc.io/doc/.../latest/` serves a JS shell with no resolvable anchors, and the `/static/` variant that has them requires a pinned version that would go stale.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217253263128206">EDU-6904 Document replay-safe Workflow APIs per SDK</a>


docs/develop/dotnet/workflows/basics.mdx
```diff
@@ -86,6 +86,82 @@ This means there are several things Workflows shouldn't do such as:
...
+to get a deterministic random instance seeded per Workflow Execution, and
+[`Workflow.NewGuid()`](https://dotnet.temporal.io/api/Temporalio.Workflows.Workflow.html#Temporalio_Workflows_Workflow_NewGuid)
+instead of `Guid.NewGuid()`:
+
+```csharp
+// Good - deterministic across replays
+var value = Workflow.Random.Next(1, 100);
+var uniqueId = Workflow.NewGuid();
+
+// Bad - different result on every replay
```
lennessyy: No need to include bad references

docs/develop/java/workflows/basics.mdx
```diff
@@ -265,4 +265,74 @@ The following constraints apply when writing Workflow Definitions:
...
+to get a `Random` instance seeded per Workflow Execution, and
+[`Workflow.randomUUID()`](https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/workflow/Workflow.html)
+instead of `UUID.randomUUID()`:
+
+```java
+// Good - deterministic across replays
+int value = Workflow.newRandom().nextInt(100);
+UUID uniqueId = Workflow.randomUUID();
+
+// Bad - different result on every replay
```
lennessyy: ditto

lennessyy (commented):
no need to have bad examples here. we should remove the bad example from python as well

Duncanma (approved):

PR #5068, opened 2026-08-07
Nexus V2 Documentation - don't review or merge!!

Evanthx:
Note for documentation reviewers - do not review or merge. I am sending this to some internal folks as we are discussing what we want to do.

[Microservice Development Walkthrough](https://temporal-documentation-git-nexus-v2.preview.thundergun.io/develop/java/nexus/development-walkthrough) is main part of this PR. 

Matching sample Java code: https://github.com/temporalio/samples-java/pull/793

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217291252557541">EDU-6917 Nexus V2 Documentation - don&#39;t review or merge!!</a>


docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,254 @@
+---
+id: nexus-standalone-activity
+title: Nexus Standalone Activity
```
KeithB-Temporal, line 3: Reading this gave me an interesting thought . Should this doc live under SAA rather than under Nexus. The PRD names this feature is Standalone Activity: Start from Nexus, which is actually a clearer name for this. If we leave here the name really should be something like Standalone Activity Started By Nexus

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,254 @@
...
+tags:
+  - Nexus
+  - Concepts
+---
+
+import { SdkTabs } from '@site/src/components';
+
+:::caution
+
+Activity-backed Nexus Operations are pre-release and build on [Nexus SDK V2](/nexus/sdk-v2).
```
KeithB-Temporal: build -> built

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,254 @@
...
+`TemporalOperationHandler` is marked experimental in the SDKs and may change in backwards-incompatible ways.
+
+:::
+
+A Nexus Operation can be backed by a [Standalone Activity](/standalone-activity) instead of a Workflow.
```
KeithB-Temporal: I don't like framing this as "instead of a workflow" . The framing really should be A NexusOperations can Start a Workflow or an Activity. 

Note we also need to update the https://docs.temporal.io/nexus/operations to highlight this. 
"...Nexus Operation has an operation token that can be used to re-attach to a long-running Operation backed by a Workflow **_or a Standalone Activity_**".

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,254 @@
...
+Starting the Operation starts an Activity Execution that has no parent Workflow, and the Operation completes when that Activity returns.
+
+This is the right shape when the work behind an Operation is a single step with no orchestration: call an external API, run a computation, send a notification.
+Before Activity-backed Operations, exposing an Activity through Nexus meant writing a Workflow whose only job was to call that one Activity — a wrapper with its own Event History, its own Task Queue considerations, and no value of its own.
```
KeithB-Temporal: I don't like comparing this to before state. The key reason why is that I want this documentation to target new customers that aren't using Nexus, and want to use this a new pattern, not because it is cleaner than than some old pattern that they weren't using.

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+Activity-backed Nexus Operations are pre-release and build on [Nexus SDK V2](/nexus/sdk-v2).
+`TemporalOperationHandler` is marked experimental in the SDKs and may change in backwards-incompatible ways.
+
+:::
+
+A Nexus Operation can be backed by a [Standalone Activity](/standalone-activity) instead of a Workflow.
+Starting the Operation starts an Activity Execution that has no parent Workflow, and the Operation completes when that Activity returns.
+
+This is the right shape when the work behind an Operation is a single step with no orchestration: call an external API, run a computation, send a notification.
```
KeithB-Temporal: We need to beef up this section A LOT. If I'm a new user why do I want to use the Standalone Activities. We need "real world use cases": 
1) This is a GREAT pattern for invoking an MCP Tool in a different security context to sandbox it.
2) This is a great pattern to durably invoke any other system. (IE any other service)
3) It's a great way to execute the ensure durable write to unreliable 3rd party systems ((IE to build a home-grown connector, or integrate with 3rd party systems, or unreliable internal infrastructure)

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,254 @@
...
+Before Activity-backed Operations, exposing an Activity through Nexus meant writing a Workflow whose only job was to call that one Activity — a wrapper with its own Event History, its own Task Queue considerations, and no value of its own.
+
+These compose. A Standalone Nexus Operation can be backed by a Standalone Activity, which means neither side has a Workflow.
```
KeithB-Temporal: Get rid of these compose. One common pattern is to Use a Standalone Nexus Operation to invoke a

KeithB-Temporal: Focus here on use cases and patterns it could be used for.

The signature use case: durable webhook processing without owning a queue

The scenario: A product ingests webhooks from third-party SaaS providers — Stripe payment events, GitHub push events, Twilio delivery receipts, Segment events, Salesforce change notifications. Each webhook needs to trigger a durable one-shot task: update a downstream system, kick off a notification, index into search, sync into a data warehouse.

The current shape most customers land on:
- Stateless HTTP receiver behind a load balancer accepts the webhook
- Receiver drops the payload onto SQS / Kafka / Redis
- A worker fleet consumes and processes with hand-rolled retries + dead-letter handling
- Idempotency is enforced with a Redis SETNX on the webhook ID

That's a queue, a worker, retry logic, DLQ tooling, idempotency store, dashboards — all just to run one function reliably.

With Standalone Nexus Op → Standalone Activity, it collapses to:

// Inside the webhook HTTP handler — no workflow, no proxy
sc := temporalClient.NexusServiceClient(client.NexusServiceClientOptions{
    Endpoint: "billing-events",
    Service:  "stripe",
})
handle, _ := sc.StartOperation(ctx, "process-
    client.StartNexusOperationOptions{
        OperationID:            stripeEventID,   // idempotency, dedup
        ScheduleToCloseTimeout: 24 * time.Hou
    })
// return 200 to Stripe immediately — Temporal owns delivery

On the other side, process-payment-event is a Nexus operation whose handler is a Standalone Activity. That's it. No caller workflow, no handler workflow, no queue, no DLQ table, no retry harness.

Why this specific combination wins

┌────────────────────────┬─────────────────────────────────────────────────────────────┐
│        Property        │                               Why it matters for webhooks                                │
├────────────────────────┼─────────────────────────────────────────────────────────────┤
│ No caller workflow     │ HTTP webhook receivers are inherently stateless. Wrapping every incoming request in a    │
│ (Standalone Nexus Op)  │ proxy workflow is  Cloud Actions before any real work       │
│                        │ happens.                                                                                 │
├────────────────────────┼─────────────────────────────────────────────────────────────┤
│ No handler workflow    │ Most webhook processors do one thing: transform + write. A state machine adds nothing. 1 │
│ (Standalone Activity)  │  Cloud Action per event vs. 2 is 50% cost savings at scale — and Stripe/GitHub webhook   │
│                        │ volume is high.                                             │
├────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│ Operation ID =         │ Stripe/GitHub/Twilio all send event.id you can use as the dedup key. Nexus conflict      │
│ provider event ID      │ policy gives you exactly-once semantics for free — no Redis, no dedup table.             │
├────────────────────────┼─────────────────────────────────────────────────────────────┤
│                        │ The team owning the webhook receiver is usually not the team owning the downstream       │
│ Cross-namespace /      │ processing. The bients; the growth team consumes Segment    │
│ cross-team ownership   │ events. Nexus endpoint = clean team boundary with scoped access, not full namespace      │
│                        │ write.                                                      │
├────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│                        │ Webhook processing hits unreliable downstreams (CRM APIs, analytics tools). Standalone   │
│ Circuit breaker safety │ Activity wrapping  the Nexus circuit breaker on repeated    │
│                        │ retryable errors — this is the flagship reason this pattern exists.                      │
├────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│ 200-back-fast          │ Stripe/GitHub retry aggressively if you don't 200 within seconds. Fire the Nexus op →    │
│                        │ return 200 → Tempoe.                                        │
├────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────┤
│ Long-tail retry        │ Downstream CRM is try policy handles it. No DLQ triage      │
│                        │ ritual.                                                                                  │
├────────────────────────┼─────────────────────────────────────────────────────────────┤
│ Addressable            │ Support asks "what23?" → look it up by Operation ID. Get    │
│                        │ status, retry count, last error, result. Cancelable if the event was later invalidated.  │
└────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────┘

Alternatives it beats

- SQS + Lambda worker fleet: You've built a small queueing system with hand-rolled dedup + DLQ. Nexus SIA gives you all
of that plus visibility and dedup for free.
- Wrapper workflow per webhook: Works today, but doubles Cloud Actions and adds workflow-history bloat for genuinely
stateless work.
- Direct activity from a proxy workflow (Duolingo gateway pattern): Adds a caller workflow you don't need — Standalone
Nexus Ops eliminate it.
- Raw HTTP call from receiver to downstream service: No durability. The moment your process crashes after the DB write, the event is lost.

Adjacent use cases with the same shape

The same Standalone-Nexus-Op → Standalone-Activity pattern fits any "external trigger, one durable step, cross-team"
scenario:

- BFF-triggered async user actions — user clite my report," "revoke my sessions." BFF fires a Standalone Nexus Op, returns an operation handle to the client, no proxy workflow.
- Kafka/event-consumer offload — consumer reas, commits its offset. Temporal ownsdurability from there. Event ID = Operation ID for dedup.
- CI/CD infra actions — GitHub Action or Jenknce scan," "kick off canary deploy step,""regenerate credentials." Platform team owns the handler; consumer teams get scoped endpoint access.
- Cron/scheduler-triggered platform tasks — K scheduler fires a Standalone Nexus Op →shared platform team runs the task.

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,254 @@
...
+
+  </SdkTabs.TypeScript>
+  <SdkTabs.DotNet>
+
+Code coming in next draft
+
+  </SdkTabs.DotNet>
+</SdkTabs>
+
+The Activities themselves are ordinary Activities.
```
KeithB-Temporal: I feel weird about saying the "Activities are ordinary Activities". This assumes knowledge of the Activity and I don't like characterizing an Activity as ordinary, that is one of our most powerful primatives. 

I think the thing to highlight is the same Activities can be invoked from either traditional workfllows or by Nexus invoked SAA

docs/encyclopedia/nexus/nexus-sdk-v2.mdx
```diff
@@ -0,0 +1,502 @@
...
+**Async backings — at most one per Operation invocation.** These determine what the Operation *is*, and their result is delivered to the caller through the Nexus completion callback when the underlying Execution finishes.
+
+- `client.startWorkflow(...)` — the Operation completes when the Workflow returns
+- `client.startWorkflowUpdate(...)` — the Operation completes when the Update completes
+- `client.startActivity(...)` — the Operation completes when the Activity returns; see [Nexus Standalone Activity](/nexus/standalone-activity)
+
+**Sync messaging — as many as you need.** Reach these through `client.getWorkflowClient()`.
+They take effect during the handler call, still get link propagation, and do not require an async backing.
+
+- Signal, Signal-with-Start, Query, Cancel, and Terminate
```
KeithB-Temporal: I don't think we are going to have these 3 at pre-release: Query, Cancel, and Terminate

docs/encyclopedia/nexus/nexus-sdk-v2.mdx
```diff
@@ -0,0 +1,502 @@
...
+
+## What this replaces
+
+`WorkflowRunOperation` and the synchronous `OperationHandler` are **de-emphasized, not removed**.
+Existing handlers keep working and there is no forced migration.
+
+Prefer `TemporalOperationHandler` for new work, including simple cases.
+Using one type everywhere means a handler that starts out synchronous can grow an async backing, or pick up a Signal, without changing shape.
+
+Beyond the handler, [parent-close policy](/nexus/operations) parity with Child Workflows — deciding what happens to the handler Workflow when the caller completes, fails, or is cancelled — is still outstanding in every SDK.
```
KeithB-Temporal: This is a very confusing sentence, and we aren't going to have parent-close-policy at pre-release, so I wouldn't even bring it up

docs/encyclopedia/nexus/nexus-sdk-v2.mdx
```diff
@@ -0,0 +1,502 @@
...
+:::
+
+Nexus SDK V2 changes how you implement a [Nexus Service](/nexus/services) contract.
+Instead of one helper type per Temporal primitive, there is a single handler type — `TemporalOperationHandler` — that can back an Operation with any Temporal primitive and that carries [bidirectional links](/nexus/execution-debugging#bi-directional-linking) across the Namespace boundary for you.
+
+Nothing on the wire changes, and no existing Operation stops working.
+The Service contract, [Nexus Endpoint](/nexus/endpoints) setup, and Worker registration are the same as before.
+What changes is the handler you write.
+
+## Why it changed
```
KeithB-Temporal: I'm struggling a bit with the whole framing of this doc, I **DON'T** want it to focus on why it is better than V1. TBH there were so few customers using the V1, we can basically assume that the vast majority of people coming to this  doc are new. I really want this doc to just espouse all of the new amazingness that the programing model can do.  Only at the very end should we even make reference to the old one to talk about the improvements -> migration path...

docs/develop/java/nexus/development-walkthrough/index.mdx
```diff
@@ -0,0 +1,89 @@
...
+:::caution
+
+This walkthrough covers [Nexus SDK V2](/nexus/sdk-v2), which is pre-release.
+APIs are experimental and may change in backwards-incompatible ways.
+
+:::
+
+This walkthrough builds one Nexus Service from nothing to a complete API, adding a single Nexus capability at each step.
+
+## Nexus Introduction
```
KeithB-Temporal: The Development walk through is not a good place to give an introduction to Nexus use cases. These should live on https://docs.temporal.io/evaluate/nexus (But also, keep this out of the scope of these changes)

docs/develop/java/nexus/development-walkthrough/index.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+**A durable orchestration layer for AI agents and tools.** Agent systems call tools that are slow, flaky, and occasionally expensive. Exposing each tool as a Nexus Operation gives every call durable execution, automatic retries, and a record in [Event History](/encyclopedia/event-history) — and lets the agent and the tools live in different Namespaces, owned by different teams, written in different languages. See [Build AI applications with Temporal](/with-ai) for the wider picture.
+
+**A shared facade for extensibility.** A Nexus Service can front something that is not a Temporal Workflow at all — an existing internal API, a legacy job queue, a third-party endpoint. Write the wrapper once, run it as one Worker fleet, and every team calls the same Operations instead of each writing its own integration. Because callers only depend on the contract, the team behind it can modify or update the service without breaking anyone.
+
+## The sample problem
```
KeithB-Temporal: The entire development walkthrough should stay tighter to implementing this specific scenario. And each steps should be framed around the scenario, (IE Let's define the purchase object, now we need to have our sales team use it in Golang, but our distribution team wants to use it in .NET, we need to make a call to a 3rd party system to make an Authorization call...let's use a Nexus Invoked Standalone Activity, because it is just one call, we need to make a call to an order auditing system that needs to run a multi-step workflow, let's use a Invoke a Workflow, etc)

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
Evanthx: I had a lot of similar thoughts about naming confusion. I'm not at all sure!

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
Evanthx: I get your point, but this is the doc about starting a standalone activity - so in this document, it seems to make sense to differentiate.

Either way, I removed it from the description, and modified the text in the doc a few lines lower.

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,314 @@
...
+
+- **Durability per call.** Retries on the policy you set, timeouts you control, and a record of every attempt including the last error.
+- **Deduplicated starts.** A stable Activity Id means a retried or redelivered request targets the same Activity Execution instead of starting a second one, so you do not need a separate store of processed Ids.
+- **Protection from a failing dependency.** Repeated retryable failures trip the [circuit breaker](/nexus/operations#circuit-breaking) rather than piling retries onto a system that is already struggling.
+- **A boundary between teams.** Callers get scoped access to named Operations, not write access to your Namespace, and they never see your Task Queues, your Workers, or your implementation.
+- **Room to change your mind.** As long as the contract holds you can rewrite what runs behind an Operation, including replacing the Activity with a Workflow later, and no caller changes.
+
+Callers are free either way: a caller Workflow can invoke the Operation as one step of a larger process, or a Client can start it directly as a [Standalone Nexus Operation](/standalone-nexus-operation) with no Workflow on either side.
```
KeithB-Temporal: Callers are free either way... that word is scary. I think you mean free to implement either way right? We've got to be careful not confuse that free as in operations don't carry a cost factor.

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,314 @@
...
+
+Here are some example use cases.
```
KeithB-Temporal: I really like this section, but would prefer use cases this primitive is optimized for. Or more ideal, these are the scenarios that customers use this functionality for...

docs/develop/java/nexus/development-walkthrough/index.mdx
```diff
@@ -0,0 +1,84 @@
+---
+id: index
+title: Nexus Development Walkthrough - Java SDK
```
KeithB-Temporal: I'm wondering if we want to name this Microservice Development Walkthrough? 

While it is not strictly limited to be used for microservices, it more clearly cues the customer use case that we think they would use this for.

docs/develop/java/nexus/development-walkthrough/index.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+:::
+
+This walkthrough builds one Nexus Service from nothing to a complete API, adding a single Nexus capability at each step.
+
+## Nexus
+
+A [Nexus Service](/evaluate/nexus) is a contract that one team publishes and other teams call, across [Namespace](/namespaces) boundaries, without sharing code or a deployment.
+
+## A sample problem
```
KeithB-Temporal: A Reference Example

docs/develop/java/nexus/development-walkthrough/index.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+A purchase request needs approval before it can proceed.
```
KeithB-Temporal: To aid with the use of this Developer Guide, it is best to walk through it grounded in how it would be used to solve a real world problem, so we are going to solve the problem: A purchase request needs approval before it can proceed.

KeithB-Temporal: Also, the approver problem type (or Human-In-The-Loop, HITL) is a very good reference problem because it is a key scenario that customers use Temporal and Nexus for.

docs/develop/java/nexus/development-walkthrough/define-the-data-contract.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+Start with the contract, not the code.
+
+The contract is the only thing a caller and a handler share. Everything else — which language each side is written in, whether an Operation is backed by a Workflow or an Activity, which Task Queue the Worker polls — is private to one side and can change without the other side knowing.
+
+## Why the contract comes first
+
+Writing the contract first is what makes the Service polyglot.
+
+**Every sample in this walkthrough, in every language, is generated from this one contract.** A Go caller can call the Java handler built here. The Java caller built here can call a Python handler. Neither side hand-writes the request and response types, so neither side can drift from the other.
```
KeithB-Temporal: I like this sentiment, "A Go caller can call the Java handler built here. The Java caller built here can call a Python handler. Neither side hand-writes the request and response types, so neither side can drift from the other."  BUT it leaves me confused WHAT language are you going to use for caller in the tutorial and what language are you going to use to the handler in the tutorial? I think we add sentence to clarify that.

docs/encyclopedia/nexus/nexus-sdk-v2.mdx
```diff
@@ -0,0 +1,506 @@
+---
+id: nexus-sdk-v2
```
KeithB-Temporal: TODO We need a better name for V2 Nexus SDK. Maybe we just call it the Temporal Operations Handler and frame it as a significantly improved Development Primitive optimized for Nexus.

docs/develop/java/nexus/development-walkthrough/call-the-service.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+Call `requestApproval` from a Workflow in the caller Namespace. The caller knows the Endpoint name and the contract, and nothing else about the handler.
+
+## Use the generated interface as a stub
+
+In Java, the generated Service interface works directly as a Nexus Service stub. Create it inside the caller Workflow with the Endpoint name and Operation options, then call its methods as if they were local.
+
+Because the stub is the generated interface, the call is type-checked against the contract at compile time. A field the contract does not have will not compile, and a payload the contract forbids is rejected by the generated validator before it reaches the wire.
+
+`{sample code will be here}`
```
KeithB-Temporal: Implement in Java and call from Java doesn't really make the scenario pop. If the guide shows how to implement the service in Java, I propose that we show how to call from another language such as Python to really show off the power.  We can of course keep the how to call the service from Java there too (to make it easy to find), but I think we should lead showing off it being called in another language.

docs/develop/java/nexus/development-walkthrough/index.mdx
```diff
@@ -0,0 +1,85 @@
...
+7. [Add messaging](/develop/java/nexus/development-walkthrough/add-messaging)
+8. [Send messages](/develop/java/nexus/development-walkthrough/send-messages)
+9. [Add a Standalone Activity](/develop/java/nexus/development-walkthrough/add-a-standalone-activity)
+10. [Call the Standalone Activity](/develop/java/nexus/development-walkthrough/call-the-standalone-activity)
+
+Then: [Debugging, common pitfalls, and tips](/develop/java/nexus/development-walkthrough/debugging-and-tips).
+
+## Before you start
+
+You need two Namespaces, one for the handler and one for the caller, so the walkthrough crosses a real Namespace boundary. A [local development server](/develop/run-a-development-server) with two Namespaces is enough for steps 1 through 4; step 5 covers both the development server and Temporal Cloud.
```
KeithB-Temporal: Put in the commands to create the sample namespaces for them if they don't already have namespaces they want to work from.

docs/encyclopedia/nexus/nexus-client-code-generator.mdx
```diff
@@ -0,0 +1,400 @@
...
+the pure-schema [`temporal.yaml`](https://github.com/temporalio/nex-gen/blob/main/samples/schemas/temporal.yaml),
+and a multi-file closure under [`kb/`](https://github.com/temporalio/nex-gen/tree/main/samples/schemas/kb) showing how types split across files resolve through `$ref`.
+The `kb/` closure starts at [`kb.nexusrpc.yaml`](https://github.com/temporalio/nex-gen/blob/main/samples/schemas/kb/kb.nexusrpc.yaml) and pulls in types from its [`content/`](https://github.com/temporalio/nex-gen/tree/main/samples/schemas/kb/content) and [`tree/`](https://github.com/temporalio/nex-gen/tree/main/samples/schemas/kb/tree) subdirectories.
+
+## Install the generator
+
+Build the `nexgen` binary from source with a Rust toolchain:
+
+```bash
+git clone https://github.com/temporalio/nex-gen.git
```
KeithB-Temporal: Remove the hyphen here.

docs/encyclopedia/nexus/nexus-client-code-generator.mdx
```diff
@@ -0,0 +1,400 @@
...
+cd nex-gen
```
KeithB-Temporal: Remove the hyphen here.

docs/develop/java/nexus/development-walkthrough/generate-code.mdx
```diff
@@ -0,0 +1,49 @@
...
+- The **handler** provides an implementation for it, which the Worker registers.
+- The **caller** uses the interface directly as a Workflow stub, so calls are type-checked against the contract.
+
+The generated validators run when a payload is parsed and again when it is serialized, so a request that violates the contract is rejected at the boundary rather than reaching your Workflow. Violations aggregate into one error naming every field that failed, which a handler maps to `BAD_REQUEST`.
+
+## Generate for Java
+
+Java generation requires a package name whose last segment matches the output directory name. See [Generate code](/nexus/client-code-generator#generate-code) for the full command shape and the per-language flags.
+
+`{sample code will be here}`
```
KeithB-Temporal: Really need this for the sample to be useful...

docs/develop/java/nexus/development-walkthrough/choose-backing-implementation.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+The tradeoff is that an Activity has no Workflow's ability to receive messages or hold state, and cancellation only works if the Activity heartbeats. See [Nexus Standalone Activity](/nexus/standalone-activity).
+
+## Workflow
+
+More than one step, or any need for durable intermediate state.
+
+The Operation starts a Workflow and completes when that Workflow returns, so the Workflow's return value is the Operation's result. Use this when the work orchestrates several Activities, needs a timer, or needs to survive a Worker restart partway through.
+
+## Entity Workflow
```
KeithB-Temporal: As I am going through this guide, I am struggling with Entity Workflow. Workflow and Standalone Activities are documented primitives: https://docs.temporal.io/workflows and https://docs.temporal.io/standalone-activity 
EntityWorkflow is just a Workflow that is long running, it is not a different primitive, so I find this differentiation jarring and confusing.

docs/develop/java/nexus/development-walkthrough/choose-backing-implementation.mdx
```diff
@@ -0,0 +1,66 @@
+---
+id: choose-backing-implementation
+title: Step 3 - Choose the backing implementation
```
KeithB-Temporal: This step is odd to me, because this is strictly a learning step... it doesn't actually result in me setting or coding anything...

docs/develop/java/nexus/development-walkthrough/implement-the-service.mdx
```diff
@@ -0,0 +1,74 @@
...
+tags:
+  - Nexus
+  - Java SDK
+---
+
+Implement the generated Service interface, back `requestApproval` with the approval Workflow, and run a Worker that hosts both.
+
+## Write the approval Workflow
+
+The Workflow is an ordinary Temporal Workflow. Nothing in it is Nexus-specific, and it could be started directly by a Client instead.
```
KeithB-Temporal: Is it an ordinary Temporal Workflow? The previous step pretty tells us that it should be an entity workflow.

docs/develop/java/nexus/development-walkthrough/implement-the-service.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+`{sample code will be here}`
+
+## Implement the Operation with TemporalOperationHandler
+
+Use `TemporalOperationHandler` for every Temporal-backed Operation, including simple ones. It is the entry point in [Nexus SDK V2](/nexus/sdk-v2), and starting with it means an Operation can later gain a Signal or change its backing without changing shape.
+
+`TemporalOperationHandler.create(...)` gives your start handler a context, a Nexus-aware Client, and the Operation input. Call `startWorkflow` on that Client and return its result. The Operation then completes when the Workflow returns, delivering the Workflow's return value to the caller.
+
+The Client is not an ordinary Temporal Client. It propagates bidirectional links and request Ids automatically, so the caller's Execution and the approval Workflow are connected in the UI. Fetching your own Client inside a handler works but gives up that linking.
```
KeithB-Temporal: I REALLY like this line to explain the Temporal Operation Handler. We should clearly use this or similar to describe the reason for why we introduced Nexus SDK Ergonomics improvements

docs/develop/java/nexus/development-walkthrough/_steps/publish-in-nexus.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+On a development server, create it with the CLI:
+
+`{sample code will be here}`
+
+In Temporal Cloud, create it in the UI under Nexus, or with `tcld`. See [Create a Nexus Endpoint](/nexus/registry#create-a-nexus-endpoint).
+
+Endpoint names are unique within the Registry. In Temporal Cloud the Registry is global across your whole Account and spans every Namespace; in a self-hosted deployment it is scoped to the Cluster.
+
+## Allow caller Namespaces
```
KeithB-Temporal, line 23: There is an important distinction here, this step only applies if you are doing this guide against a Temporal Cloud backing instance. If this is being done against a self hosted Temporal instance on your box this step doesn't apply.

docs/develop/java/nexus/development-walkthrough/call-the-service.mdx
```diff
@@ -0,0 +1,59 @@
...
+toc_max_heading_level: 4
+unlisted: true
+tags:
+  - Nexus
+  - Java SDK
+---
+
+Call `requestApproval` from a Workflow in the caller Namespace. The caller knows the Endpoint name and the contract, and nothing else about the handler.
+
+## Use the generated interface as a stub
```
KeithB-Temporal: I'm struggling to follow this section. Needs more examples, grounding to reference code.

docs/develop/java/nexus/development-walkthrough/add-messaging.mdx
```diff
@@ -0,0 +1,79 @@
...
+---
+
+The approval blocks waiting for a decision. Now give callers a way to interact with it while it waits.
+
+Three Operations get added, one per [message](/sending-messages) type. Which message type to use is determined by what the caller needs back, not by preference.
+
+| Operation | Message type | Why this type |
+| --- | --- | --- |
+| `remindApprover` | Signal | Fire-and-forget. The caller does not need a response, only for the nudge to happen. |
+| `getApprovalStatus` | Query | Reads state without changing it. Never blocks, never writes. |
```
KeithB-Temporal: I see what we are trying to do with putting query in here, but our docs shouldn't be referencing unreleased stuff.

docs/develop/java/nexus/development-walkthrough/add-messaging.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+Because it is an async backing, there is at most one per Operation invocation. A handler can still combine it with sync side effects.
+
+`{sample code will be here}`
+
+:::caution Query is not in the pre-release
+
+`getApprovalStatus` belongs in the contract, but Query is not available as sync messaging on the Nexus-aware Client in the pre-release, so this Operation cannot be implemented yet. See [Nexus SDK V2](/nexus/sdk-v2).
+
+Adding the Query handler to the Workflow is still worth doing — it costs nothing and the Operation can be wired up once Query lands. Until then, a caller that needs in-flight progress has to obtain it outside this Service.
```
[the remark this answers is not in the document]
KeithB-Temporal: I understand this comment is true, but as this guide is about code gen and the improved Temporal Operation Handler class, I would exclude things that fall outside of this class for now. (or save it for a future update when we add query)

docs/develop/java/nexus/development-walkthrough/add-messaging.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+:::
+
+## Keep the responsibilities separate
```
KeithB-Temporal: The entire wording of this section is hard to follow. 

1) I don't like just saying mixing these up is a common mistake.
2) It is better to just explain the negative behavior. Query is better for when. you are trying to get state without updating it, and it can't be relied on to get the state after retention period expires,'
3) But this gives no explanation about how to misuse signals or update...

I think this section it really just to discuss DON'T Use Query for getting the final workflow status...

docs/develop/java/nexus/development-walkthrough/send-messages.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+## Nudge, check, and decide
+
+`{sample code will be here}`
+
+What differs between them is what you get back and how long it takes.
+
+`remindApprover` returns nothing and completes as soon as the Signal is accepted. Accepted is not the same as handled — the Signal is durably recorded and the Workflow will process it, but the Operation does not wait for that. If the caller needs confirmation that the nudge took effect, it needs an Update, not a Signal.
+
+`getApprovalStatus` returns the current progress immediately. Call it when something needs to report on a pending approval; do not call it in a loop waiting for the decision. Note that this Operation is [not implementable in the pre-release](/develop/java/nexus/development-walkthrough/add-messaging#expose-them-as-nexus-operations), because Query is not yet available as sync messaging.
```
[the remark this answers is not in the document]
KeithB-Temporal: Remove things not included in the pre-release until they are available in a Published Release.

docs/develop/java/nexus/development-walkthrough/send-messages.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+Sometimes a caller wants to start an approval and immediately attach information to it, without a race between the two calls.
+
+Signal-with-Start does both atomically: if the target Workflow is not running it is started, and either way the Signal is delivered. It is a sync side effect on the Nexus-aware Client, reached through `client.getWorkflowClient()`, so a handler Operation can offer it directly.
+
+This is also how you make a Signal safe to send to an approval that may not exist yet. A plain Signal to a missing Workflow fails; Signal-with-Start creates it.
+
+`{sample code will be here}`
+
+:::caution Update-with-Start is not yet available
```
KeithB-Temporal: Remove this. It's not included in the Temporal Operation Handler yet, so we should not be discussing it.

docs/develop/java/nexus/development-walkthrough/send-messages.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+Until it lands, an Operation that needs both must start the Workflow and then submit the Update as a separate call, which is not atomic. Signal-with-Start is available and covers the case where the caller does not need a response.
+
+:::
+
+## You cannot re-attach to get a result
+
+A caller that did not start an approval cannot ask Nexus for its final decision.
+
+`requestApproval` delivers the decision to whoever awaited it. There is no Operation that attaches to an already-running approval and waits for its result, because Get Workflow Result as an async backing is [not yet available](/nexus/sdk-v2) in any SDK.
```
KeithB-Temporal: Remove " because Get Workflow Result as an async backing is [not yet available](/nexus/sdk-v2) in any SDK."  (From discussions with product we have not decided to add this. We shouldn't theorize about potential future features in the docs, we should explain what is there.)

Just change this focus to be only on no way to attach to already running approval, so instead use this pattern...

docs/develop/java/nexus/development-walkthrough/add-a-standalone-activity.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+Nothing about it is Nexus-specific. The same Activity could be called from a Workflow.
+
+`{sample code will be here}`
+
+## Back the Operation with it
+
+Use `TemporalOperationHandler` as with every other Operation, but call `startActivity` on the Nexus-aware Client instead of `startWorkflow`. The Operation starts an Activity Execution with no parent Workflow and completes when the Activity returns.
+
+Before Activity-backed Operations, this Operation would have needed a Workflow whose only job was to call this one Activity — a wrapper with its own Event History and Workflow Id, providing nothing.
```
KeithB-Temporal: Remove this. We should not be explaining this feature in terms of how it used to be done. Assume the caller has not idea about how it used to be done, we should just explain why it is a good pattern and maybe other things it could be used for.

docs/develop/java/nexus/development-walkthrough/add-a-standalone-activity.mdx
```diff
@@ -0,0 +1,68 @@
...
+`{sample code will be here}`
+
+### Options an Activity-backed Operation requires
+
+`StartActivityOptions` needs two things that a Workflow-called Activity does not, because there is no parent Workflow to supply them:
+
+- **An Activity Id**, unique within the Namespace.
+- **A Task Queue.** It does not have to be the Endpoint's target Task Queue, so notifications can run on their own Worker fleet.
+
+Derive the Activity Id from the Nexus request Id to make the start idempotent. The server retries a Nexus start request with the same request Id, so each retry targets the same Activity Id instead of sending a second notification.
```
KeithB-Temporal: This explanation could be improved. Lead with the outcome. In order to ensure that server retries are idempotent and don't spam notifications to the user, it is important to do this pattern.....

(Also consider mentioning a few other cases where this pattern might be helpful)

docs/develop/java/nexus/development-walkthrough/_steps/add-a-standalone-activity.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+That last point matters here more than usual. A duplicate Workflow start is usually harmless; a duplicate notification is a second email to a real person.
+
+## Register the Activity on the Worker
+
+Add the Activity implementation to the same Worker that hosts the Nexus Service. An Activity-backed Operation needs no Workflow implementation registered for it.
+
+`{sample code will be here}`
+
+## Cancellation needs heartbeating
```
KeithB-Temporal, line 110: This is good information, but not useful to the guide. The guide should stay a bit tighter to How-to-Guide, move this content to Reference.

docs/develop/java/nexus/development-walkthrough/call-the-standalone-activity.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+`{sample code will be here}`
+
+Every step crossed a Namespace boundary, and the caller never learned a Workflow Id, a Task Queue, or which primitive backed any Operation.
+
+## Trace it end to end
+
+Open the caller Workflow in the UI and follow the links. Because the handlers used the Nexus-aware Client, each Operation is connected to the Execution it started or messaged in the handler Namespace, and you can move between the two Namespaces in one view.
+
+The exception is `getApprovalStatus`, which is [not implementable in the pre-release](/nexus/sdk-v2) because Query is not yet available as sync messaging.
```
KeithB-Temporal: Just remove this until we are ready to add query

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
Evanthx: Fixed the wording!

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
Evanthx: Took me a few drafts but how is "A sampling of customer use cases this was built to address follows."

jsundai:
Looks great! When it's ready for review, feel free to drop a comment and post it in the slack channel then someone from the team will review it!

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,451 @@
...
+		Name: "greet",
+		Start: func(
+			ctx context.Context,
+			nc temporalnexus.NexusClient,
+			input GreetingInput,
+			opts temporalnexus.StartTemporalOperationOptions,
+		) (temporalnexus.TemporalOperationResult[GreetingOutput], error) {
+			return temporalnexus.StartActivity(ctx, nc, client.StartActivityOptions{
+				ID:                  "greet-" + opts.RequestID,
+				TaskQueue:           TaskQueueName,
```
Quinn-With-Two-Ns: nit: Task queue should be optional

docs/encyclopedia/nexus/nexus.mdx
```diff
@@ -130,3 +130,14 @@ Each step is a separate, durable Operation with its own retries and failure hand
 - [Nexus error handling](/nexus/error-handling) - Error types and how they surface in caller Workflows.
 - [Nexus metrics](/nexus/metrics) - SDK, Cloud, and OSS cluster metrics.
 
+## Pre-release: the new Nexus developer experience
+
+A new way of building Nexus Services is in pre-release. It combines a contract-first workflow, where one schema generates typed models and Service definitions for every language, with a single handler type that can back an Operation with a Workflow, an Update, or an Activity.
```
Quinn-With-Two-Ns:
```suggestion
A new way of building Nexus Services is in pre-release. It combines a contract-first approach, where one schema generates typed models and Service definitions for every language, with a single handler type that can back an Operation with a Workflow, an Update, or an Activity.
```

nit: I would avoid workflow since it has other meaning in Temporal

docs/encyclopedia/nexus/temporal-operation-handler.mdx
```diff
@@ -0,0 +1,640 @@
...
+
+**Get observability across the Namespace boundary without wiring it.** The Client handed to your handler propagates [bidirectional links](/nexus/execution-debugging#bi-directional-linking) and request Ids on every call it makes. The caller-side and handler-side Executions are connected in the UI and in [Event History](/encyclopedia/event-history), so a single trace crosses the boundary between two teams' Namespaces.
+
+**Stay idempotent through retries.** The server retries Nexus start requests, and the request Id travels with them. Deriving the backing Execution's Id from it means a retry targets the same Execution rather than starting a second one.
+
+**Cancel through the same handler.** The Operation token records which kind of Execution backs the Operation, so a cancellation request reaches the right place. The default behavior is usually what you want, and each kind can be overridden when it is not.
+
+**Grow a handler without rewriting it.** An Operation that starts out completing inline can later gain an async backing, or pick up a Signal, without changing handler type or breaking its contract.
```
Quinn-With-Two-Ns:
```suggestion
**Grow a handler without rewriting it.** An Operation that starts out completing inline can later gain an async backing, or send a Signal, without changing handler type or breaking its contract.
```

I am not sure what pick up means in this context, my initial read was it could receive a signal but that is not true.

docs/encyclopedia/nexus/temporal-operation-handler.mdx
```diff
@@ -0,0 +1,640 @@
...
+            wf => wf.UpdateShippingAddressAsync(input),
+            new() { WaitForStage = WorkflowUpdateStage.Accepted }));
+```
+
+  </SdkTabs.DotNet>
+</SdkTabs>
+
+Three constraints apply to Update-backed Operations, and the first two fail the Operation rather than degrading:
+
+- **Only the accepted stage is supported.** A Nexus Operation can only back an asynchronous Update, so the wait-for-stage must be "accepted". Go, Java, and .NET require you to set it and fail the Operation if it is anything else. Python and TypeScript set it for you.
```
Quinn-With-Two-Ns: This is a bug in Python and Typescript. Lets fix it not document it

docs/encyclopedia/nexus/temporal-operation-handler.mdx
```diff
@@ -0,0 +1,640 @@
...
+- **A callback URL is required.** The Operation is how the Update's result reaches the caller, so a caller that provided no callback URL gets a `BAD_REQUEST` handler error.
```
Quinn-With-Two-Ns: Remove this , callback url is not something exposed to the user. If the callback URL is not present that is a server bug

docs/encyclopedia/nexus/temporal-operation-handler.mdx
```diff
@@ -0,0 +1,640 @@
...
+```
+
+  </SdkTabs.DotNet>
+</SdkTabs>
+
+The same Client also offers Signal-with-Start, and a handler may send several messages before returning.
+
+### Back an Operation with an Activity
+
+Call `startActivity` when the work is a single durable step. The Activity runs with no parent Workflow, so the options require an Activity Id and at least one timeout. Java also requires a Task Queue; the other SDKs default it to the Task Queue the Operation is running on. See [Nexus Standalone Activity](/nexus/standalone-activity).
```
Quinn-With-Two-Ns: Remove the note about Java that is a bug

docs/encyclopedia/nexus/temporal-operation-handler.mdx
```diff
@@ -0,0 +1,640 @@
...
+	})
+```
+
+  </SdkTabs.Go>
+  <SdkTabs.Java>
+
+```java
+@OperationImpl
+public OperationHandler<GreetingInput, GreetingOutput> startGreeting() {
+  return TemporalOperationHandler.create(
```
Quinn-With-Two-Ns: Should use the `@TemporalOperation` annotation

Quinn-With-Two-Ns: This goes for all samples

docs/encyclopedia/nexus/temporal-operation-handler.mdx
```diff
@@ -0,0 +1,640 @@
...
+    });
+  },
+});
+```
+
+  </SdkTabs.TypeScript>
+  <SdkTabs.DotNet>
+
+```csharp
+TemporalOperationHandler.FromHandleFactory<GreetingInput, GreetingOutput>(
```
Quinn-With-Two-Ns: Should use `[TemporalOperation]` decorator, this goes for all dotnet samples

docs/encyclopedia/nexus/temporal-operation-handler.mdx
```diff
@@ -0,0 +1,640 @@
...
+                Id = $"greet-{context.RequestId}",
+                TaskQueue = TaskQueueName,
+                StartToCloseTimeout = TimeSpan.FromSeconds(10),
+            }));
+```
+
+  </SdkTabs.DotNet>
+</SdkTabs>
+
+## Coming from the earlier handler APIs
```
Quinn-With-Two-Ns: I would add a note that there is no issue if they have workflow operations in progress. `TemporalOperationHandler` can cancel them just fine

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,451 @@
...
+  </SdkTabs.DotNet>
+</SdkTabs>
+
+### Required options
+
+Starting an Activity this way needs values that a Workflow-called Activity does not.
+
+- **An Activity Id**, unique within the Namespace. There is no parent Workflow to scope it.
+- **A timeout.** At least one of start-to-close or schedule-to-close.
+- **A Task Queue.** Java requires it. Go, Python, TypeScript, and .NET default it to the Task Queue the Operation is running on. Setting it explicitly lets the Activity run on its own Worker fleet rather than the one the Endpoint targets.
```
Quinn-With-Two-Ns: Noted elsewhere but lets not document bugs since we will fix these before we go to pre-release

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
```diff
@@ -0,0 +1,451 @@
...
+    client,
+    new TemporalWorkerOptions(TaskQueueName).
+        AddActivity(GreetingActivities.GreetAsync).
+        AddNexusService(new GreetingNexusServiceHandler()));
+```
+
+  </SdkTabs.DotNet>
+</SdkTabs>
+
+## Cancellation requires heartbeating
```
Quinn-With-Two-Ns: A lot of this is basically duplicate with https://docs.temporal.io/activity-execution#cancellation so I would just link to the reference docs and treat this as a reminder

Quinn-With-Two-Ns: basically I am saying there is nothing Nexus specific here we need to document

docs/encyclopedia/nexus/nexus-standalone-activity.mdx
Evanthx: If I am getting this right, it seems like it is in the other four languages, but not Java? That might be something we need to fix

docs/encyclopedia/nexus/temporal-operation-handler.mdx
Evanthx: Done

docs/develop/java/nexus/development-walkthrough/_steps/add-a-standalone-activity.mdx
```diff
@@ -0,0 +1,140 @@
...
+                ApprovalActivities.class,
+                ApprovalActivities::notifyRequester,
+                input.getRequester(),
+                input.getDecision(),
+                StartActivityOptions.newBuilder()
+                    // Deriving the Activity Id from the request Id keeps a retried Nexus start
+                    // request targeting the same Activity Execution instead of sending a second
+                    // notification.
+                    .setId("notify-" + ctx.getRequestId())
+                    .setTaskQueue(HandlerWorker.DEFAULT_TASK_QUEUE_NAME)
```
Quinn-With-Two-Ns, line 63: Nit: I think we can remove this from all the code since it is not needed

jsundai:
In the process of reworking this (moving this into the sidebar, turning it into one page, added tabbed component, etc) - gonna make a few more commits and then I'll review the content.

PR #5135, opened 2026-08-18
Add guide for migrating Celery task queues to Standalone Activities

brianmacdonald-temporal:
## What does this PR do?

Adds a Python guide, [Migrate a Celery task queue to a Temporal Standalone Activity](https://github.com/temporalio/documentation/blob/celery-to-standalone-activity/docs/guides/celery-to-standalone-activity.mdx), that walks through converting a Celery task into a Standalone Activity:

- Maps Celery concepts (task, worker, broker, `delay()`, `AsyncResult.get()`, `max_retries`, Flower) onto their Temporal equivalents
- Runs an Activity-only Worker with no Workflow code
- Replaces `.delay()` and `.get()` with `client.start_activity()` and `client.execute_activity()`
- Migrates `max_retries` / `self.retry()` to a Retry Policy
- Uses `client.list_activities()` and `client.count_activities()` in place of Flower

Also links the page from the Guides sidebar and adds a card to the Guides landing page grid under a new `Migration` tag.

## Notes to reviewers

Opened as a draft: the guide documents Standalone Activities, which are `publicPreview` in `src/constants/featureReleaseTypes.js`. Happy to mark it ready once the docs team is clear to merge preview coverage for this feature.

`yarn build` passes; the page and its landing page card both appear in the build output.

Two follow-ups intentionally left out of this PR:

- `docs/guides/loyalty-points-revised.mdx` is a separate untracked draft in my working tree and is not included here. It's currently an unlinked orphan and needs its own PR.
- The new page carries `author: n/a` in its frontmatter, which no other guide uses. Let me know if you'd like it dropped.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217611767947090">EDU-6982 Add guide for migrating Celery task queues to Standalone Activities</a>


docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,449 @@
...
+Start the development server:
+
+```bash
+temporal server start-dev
+```
+
+You will see output confirming the server is running, including two addresses:
+
+```
+[secondary_label Output]
```
MasonEgger: This isn't rendering the standard DO style, but instead just adding secondary_label to the rendered output.

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,449 @@
...
+Server:  localhost:7233
+UI:      http://localhost:8233
+```
+
+Your application code will connect to `localhost:7233`. The Web UI at `http://localhost:8233` lets you inspect Activities and their results; Standalone Activities appear under their own item in the UI's navigation. Leave this process running and open a new terminal for the remaining steps.
+
+## Step 4 — Converting a Celery task into an Activity
+
+In this step, you will take a Celery task and rewrite it as a Temporal Activity. The work inside — network calls, database writes, file I/O — stays the same. You write a Standalone Activity exactly the way you would write any Temporal Activity; nothing about the function marks it as "standalone." What makes it standalone is how you invoke it, which you will do in Step 6.
```
MasonEgger: Are you adopting an anti-emdash rule? While these are correct (because of course they are you wrote it) I worry about things like this in docs. The em-dash is the first thing anyone points (correctly or not) as AI content.

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,449 @@
...
+```python
+handle = client.get_activity_handle(
+    activity_id="welcome-email-42",
+    run_id="the-run-id",
+)
+```
+
+## Step 8 — Migrating task retries to a Retry Policy
+
+In Celery, retries are your responsibility: you set `max_retries` and call `self.retry()` inside the task. With Temporal, retries are automatic and declarative. By default, a Standalone Activity retries indefinitely with exponential backoff, so migrating usually means *adding a limit back in* to match Celery's `max_retries`. In this step, you will attach a retry policy to the invocation.
```
MasonEgger: While this is correct, it makes my brain itch a little bit. Because you in one breath say that retries are automatic by default and then in the next breath, say let's add a retry policy. If there's one thing that I've noticed about AI generated Temporal code is that it very often tries to add a retry policy. When in reality, the default retry policy is designed to be ideal for nearly every use case. I don't know how in this pros you would call this out without having a retry policy in there, but more often than not whenever AI tries to add a retry policy to something. It artificially creates a worse experience because it adds these constraints that the business logic itself did not need. So it may be worth exploring why a limitation on number of retries is even needed from a business logic perspective. But that might be getting too deep in the weeds of teaching Temporal and proper business logic design.

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,449 @@
...
+Run it:
+
+```bash
+python inspect_activities.py
+```
+
+You will see one line per Activity execution, followed by a total count:
+
+```
+[secondary_label Output]
```
MasonEgger: Bare render of this again

MasonEgger (commented):
One other thing that might be worth mentioning is using start delay on a SAA to delay the invocation for a set amount of time. This is basically the equivalent of if someone wants to have a workflow that immediately fires a timer and then does an activity. I will also gladly accept if you call that scope Creek for the purpose of this article. It's just a neat little tool.

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,449 @@
...
+
+import { ReleaseNoteHeader } from '@site/src/components';
+
+<ReleaseNoteHeader
+  featureName="standaloneActivity"
+/>
+
+[Celery](https://docs.celeryq.dev/en/stable/) is a distributed task queue that runs background jobs by pushing messages through a broker (such as Redis or RabbitMQ) to a pool of worker processes. Most Celery tasks are self-contained: send one email, resize one image, call one API. For that kind of single-step job, you want Durable Execution and automatic retries without having to stand up an orchestration layer around each task.
+
+[Temporal](https://docs.temporal.io/) **Standalone Activities** fit that need. A Standalone Activity is an Activity you start directly from a Temporal Client, without wrapping it in a Workflow. You get Temporal's durability, retries, timeouts, and visibility for an individual unit of work, which maps almost one-to-one onto a Celery task. Because there is no Workflow to run a single Activity, Standalone Activities also use fewer resources than orchestrating one Activity through a Workflow.
```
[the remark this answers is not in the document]
Duncanma: @brianmacdonald-temporal I also think it's weird to link the word Temporal at all... and maybe we should just say "Standalone Activities"?

flippedcoder (approved):

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,448 @@
...
+        ),
+    )
+    print(f"Result: {result}")
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
+```
+
+Here, `maximum_attempts=5` mirrors Celery's `max_retries`, and `maximum_interval` caps the backoff between attempts. The `non_retryable_error_types` list names errors that should fail immediately without retrying — the equivalent of *not* calling `self.retry()` for a permanent failure. To raise such an error from the Activity, use `ApplicationError` with `non_retryable=True` in `my_activity.py`:
```
joshmsmith: seems to be in conflict with discussion about max retries=6 above on #332

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,448 @@
...
+from my_activity import WelcomeEmailInput, send_welcome_email
+
+
+async def main():
+    client = await Client.connect("localhost:7233")
+
+    handle = await client.start_activity(
+        send_welcome_email,
+        args=[WelcomeEmailInput(42)],
+        id="welcome-email-42",
```
joshmsmith: optional - suggest you might want to change  activity id between steps, a la "welcome-email-42-step6", "-step7"
for example, someone could run through this quickly and do step 6 and then immediately step 7 or add a note about if they do that, they're hitting intended dedupe behavior

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,448 @@
...
+| Flower / `celery inspect` | `client.list_activities()` / `client.count_activities()` | Monitor jobs |
+
+## Prerequisites
+
+Before you begin, you will need the following:
+
+- Python 3.9 or higher installed on your machine.
+- The Temporal Python SDK, version 1.23.0 or higher (installed in Step 2).
+- The Temporal CLI, version 1.7.0 or higher (installed in Step 2).
+- An existing Celery task you want to migrate, or the sample task shown in Step 4 if you are following along from scratch.
```
joshmsmith, line 50: Optional: suggest to add Temporal Server v1.31.0+ will be needed for production or "just use cloud" because standalone activities won't work on old self hosted versions

docs/guides/celery-to-standalone-activity.mdx
```diff
@@ -0,0 +1,448 @@
...
+
+import { ReleaseNoteHeader } from '@site/src/components';
+
+<ReleaseNoteHeader
+  featureName="standaloneActivity"
+/>
+
+[Celery](https://docs.celeryq.dev/en/stable/) is a distributed task queue that runs background jobs by pushing messages through a broker (such as Redis or RabbitMQ) to a pool of worker processes. Most Celery tasks are self-contained: send one email, resize one image, call one API. For that kind of single-step job, you want Durable Execution and automatic retries without having to stand up an orchestration layer around each task.
+
+[**Standalone Activities**](/standalone-activity) fit that need. A Standalone Activity is an Activity you start directly from a Temporal Client, without wrapping it in a Workflow. You get Temporal's durability, retries, timeouts, and visibility for an individual unit of work, which maps almost one-to-one onto a Celery task. Because there is no Workflow to run a single Activity, Standalone Activities also use fewer resources than orchestrating one Activity through a Workflow.
```
joshmsmith, line 24: it _may_ be useful to (temporarily) mention these are still public preview, in case that caveat is important to readers before they do the whole guide - it's mentioned at the end.

joshmsmith (approved):
This is good. Great guidance, examples for different ways of using Celery->Temporal, and i love the workflow guidance at the end.

made a few minor suggestions-

docs/guides/celery-to-standalone-activity.mdx
brianmacdonald-temporal: We've got a release-stage banner above the intro, so I think we're good there.

docs/guides/celery-to-standalone-activity.mdx
brianmacdonald-temporal: Fixed.

docs/guides/celery-to-standalone-activity.mdx
brianmacdonald-temporal: Nice catch; fixed.

docs/guides/celery-to-standalone-activity.mdx
brianmacdonald-temporal: Artifact of an earlier version; thanks for that.

PR #5139, opened 2026-08-19
Add missing integrations to the Python and TypeScript landing pages

DABH:
## What

Adds two integration bullets that were missing from SDK landing pages:

- **Langfuse** on `/develop/python` — added to `integrations-data.json` in #5086, but the landing-page bullet was never added
- **LangSmith** on `/develop/typescript` — in the grid data *and* `sidebars.js` with a local guide at `docs/develop/typescript/integrations/langsmith.mdx`, but missing from the landing page

## Why

Each SDK landing page lists every `integrations-data.json` entry for that SDK. I checked whether these omissions were deliberate, and they don't appear to be:

- Not an "external links are excluded" rule — 5 of the 12 existing Python bullets already point off-site (OpenAI Agents SDK, OpenBox, Parseable, Pydantic AI, Tenuo).
- Not a "mirror `sidebars.js`" rule — #5133 gave that as the rationale, but `sidebars.js` has only the 7 local Python pages while the list has 12. The actual convention is "every grid entry for this SDK."
- Not a scoping issue — Langfuse is tagged `sdk: "Python"` / `"Agent observability"`, the same shape as LangSmith and Braintrust, both already listed.

## Verification

I diffed all five SDK landing pages against `integrations-data.json`. Go, Java, and Ruby were already in sync; Python and TypeScript each had one gap, both fixed here. All five now match.

Ordering follows the grid's `localeCompare` sort, so `Langfuse` sorts before `LangGraph`.

The `sdk: null` entries (Datadog, Grafana Cloud, ClickStack, New Relic, SigNoz, Vantage, Prometheus) are Cloud metrics/cost integrations with no SDK binding, and are correctly absent from SDK landing pages.

Checked locally: `yarn build`, `vale --config .vale-ci.ini` on both edited files (0 errors/warnings), `yarn check:orphans`, `yarn lint:mermaid`, `bin/check-redirects-for-moved-pages.js`.

## Follow-up worth considering

This hand-maintained list duplicates `integrations-data.json` and has now drifted three times in eight days (#5133, and both gaps here). Options: render `<IntegrationsGrid defaultSdks={["Python"]} />` on the landing pages instead of hardcoded bullets, or add a `yarn check:integrations` script alongside the existing `check:metrics` / `check:orphans`. The check script is lower-risk — it keeps the current page design and just fails CI on drift. Happy to open that separately.

┆Attachments: <a href="https://app.asana.com/app/asana/-/get_asset?asset_id=1217621914136391">EDU-6986 Add missing integrations to the Python and TypeScript landing pages</a>


Duncanma:
Created https://github.com/temporalio/documentation/issues/5144 to track the larger problem where the SDK landing pages duplicate the sidebar, but with no system to keep them in sync

Duncanma (approved):

PR #5146, opened 2026-08-19
feat: add OpenTelemetry v2 docs

patbeqo:
## What does this PR do?

- Adds Go SDK docs for the OpenTelemetry v2 integration
- New guide covers traces, metrics and automatic instrumentation
- Links the new guide from the Go SDK observability and context propagation guides


flippedcoder (approved):
I made a few changes, but overall this was great!

docs/develop/go/integrations/opentelemetry-v2.mdx
```diff
@@ -0,0 +1,437 @@
...
+		)),
+	)
+	otel.SetMeterProvider(provider)
+	return provider, nil
+}
+
+```
+<!--SNIPEND-->
+
+Start a loopback-only HTTP endpoint that serves the Prometheus handler:
```
xumaple: The documentation is not supposed to simply be a How-to-use-the-samples-101.

This prometheus endpoint is purely a samples setup thing that allows the user to see their samples working e2e; it is not something every user must follow to get their own plugin setup working. Hence this doesn't belong here.

docs/develop/go/integrations/opentelemetry-v2.mdx
```diff
@@ -0,0 +1,437 @@
...
+			log.Println("Prometheus endpoint failed:", err)
+		}
+	}()
+	return server, nil
+}
+
+```
+<!--SNIPEND-->
+
+Opening the listener before starting the serving goroutine makes an address conflict fail during Worker startup. The
```
xumaple: We don't need all this fluff about specific edge cases to the samples. (going to stop commenting on the rest of this page as I think most of it can be cut)

xumaple (changes requested):
Take a look at the [langsmith docs](https://docs.temporal.io/develop/python/integrations/langsmith) for inspiration. Documentation is supposed to point the reader towards how to use the package, not how to run the samples.

docs/develop/go/best-practices/context-propagation.mdx
```diff
@@ -16,13 +16,15 @@ tags:
 description: How to propagate custom key-value data across Workflow, Activity, and Child Workflow boundaries using the Temporal Go SDK.
 ---
 
-Context propagation lets you pass custom key-value data from a Client to Workflows, and from Workflows to Activities and Child Workflows, without threading it through every function signature. Common use cases include propagating tracing IDs, tenant IDs, auth tokens, or other request-scoped metadata.
+Context propagation lets you pass custom key-value data from a Client to Workflows, and from Workflows to Activities and Child Workflows, without threading it through every function signature. Common use cases include propagating tenant identifiers, auth tokens, or other request-scoped metadata.
 
 {/* TODO: Link to /encyclopedia/context-propagation once that page lands */}
```
xumaple: @flippedcoder has this page landed yet?

xumaple: wait [it does](https://docs.temporal.io/encyclopedia/context-propagation) but it links back to this page. seems like there was some disagreement in the direction of who should link who. Can this line be deleted then?

docs/develop/go/best-practices/context-propagation.mdx
```diff
@@ -16,13 +16,15 @@ tags:
...
 
 :::tip
 
-If you want to propagate tracing context, check if there is a [built-in tracing interceptor](/develop/go/platform/observability#tracing) for your library before building a custom context propagator.
+For OpenTelemetry tracing, use the [OpenTelemetry v2 integration](/develop/go/integrations/opentelemetry-v2). The plugin
```
xumaple:
```suggestion
If you want to propagate tracing context specifically, there are several tracing utilities pre-built for you to use. [Check if any fit your needs.]((/develop/go/platform/observability#tracing))
```

xumaple: and then below, in the tracing section, we should do the actual link to OpenTelemetry.

Doing it here is misleading because otel isn't the only one we have.

docs/develop/go/integrations/opentelemetry-v2.mdx
```diff
@@ -0,0 +1,437 @@
...
+	if err := handle.Get(ctx, &result); err != nil {
+		return "", fmt.Errorf("unable to get Workflow Update result: %w", err)
+	}
+	return result, nil
+}
+
+```
+<!--SNIPEND-->
+
+The Update validator creates a replay-safe `validate-update` span. The handler creates a `handle-update` span,
```
xumaple: these are some pretty massive snippets to be putting into docs. I would look to see if you can isolate just the behavior needed to demonstrate what you're trying to do. It doesn't even need to be a full function.

docs/develop/go/platform/observability.mdx
```diff
@@ -41,6 +41,15 @@ This section covers features related to viewing the state of the application, in
 Each Temporal SDK is capable of emitting an optional set of metrics from either the Client or the Worker process.
 For a complete list of metrics capable of being emitted, see the [SDK metrics reference](/references/sdk-metrics).
 
+:::tip[Use the OpenTelemetry v2 integration]
+
+For new metrics setups that use OpenTelemetry, use the
+[OpenTelemetry v2 integration](/develop/go/integrations/opentelemetry-v2#export-temporal-sdk-metrics). It
```
xumaple: Something like this, not sure on wording. to communicate why this is actually better.

```suggestion
[OpenTelemetry v2 integration](/develop/go/integrations/opentelemetry-v2#export-temporal-sdk-metrics), which solves the context propagation issue in a more correct way. It
```

docs/develop/go/best-practices/context-propagation.mdx
flippedcoder: Yep, that can be deleted. Looks like it was left over from a while back.

xumaple (approved):

Issue #5144, opened 2026-08-19
SDK language pages duplicate the content from the sidebar


Duncanma:
But aren't automated, so they drift *often*

We should either automate this or possibly remove the duplication (should it just be intros to each subsection?)

DABH:
@Duncanma I looked into this and opened #5179 as a draft. Summary of what I found, since a couple of things argue against my own first instinct.

## The clearest example is four days old

#5118 added 19 Serverless Workers pages across seven SDKs. Landing-page links to that feature afterwards:

| SDK | pages | landing-page links |
|---|---|---|
| go | 3 | 1 |
| java, python, typescript, dotnet | 3 each | **0** |
| ruby, rust | 2 each | **0** |

A documented feature was unreachable from six of seven SDK landing pages. That is a better argument for this issue than any count of duplicated links.

## On your fork: automate or remove?

I researched both, deliberately including a pass whose brief was to argue *against* automating. It still came out for automating, on three pieces of evidence:

- `docusaurus.config.js` sets `onBrokenLinks: 'throw'`, so every one of the 319 bullet links on those pages is machine-checked. Replacing them with prose moves that content into the unchecked category — and every frontmatter defect I found this week (four wrong `description`/`title` values, fixed in #5155) was in exactly that unchecked category.
- `docs/develop/go/index.mdx` has a 16-line hand-written prose block (`### Where are SDK-specific code examples?`) that exists on Go and no other SDK, 4.5 months on.
- A tracked correctness sweep added "This timeout is non-retryable by design" to `activities/timeouts.mdx` across the SDKs and landed on 7 of 8 — Ruby missed. Of 117 commits touching an SDK landing page in the last 18 months, 52 touched exactly one page and 3 touched all eight.

So "one descriptive sentence per section" means 72 strings maintained in parallel across 8 SDKs, and the history says parallel maintenance is the thing that does not happen here. Prose is still the right answer for the *section index* pages, 52 of which are a banner and a bullet list with no prose at all — but that is a separate change.

## What #5179 does

Wraps each `## [Section](url)` list on the 8 landing pages in markers and fills it from `sidebars.js`. It adds 11 missing links, including Serverless Workers to all six SDKs that lacked it.

[...]

PR #5155, opened 2026-08-21
Fix copy-pasted titles and labels on SDK section index pages

DABH:
## What

Frontmatter on `docs/develop/<sdk>/<section>/index.mdx` pages that was copied from another page and never updated.

| File(s) | Field | Was | Now |
|---|---|---|---|
| all 7 `*/platform/index.mdx` | `title` | `Client - <X> SDK` | `Platform - <X> SDK` |
| all 7 `*/platform/index.mdx` | `sidebar_label` | `Client` | `Platform` |
| `typescript/activities` | `title` | `Workflows - TypeScript SDK` | `Activities - TypeScript SDK` |
| `typescript/activities` | `description` | "...implement **Workers**..." | "...implement **Activities**..." |
| `typescript/workflows` | `description` | "...implement **Workers**..." | "...implement **Workflows**..." |
| `java/workers` | `sidebar_label` | `Nexus` | `Workers` |
| `java/workflows` | `sidebar_label` | `Nexus` | `Workflows` |

## Why it matters (and where it doesn't)

`sidebars.js` sets an explicit `label:` on each of these categories, so the **left nav was already correct** and the `sidebar_label` values were inert — wrong, but invisible to readers. Fixed anyway so the next person to copy one of these files doesn't inherit it.

The `title` and `description` values were **not** inert. They drive the H1, the browser title, the OG card, and the search snippet. Verified in the built output before the fix:

```
build/develop/python/platform/index.html   <h1>Client - Python SDK
```

A reader clicking **Platform** in the sidebar landed on a page whose heading said **Client**. After:

```
python/platform          Platform - Python SDK
typescript/activities    Activities - TypeScript SDK
typescript/workflows     Workflows - TypeScript SDK
java/workers             Workers - Java SDK
```

## How these were found

Not by eye — by scripted audit over all 68 `docs/develop/**/index.mdx` pages, comparing the leading noun of each `title` against its own directory name, and separately flagging any `description` naming a section other than its own. Both audits are clean after this change. The same audit is worth re-running after any future page-copy.

## Checks

- `yarn build` passes; H1s verified in the built HTML
- `vale --config .vale-ci.ini` on all 11 changed files: 0 errors, 0 warnings, 0 suggestions
- No URL, slug, or `id` changes, so no redirects needed

## Not included

The 7 platform pages also share the description "This section explains how to implement platform with the `<X>` SDK" — grammatically odd and lowercase, but not factually wrong, so rewording it is a separate editorial change rather than a bug fix.

Found while investigating #5144.


DABH:
Widened the audit and found a fourth instance, now fixed in this PR.

My first pass only scanned `**/index.mdx`, which was too narrow — the same copy-paste defect exists on a leaf page:

`docs/develop/java/activities/timeouts.mdx` described itself as *"This section explains how to implement **Workers** with the Java SDK"* on an Activity Timeouts page. It has been live since `8e508b879` and survived `e9ace0b2a`, which edited 8 lines of that very file without noticing.

Re-ran across **every** `.mdx` under `docs/develop/` (not just index pages): clean now.

For anyone re-checking later, this is the audit — it flags a page whose `description` names a section other than the one it lives in:

```bash
for f in $(find docs/develop -name '*.mdx'); do
  d=$(awk '/^description:/{p=1;sub(/^description: */,"");if($0!="")print;next} p&&/^[a-zA-Z_]+:/{exit} p{print}' "$f" | tr '\n' ' ')
  sec=$(echo "${f#docs/develop/}" | cut -d/ -f2)
  for other in Workflows Activities Workers Client Nexus; do
    lo=$(echo "$other" | tr 'A-Z' 'a-z')
    [ "$(echo $sec | tr 'A-Z' 'a-z')" != "$lo" ] && echo "$d" | grep -qw "implement $other" && echo "$f"
  done
done
```

Worth noting what this defect class implies: every one of these pages' *links* is correct, because `docusaurus.config.js` sets `onBrokenLinks: 'throw'` so a bad link fails the build. The prose one-liners have no such check, and that is exactly where all four errors are.

Duncanma (approved):

PR #5202, opened 2026-08-28
Add guide for migrating Sidekiq job queues to Standalone Activities

brianmacdonald-temporal:
## What does this PR do?

Adds a Ruby guide, **Migrate a Sidekiq job queue to a Temporal Standalone Activity**, that walks through converting a Sidekiq job into a Standalone Activity:

- Maps Sidekiq concepts (job, `sidekiq` process, Redis, `perform_async`, `sidekiq_options retry:`, Web UI) onto their Temporal equivalents
- Runs an Activity-only Worker with no Workflow code
- Replaces `perform_async` with `client.start_activity()`, and adds a return value via `client.execute_activity()` — something Sidekiq's fire-and-forget model can't provide
- Migrates `sidekiq_options retry: N` to a Retry Policy (with the total-attempts vs. retries adjustment: `retry: 5` -> `max_attempts: 6`)
- Uses `client.list_activities()` and `client.count_activities()` in place of the Sidekiq Web UI

Also links the page from the Guides sidebar and adds a card to the Guides landing-page grid under the existing `Migration` tag.

## Notes to reviewers

Opened as a draft: the guide documents Standalone Activities, which are `publicPreview` in `src/constants/featureReleaseTypes.js`.

All Temporal doc links are relative to satisfy `Temporal.RelativeLinks`. Sidekiq-specific external links (sidekiq.org) remain absolute.

Two review conventions from #5135 are already applied: core identifiers use `Id` (Activity Id, Run Id) with generic "user identifier", and product name is "Temporal Service".


docs/guides/sidekiq-to-standalone-activity.mdx
```diff
@@ -0,0 +1,407 @@
...
+| `sidekiq_options retry: N` | `RetryPolicy` | Automatic retries |
+| Sidekiq Web UI / API | `client.list_activities` / `client.count_activities` | Monitor jobs |
+
+One difference stands out in that table: Sidekiq has no result backend, so a job cannot return a value to its caller. Temporal can, which is why the `execute_activity` row has no Sidekiq equivalent. You will see this pay off in Step 6.
+
+## Prerequisites
+
+Before you begin, you will need the following:
+
+- Ruby 3.2 or higher installed on your machine (3.3 or higher is recommended, since the Worker can then run on fibers).
```
chris-olszewski, line 49: SAA support was only added to versions of the Ruby SDK that require 3.3 or higher

docs/guides/sidekiq-to-standalone-activity.mdx
```diff
@@ -0,0 +1,407 @@
...
+User = Struct.new(:user_id, :email)
+
+class SendWelcomeEmail < Temporalio::Activity::Definition
+  # Sidekiq jobs take simple, JSON-safe arguments. Here we accept a single
+  # hash so you can add fields later without breaking callers. Use string
+  # keys so values round-trip cleanly through Temporal's JSON conversion.
+  def execute(input)
+    user = get_user(input["user_id"])
+    deliver_email(user.email, "Welcome!")
+    "sent to #{user.email}"
```
chris-olszewski, line 165: I find it a little odd that the activity definition doesn't match the sidekiq job. It gives the appearance of SAA needing to be handled in a special manner. Maybe having shared `get_user`/`deliver_email` helpers between the sidekiq job and activity to show that they use the same exact logic.

chris-olszewski (approved):
Overall looks good to me. I do think keeping the sidekiq / activity definition as close as possible will help illustrate how straightforward this migration is.

Also passed this onto @GregoryTravis who implemented SAA in Ruby for additional 👀

docs/guides/sidekiq-to-standalone-activity.mdx
```diff
@@ -0,0 +1,407 @@
...
+  end
+
+  private
+
+  # --- Mock helpers -------------------------------------------------------
+  # Replace these with your real user lookup and mailer when you adapt this.
+  def get_user(user_id)
```
GregoryTravis, line 172: I agree with what @chris-olszewski said above -- perhaps have a single class that incorporates the functions below as methods, and include it in both code samples, so the user could actually copy + paste the whole example and see it run.

docs/guides/sidekiq-to-standalone-activity.mdx
```diff
@@ -0,0 +1,407 @@
...
+puts "Result: #{result}"
+```
+
+Here, `max_attempts: 6` matches Sidekiq's `retry: 5`, and `max_interval` (in seconds) caps the backoff between attempts. The `non_retryable_error_types` list names errors that should fail immediately without retrying — useful for permanent failures such as a missing record, where retrying cannot help. To raise such an error from the Activity, use `Temporalio::Error::ApplicationError` with `non_retryable: true` in `my_activity.rb`:
+
+```ruby
+# my_activity.rb (excerpt)
+require "temporalio/error"
+
+def execute(input)
```
GregoryTravis, line 338:
```suggestion
# ...
        def execute(input)
```
Indent the method so it matches the full source from before.

docs/guides/sidekiq-to-standalone-activity.mdx
```diff
@@ -0,0 +1,407 @@
...
+  user = get_user(input["user_id"])
+  raise Temporalio::Error::ApplicationError.new(
```
GregoryTravis, line 340: since get_user() (above) always returns a value, and since you say that the mock `deliver_email` will throw, I think you can leave this out.

GregoryTravis (approved):

PR #5210, opened 2026-08-31
Add Arize AX to Python integration listings

arizedatngo:
## Summary
- Add Arize AX to the Temporal integrations catalog as a Python agent observability option.
- Add the same Arize AX Temporal guide to the Python SDK integrations list.

## Notes
The linked guide uses Temporal Python OpenTelemetry tracing plus OpenInference instrumentation so Temporal AI-agent spans can be exported to Arize AX. This mirrors the existing external observability listings for Braintrust and Langfuse without adding a new Temporal SDK plugin page.

## Testing
- `jq empty src/components/IntegrationsGrid/integrations-data.json`
- `git diff --check`

### DIFF
--- a/docs/develop/python/index.mdx
+++ b/docs/develop/python/index.mdx
@@ -83,6 +83,7 @@ From there, you can dive deeper into any of the Temporal primitives to start bui
 
 ## [Integrations](/develop/python/integrations)
 
+- [Arize AX integration](https://arize.com/docs/ax/integrations/python-agent-frameworks/temporal/temporal-tracing)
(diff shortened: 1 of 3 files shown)


arizedatngo:
Updated the Python integrations intro to clarify the split between Temporal-maintained Plugin-system integrations and external observability guides that use OpenTelemetry/open standards. That should avoid implying Arize AX is a Temporal SDK plugin.

arizedatngo:
Closing this PR since the Temporal team will make the docs update directly on their side. Thanks for taking a look.

PR #5251, opened 2026-09-04
GRWMK-419: Improve Algolia Search Experience

davidcarlsonberg:
## What does this PR do?

- Uses the experimental Algolia index to evaluate search improvements without changing production search.
- Adds inline SDK context to ambiguous results, for example `Quickstart - Ruby SDK`.
- Avoids repeating the SDK label when it is already included in a result title.
- Adds hierarchy context to section results so readers can identify the source page and area.
- Shows matched content snippets for content-bearing results, including section-level results.
- Retains category grouping while making SDK-specific results easier to distinguish.
- Supports Algolia Dashboard query rules that promote SDK landing pages for exact SDK-name searches.
- Evaluates an experimental crawler configuration that indexes headings through H3 rather than H6 to reduce noisy, low-value results.

## Notes to reviewers

- Test navigational queries: `Quickstart`, `Java`, and `Workflow`.
- Test conceptual queries: `what is durable execution` and `how do I start a Workflow`.
- Snippets appear when Algolia returns a content-bearing record. Title and heading matches may return heading records with no content snippet.
- The experimental crawler and index configuration are managed in the Algolia Dashboard, not in this repository.


davidcarlsonberg:
@Duncanma How does this look to you? Should we go over anything or are all the criteria met? Happy to do the work to get this to "done" even if that means more improvements that we brainstorm.

Duncanma:
> Preview

@davidcarlsonberg this is good... I still think there is room for improvement, but if you want to make the algolia changes to the production index and then go ahead with this, go for it

Duncanma (approved):

PR #5276, opened 2026-09-09
Correct Priority and Fairness design patterns

jpvotta:
## Summary

- Correct the Priority and Fairness design patterns to match Temporal's dispatch behavior. Priority orders backlogged Tasks, and Fairness distributes dispatches across Fairness keys. Neither feature allocates Worker capacity or preempts running Tasks.
- Keep the original use-case framing while correcting inaccurate terminology, diagrams, comparisons, and limitations.
- Link to the Task Queue Priority and Fairness guide for setup and SDK examples instead of duplicating them.
- Rename the Priority Task Queues page and related references to Priority.

## Why

The pattern pages described execution and Worker-capacity guarantees that the Matching Service does not provide. These changes align the design guidance with the development guide and server behavior while keeping implementation details in the canonical guide.


joshmsmith (approved):
Looks good to me. Thanks for the PR!

Duncanma (approved):

PR #5279, opened 2026-09-09
Scope Workflow determinism language to Workflow code (core concept pages)

Duncanma:
## Summary
- Scopes "Workflow" + "deterministic" language to "Workflow code" on the canonical `/workflows` concept page and all five Event History encyclopedia pages, so it reads as an implementation constraint on the Workflow's code, not a claim about the whole Workflow/business process (which is never deterministic once Activities, Signals, and external calls are involved).
- Highest-value fix: the Event History pages' "your Workflows themselves need to be deterministic, your application itself does not" sentence — already the right Workflow-vs-application distinction, just needed the "code" scoping word.
- Also drops one redundant "which makes the Workflow's code deterministic" label on workflow-overview.mdx, where the preceding clause already stated the requirement concretely.

## Test plan
- [x] `vale --config .vale-ci.ini docs/` clean on touched files (pre-existing heading-capitalization suggestions only, unrelated to this change)
- [x] `yarn build` passes
- [ ] Reviewer sanity-check: the 5 Event History pages stayed in sync with each other

PR #5281, opened 2026-09-09
Scope Workflow determinism language in AI-integration docs; add advisory Vale rule

Duncanma:
## Summary
- Same scoping fix as #5279 and #5280, applied to the two AI-agent integration guides where the Workflow/Activity split matters most (Google ADK, LangGraph) — including reverting an unnecessary, ungrammatical edit to a table row that was never actually a scoping violation (the subject was a specific function, already correctly scoped).
- Drops "deterministic" entirely from a one-line Code Exchange card blurb on the `/ai` landing page — the term doesn't carry its weight in top-of-funnel marketing copy.
- Adds an advisory (non-CI-gating) Vale rule, `Temporal.DeterminismScope`, that flags bare "Workflow is/are/must be deterministic" and "Workflow determinism" going forward, so future contributions don't reintroduce the pattern this sweep fixes.

## Test plan
- [x] `vale --config .vale-ci.ini docs/` clean on touched files
- [x] Full `vale docs/` run confirms the new rule doesn't false-positive on the many correctly-scoped "Workflow code" instances elsewhere in the repo
- [x] `yarn build` passes


Duncanma:
Updated per team feedback: replaced "Workflow code" with "Workflow Definition" throughout, same correction as #5279 and #5280. Also updated the new advisory Vale rule's message to recommend "the Workflow Definition" instead of "Workflow code/logic," since that's now the standard being taught.

flippedcoder (approved):

PR #5295, opened 2026-09-10
docs: update TS Worker Docker images from EOL Debian bullseye to bookworm

thiagojv:
## Summary

Update the TypeScript SDK Worker Docker guide to stop recommending Debian `bullseye`-based images and use `bookworm` instead.

Changes in `docs/develop/typescript/workers/run-process.mdx`:

- `node:20-bullseye` -> `node:20-bookworm`
- `node:20-bullseye-slim` -> `node:20-bookworm-slim`
- `gcr.io/distroless/nodejs20-debian11` -> `gcr.io/distroless/nodejs20-debian12`

## Why

Debian 11 (`bullseye`) has reached end-of-life for standard support, so the docs should point readers at a currently supported base image. Debian 12 (`bookworm`) is the current stable release. The `node:20-bookworm`, `node:20-bookworm-slim`, and `gcr.io/distroless/nodejs20-debian12` tags are official, published, and drop-in replacements, so the surrounding guidance (glibc requirement, `ca-certificates` on slim images, distroless multi-stage build) stays accurate.

## Testing

- Docs-only change limited to code blocks and inline references; no prose/structure changes.
- Verified no other `bullseye`/`debian11` references remain in the file.

## Risks

Low. The guidance and Dockerfile structure are unchanged; only the base image tags are updated to their supported equivalents.

### DIFF
--- a/docs/develop/typescript/workers/run-process.mdx
+++ b/docs/develop/typescript/workers/run-process.mdx
@@ -255,10 +255,10 @@ Workers based on the TypeScript SDK can be deployed and run as Docker containers
 Use an LTS Node.js release such as 18, 20, 22, or 24. Both `amd64` and `arm64` architectures are supported. A
 glibc-based image is required; musl-based images are _not_ supported (see below).
 
-The most direct way to deploy a TypeScript SDK Worker on Docker is to start with the `node:20-bullseye` image. For example:
+The most direct way to deploy a TypeScript SDK Worker on Docker is to start with the `node:20-bookworm` image. For example:
 
 ```dockerfile
-FROM node:20-bullseye
+FROM node:20-bookworm
 
 # For better cache utilization, copy package.json and lock file first and install the dependencies before copying the
 # rest of the application and building.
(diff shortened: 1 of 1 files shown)


Duncanma:
Thanks for this — clean, well-scoped fix.

**Review notes:**
- Checked the rest of `docs/` for other `bullseye`/`debian11` references: none outside this file, and this PR's diff covers every instance in `run-process.mdx`. No collateral changes needed.
- No structural or tone issues — this is a straight tag swap inside existing code blocks, no prose or heading changes.

**Worth considering (not a blocker):** Debian 13 (`trixie`) has been the current stable release since August 2025 — `bookworm` is now *oldstable* (LTS-supported until 2028). Both `node:20-trixie` and `gcr.io/distroless/nodejs20-debian13` are already published. `bookworm` is still a perfectly reasonable, well-supported choice (and arguably more conservative than jumping straight to the newest release), but wanted to flag it in case you'd rather target `trixie` directly so this doesn't need a follow-up in a year.

One more thing: the CLA check above is still showing as not signed — could you sign the [Contributor License Agreement](https://cla-assistant.io/temporalio/documentation?pullRequest=5295)? We won't be able to merge until that's done.

thiagojv:
Thanks a lot for the thorough review, @Duncanma — really appreciate it!

Good call on `trixie`. I fully agree it's the current stable and a valid target. I intentionally went with `bookworm` here to stay conservative and avoid jumping two Debian releases at once: since the previous guidance was still on the EOL `bullseye`, moving to `bookworm` keeps this a minimal, low-risk drop-in swap that's easy to verify. Targeting `trixie` directly feels like a good follow-up once there's appetite for it, rather than bundling the extra jump into this fix. Happy to open that follow-up (or switch this PR to `trixie`) if you'd prefer.

Also, the CLA is signed now — the `license/cla` check is green. Thanks again!

PR #5305, opened 2026-09-11
Document the OpenTelemetry plugin on the Python observability page

DABH:
The Python observability page's tracing section only covered the `TracingInterceptor`, while the SDK has shipped `OpenTelemetryPlugin` and `create_tracer_provider()` for several releases and the AI integration pages already build on them.

This change rewrites the "Set up tracing" section into two subsections. The plugin subsection shows the replay-safe tracer provider, registering the plugin on the Client, the `add_temporal_spans` toggle, a custom span created from Workflow code, what happens on replay and Worker restart, and a note that the API is experimental. The interceptor subsection keeps the existing guidance, explains why its Workflow-side spans have no duration, and points to `completed_span()`. It also adds the exporter install line, since `temporalio[opentelemetry]` does not include one.

Ran `vale --config .vale-ci.ini` on the page: clean. Follows up on the Arize integration work (AI-411).

### DIFF
--- a/docs/develop/python/platform/observability.mdx
+++ b/docs/develop/python/platform/observability.mdx
@@ -49,17 +49,106 @@ Tracing allows you to view the call graph of a Workflow along with its Activitie
 
 Temporal Web's tracing capabilities mainly track Activity Execution within a Temporal context. If you need custom tracing specific for your use case, you should make use of context propagation to add tracing logic accordingly.
 
-To configure tracing in Python, install the `opentelemetry` dependencies.
+To configure tracing in Python, install the `opentelemetry` dependencies and an exporter for your tracing backend.
 
 ```bash
 # This command installs the `opentelemetry` dependencies.
 pip install temporalio[opentelemetry]
+# Any OpenTelemetry exporter works; this one speaks OTLP.
+pip install opentelemetry-exporter-otlp
 ```
 
-Then the [`temporalio.contrib.opentelemetry.TracingInterceptor`](https://python.temporal.io/temporalio.contrib.opentelemetry.TracingInterceptor.html) class can be set as an interceptor as an argument of [`Client.connect()`](https://python.temporal.io/temporalio.client.Client.html#connect).
+The Python SDK offers two ways to emit OpenTelemetry spans: the `OpenTelemetryPlugin`, which supports spans with real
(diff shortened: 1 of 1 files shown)


docs/develop/python/platform/observability.mdx
```diff
@@ -49,17 +49,106 @@ Tracing allows you to view the call graph of a Workflow along with its Activitie
...
+                my_activity, name, start_to_close_timeout=timedelta(seconds=10)
+            )
+```
+
+Replay is safe. The provider generates span identifiers deterministically from Workflow state and does not export
+spans while a Workflow replays, so Worker restarts and cache evictions never produce duplicate spans. A span that is
+open when a Worker stops is exported once, when the Workflow finishes it on another Worker. Each Activity attempt
+produces its own `RunActivity` span, so retries stay visible.
+
+:::note
```
Duncanma, line 123: Elsewhere on the site, an unstable-API disclaimer uses `:::caution`, not `:::note` — e.g. [openai-agents.mdx](https://github.com/temporalio/documentation/blob/main/docs/develop/python/integrations/openai-agents.mdx#L644-L649) ("Streaming is experimental and may change before general availability") and the Nexus standalone-operations pages ("All APIs are experimental..."). That same openai-agents.mdx page also calls the OpenTelemetry integration built on this same `create_tracer_provider()` **"Public Preview"** rather than "experimental" ([L788-792](https://github.com/temporalio/documentation/blob/main/docs/develop/python/integrations/openai-agents.mdx#L788-L792)). Worth switching this to `:::caution` and confirming the correct stability label with engineering so the two pages agree.

Duncanma: Checked the source to close this out: `sdk-python`'s docstrings for both `OpenTelemetryPlugin` and `create_tracer_provider()` use the identical warning — "This class/function is experimental and may change in future versions. Use with caution in production environments" — and the module README's stability table backs it up. So "experimental" is right as written; no need to check with engineering on the label. The one thing still worth doing is swapping `:::note` for `:::caution` here, since that's the admonition type used for every other experimental-API disclaimer on the site (and maps directly to the SDK's own `.. warning::`).

Duncanma: Pushed a fix directly to this branch: `:::note` → `:::caution` (6abc9f0). Resolving.

Duncanma (approved):

PR #5310, opened 2026-09-11
Add AI-slop vocabulary Vale rules, fix flagged intensifiers in docs

Duncanma:
## What does this PR do?

Adds five new Temporal-owned Vale rules and fixes every real hit one of them found in `docs/`:

- Ports four high-precision rules from [jdkato/voices](https://github.com/jdkato/voices) into `vale/styles/Temporal/` — `Banned`, `InflatedWords`, `WeakVerbs`, `BinaryContrast` — rather than adding the package as a dependency, matching the same-day decision in #5299 to keep Vale limited to Std + Temporal's own hand-picked rules.
- Adds `Temporal.Intensifiers` for "powerful", "seamless", and "unlock" — named in `AGENTS.md`'s style guide but not covered by jdkato/voices.
- Rewrites all 28 real `Temporal.Intensifiers` hits found across `docs/` to state the concrete fact instead of the vague intensifier.

All five new rules are disabled in `.vale-ci.ini` for now (validated against a full-repo run, but whether to gate CI with them is a separate decision) and live in the plain `vale docs/` run as advisory suggestions.

## Notes to reviewers

- Two apparent hits were left alone as false positives: `docs/develop/dotnet/best-practices/testing-suite.mdx` and the Ruby equivalent both use "locked/unlocked" as a technical lock-state term, not the marketing sense the rule targets. Noted in `Intensifiers.yml` for future tuning.
- `Temporal.BinaryContrast` will occasionally flag legitimate "not X, but Y" sentences that aren't AI-slop filler — it's a judgment-call rule by design, kept at `warning`, not something to gate CI on as-is.


jsundai (approved):
Nice!

PR #5314, opened 2026-09-13
Add Task Queue limit warning to Worker Versioning docs

rkannan82:
## What

Adds a prominent caution callout to the Worker Versioning landing page and expands the Task Queue limit entry on the limits page to call out the per-worker Task Queue anti-pattern.

## Why

We want to make it easy for users and AI agents to discover the anti-pattern of enabling versioning for worker-specific Task Queues (for example, a Task Queue name with a UUID). Versioning is not relevant for worker-specific Task Queues, and enabling it will quickly exhaust the Task Queue per Deployment Version limit.


Shivs11 (approved):

Duncanma (approved):

docs/design-patterns/worker-specific-taskqueue.mdx
```diff
@@ -761,6 +761,12 @@ You need to handle cleanup if the Workflow fails mid-process.
 
 ## Best practices
 
+:::caution Worker Versioning
+
+Do not enable [Worker Versioning](/production-deployment/worker-deployments/worker-versioning#definition) for worker-specific Task Queues. See [Task Queue limits](/cloud/limits#max-task-queues-in-deployment-version-limits).
```
ShahabT: should we add the explanation that why it is not needed as you have it in the limits.mdx page? (same for worker-execution-affinity.mdx)

> A per-worker Task Queue's lifetime is tied to the Worker itself. When the Worker is updated to a new version, it gets a new Task Queue and the old one is gone, so versioning is not relevant.

rkannan82: I didn't want to repeat the rationale in multiple places. So slightly reworded the call-site so it does mention the problem - it is not needed, and also will exhaust the task queue limit. User can click on it for details as to why that's the case.

bechols (approved):

PR #5325, opened 2026-09-14
always use latest javadoc link

MasseGuillaume:
I was reading the documentation at https://docs.temporal.io/develop/java/workflows/basics

I noticed that when I click on the link for DataConverter it points to

https://www.javadoc.io/static/io.temporal/temporal-sdk/1.17.0/io/temporal/common/converter/DataConverter.html

this is a 404 page at the time of writing this ticket

it's because javadoc loads documentation in lazy manner

you can see the available versions here:
https://www.javadoc.io/versions/io.temporal/temporal-sdk

I synched the javadoc for this version, it should not be a 404 anymore


Duncanma (commented):
Thanks for chasing down the broken javadoc links — the old pinned versions (e.g. `0.19.0`, `1.17.0`) do go 404, so the underlying problem is real.

However, the replacement links in this PR also 404 for me. They keep the `/static/` path and just swap the pinned version for the literal word `latest`:

```
https://www.javadoc.io/static/io.temporal/temporal-sdk/latest/io/temporal/common/converter/DataConverter.html
```

javadoc.io's `/static/` mirror only pre-renders specific, resolved version numbers — `latest` isn't one, so it never gets generated there and 404s. I checked all four `/static/.../latest/...` URLs this PR introduces (in `activities/basics.mdx`, `best-practices/data-handling/data-conversion.mdx`, `client/temporal-client.mdx`, `workflows/basics.mdx`) and all return 404.

The pattern that already works — and is what the rest of this repo uses in 100+ other javadoc links, e.g. [`workflows/timers.mdx`](https://github.com/temporalio/documentation/blob/main/docs/develop/java/workflows/timers.mdx#L26) — is the `/doc/` path instead of `/static/`. It supports `latest` and resolves it dynamically:

```
https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/common/converter/DataConverter.html
```

I verified this returns 200 for the same five pages.

There's also a second, unrelated issue: in `workflows/side-effects.mdx`, the new link drops `io/temporal/` from the path (`.../latest/workflow/Workflow.html` instead of `.../latest/io/temporal/workflow/Workflow.html`), which also 404s.

I've opened #5328 against `main` with these links switched to `/doc/` (and the side-effects.mdx path fixed), so this can be closed once that merges. Thanks again for flagging the original 404s!

